PR #7, opened 2016-03-10
Popover Component

### LINKED ISSUE #5: Component: Popover
http://v4-alpha.getbootstrap.com/components/popovers/

Should use TetherContent similar to [Tooltip](https://github.com/eddywashere/reactstrap/blob/master/lib/Tooltip.jsx)

Props

- `isOpen` - boolean
- `toggle` - function
- `placement` - string
- `target` - string
- `tether` optional object

Inner Components

- PopoverTitle
- PopoverContent

cbfx:
Almost got this but had two snags:

The TetherContent component was calling the toggle that is passed down into it because of its `handleDocumentClick` method. The problem was that I was also calling it by assigning a click handler to the parent `_target`. This was causing the listener to fire once and the document click inside tether to fire once which effectively toggled on and then back off the popover. I circumvented this by passing in a noop function to the TetherContent toggle. This seems wrong and I assume I should feel bad. Thought I'd bring it up.

Second, I'm not doing so hot with the test spec for the Popover. Plzhalp.

@eddywashere 

Fixes #5

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,6 +8,7 @@ import DropdownMenu from './DropdownMenu';
 import DropdownToggle from './DropdownToggle';
 import Label from './Label';
 import PopoverTitle from './PopoverTitle';
+import PopoverContent from './PopoverContent';
 import TetherContent from './TetherContent';
 import Tooltip from './Tooltip';
 
@@ -21,6 +22,7 @@ export {
   DropdownMenu,
   DropdownToggle,
   Label,
+  PopoverContent,
   PopoverTitle,
   TetherContent,
   Tooltip
--- a/lib/index.js
+++ b/lib/index.js
@@ -7,6 +7,7 @@ import DropdownItem from './DropdownItem';
 import DropdownMenu from './DropdownMenu';
 import DropdownToggle from './DropdownToggle';
 import Label from './Label';
+import PopoverTitle from './PopoverTitle';
 import TetherContent from './TetherContent';
 import Tooltip from './Tooltip';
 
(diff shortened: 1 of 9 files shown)


lib/index.js
```diff
@@ -20,6 +23,9 @@ export {
   DropdownMenu,
   DropdownToggle,
   Label,
+  Popover,
+  PopoverContent
```
eddywashere: missing `,`

eddywashere:
If you update the pr, I'll look at the tests. Thanks for your help!!!

cbfx:
@eddywashere oye, i updated and simplified the Popover component but haven't checked the tests. I assume i need to simulate a click, etc and i'll try that now. let me know if i'm in the right direction on this component. i had to add a `defaultProps.toggle` function to pass through to the TetherComponent so it wouldn't complain about the required prop.

this is current implementation:
```jsx
<Button id="Popover-top" onClick={() => { this.setState({ Popover1: !this.state.Popover1 })}}>Open Popover</Button>

<Popover placement="top" isOpen={this.state.Popover1} target="Popover-top">
  <PopoverTitle>Title</PopoverTitle>
  <PopoverContent>Content</PopoverContent>
</Popover>
```

should there be some toggle prop on Popover? if so, what? can you post expectation of implementation?

eddywashere:
Yes there should be a toggle prop on Popover if the outside/document click should toggle it. If not, it could just be the no op you set as a default. It would take a manual toggle or target toggle to close it.

It could be as simple as having the same inline toggle function the target onClick prop has. As I suggested that, I thought maybe we should demo a real world example.

Here's how people could apply the popover to a list of items https://gist.github.com/eddywashere/11430eb3e3d0e03b1cad. It wouldn't make sense to iterate through them in the container component like our tooltip example, but rather add the state and toggle via a list item component.

eddywashere:
:+1:

PR #61, opened 2016-07-28
Update to bootstrap 4 alpha 3

TheSharpieOne:
Bootstrap 4 Alpha 3 was released yesterday. This change includes differences between alpha 2 and alpha 3 for the components which currently exist.
These changes are:
 - Button outline class change from `.{variant}-outline` to `.outline-{variant}`.
 - label is now tag and all of the label variants are now tag variants.
 - Cards now have an outline variant.

I have put each of these changes as a separate commit in case there is a problem/issue with any one of them.


TheSharpieOne:
Sorry, missed the test when changing Label to Tag. That has been added in a new commit. Should be good now.

eddywashere:
Woot!!! 

Node 4 seems flakey in Travis. I'll review this asap. Thanks!!

eddywashere:
@TheSharpieOne this looks great. Thanks! I'll drop node 4 in pr that releases this and make sure to do a major version bump.

Issue #33, opened 2016-04-24
Tracking Bootstrap 4 Changes


eddywashere:
Here's some changes I've seen between the last release, [v4.0.0-alpha.2](https://github.com/twbs/bootstrap/tree/v4.0.0-alpha.2) and the [v4-dev](https://github.com/twbs/bootstrap/tree/v4-dev) branch
- [x] outline classes are no longer appended to classnames. ex: `.btn-primary-outline` -> `.btn-outline-primary`
- [x] Label is now Tag. ex: `.label` -> `.tag`

eddywashere:
Moving the `outline` option to a boolean based prop from the color prop could simplify upgrading (for the Button component).

eddywashere:
- [ ] Column offset, pull and push classes are now `.push-md-2`, `.pull-md-2`, `.offset-md-2`.

eddywashere:
Closing this in favor of specific issues that cover a change and track the status of it.

PR #62, opened 2016-07-28
feat(outline): Make outline a separate prop

TheSharpieOne:
The outline for buttons and cards is now it's own props separate from the color value.
Based on https://github.com/reactstrap/reactstrap/issues/33#issuecomment-214025488

This also adds additional tests around this new props.

Also:
Cards, unlike buttons, do not have a default color. Thus adding outline prop to a card without a color will do nothing.
Is this acceptable? Should it produce a warning/error? Should it default to 'card-outline-secondary' when outline is provided but no color is provided?

### DIFF
--- a/docs/lib/Home/index.jsx
+++ b/docs/lib/Home/index.jsx
@@ -21,7 +21,7 @@ export default () => {
                 Easy to use React Bootstrap 4 components
               </p>
               <p>
-                <Button color="danger-outline" href="https://github.com/reactstrap/reactstrap">View on Github</Button>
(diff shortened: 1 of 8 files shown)


eddywashere:
@TheSharpieOne thanks a ton for the prs!

I think this is acceptable behavior. I think card already has a solid style without background variants applied where as the default `.btn` kind of looks weird and isn't really shown in the docs.

re: warnings/errors. I'd love to throw warnings in dev mode that catch issues like this and not include those warnings in production builds.

Issue #50, opened 2016-07-10
Forms


akatebi:
Hi Eddy,
   
You have done a fantastic job with this package. Any ideas when Forms and other components will become available? 

Thanks so much!
Alex

eddywashere:
Thanks! I'm kind of prioritizing stuff based on my personal usage. Right now I'm working on some React 15.2 warnings around unknown props.

I think I kind of dropped the ball on a good contributing section to help enable folks to submit issues/prs for adding and requesting components. Once I get a prioritized list I'll update this issue and add some rough timelines to components.

TheSharpieOne:
Forms are in release 2.1.0

PR #67, opened 2016-07-29
feat(form): add forms and form element components

TheSharpieOne:
#50 
Forms and form element components as well as docs and tests including:

- Form
- FormGroup
- FormFeedback
- FormText
- Input
- Label

What this **doesn't** include for forms:

- [input-group and input-group-addon](http://v4-alpha.getbootstrap.com/components/forms/#hidden-labels)
- something about grouping [radio and checks](http://v4-alpha.getbootstrap.com/components/forms/#checkboxes-and-radios) (this seems like something twbs is changing in v4 (new `.form-check` vs `.radio` and `.check`) half of their docs use or the other)
- [custom forms](http://v4-alpha.getbootstrap.com/components/forms/#custom-forms) (Is this new to bs4, don't think I have seen this before)
- fieldsets (doesn't seem like it would be useful since everything would just pass right though `<Fieldset disabled>` -> `<fieldset disabled>`

Also, these are the basic building block components, there are no helper/wrapping components (like a single component which will build out a form group with a label and an input for you.) The helper/wrapping components can utilize these, if you determine this library should have such components (in which case, let me know and I can add them).

I did the best I could with docs, there are a lot of examples which shows everything, but not everything has a dedicated section.

### DIFF
--- a/src/FormFeedback.jsx
+++ b/src/FormFeedback.jsx
@@ -1,0 +1,34 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+const propTypes = {
+  children: PropTypes.node,
+  tag: PropTypes.string,
+  className: PropTypes.string,
+};
+
+const defaultProps = {
+  tag: 'div',
+};
+
+const Form = (props) => {
+  const {
+    className,
+    tag: Tag,
+    ...attributes,
+  } = props;
+
+  const classes = classNames(
+    className,
+    'form-control-feedback'
+  );
+
+  return (
+    <Tag {...attributes} className={classes} />
+  );
+};
+
+Form.propTypes = propTypes;
+Form.defaultProps = defaultProps;
+
+export default Form;
--- a/src/FormFeedback.jsx
+++ b/src/FormFeedback.jsx
@@ -11,7 +11,7 @@ const defaultProps = {
   tag: 'div',
 };
 
-const Form = (props) => {
+const FormFeedback = (props) => {
   const {
     className,
     tag: Tag,
@@ -28,7 +28,7 @@ const Form = (props) => {
   );
 };
 
-Form.propTypes = propTypes;
-Form.defaultProps = defaultProps;
+FormFeedback.propTypes = propTypes;
+FormFeedback.defaultProps = defaultProps;
 
-export default Form;
+export default FormFeedback;
--- a/src/FormGroup.jsx
+++ b/src/FormGroup.jsx
@@ -15,7 +15,7 @@ const defaultProps = {
   tag: 'div',
 };
 
-const Form = (props) => {
+const FormGroup = (props) => {
   const {
     className,
     row,
@@ -39,7 +39,7 @@ const Form = (props) => {
   );
 };
 
-Form.propTypes = propTypes;
-Form.defaultProps = defaultProps;
+FormGroup.propTypes = propTypes;
+FormGroup.defaultProps = defaultProps;
 
-export default Form;
+export default FormGroup;
--- a/src/FormGroup.jsx
+++ b/src/FormGroup.jsx
@@ -1,0 +1,45 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+const propTypes = {
+  children: PropTypes.node,
+  row: PropTypes.bool,
+  check: PropTypes.bool,
+  disabled: PropTypes.bool,
+  tag: PropTypes.string,
+  color: PropTypes.string,
+  className: PropTypes.string,
+};
+
+const defaultProps = {
+  tag: 'div',
+};
+
+const Form = (props) => {
+  const {
+    className,
+    row,
+    disabled,
+    color,
+    check,
+    tag: Tag,
+    ...attributes,
+  } = props;
+
+  const classes = classNames(
+    className,
+    color ? `has-${color}` : false,
+    row ? 'row' : false,
+    check ? 'form-check' : 'form-group',
+    check && disabled ? 'disabled' : false
+  );
+
+  return (
+    <Tag {...attributes} className={classes} />
+  );
+};
+
+Form.propTypes = propTypes;
+Form.defaultProps = defaultProps;
+
+export default Form;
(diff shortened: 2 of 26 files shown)


byennen:
:+1:

eddywashere:
Apologies for the delay in reviewing this. It was a bigger pr and I didn't have a large block of time until today.

- I think the examples are sufficient until it becomes clearer in bootstrap docs. either way it could be improved in another pr
- Custom forms is definitely new.
- Just read up on fieldset caveats in the [docs](http://v4-alpha.getbootstrap.com/components/forms/#caveat-about-link-functionality-of-a). That could have been the default tag for FormGroup. Seems great for modern browsers, but might be more trouble than it's worth.

re: "Also, these are the basic building block components, there are no helper/wrapping components"

I usually put together something like that for projects I'm working on. ex: forms, I usually have a formgroup,  error and validation wrapped components (tailored to app ux). If they exist, they should be part of another repo like `reactstrap-helpers`. It would be a good reference for people learning how to piece things together.

src/Input.jsx
```diff
@@ -0,0 +1,67 @@
...
+  type: PropTypes.string,
+  size: PropTypes.string,
+  state: PropTypes.string,
+  tag: PropTypes.string,
+  'static': PropTypes.bool,
+  className: PropTypes.string,
+};
+
+const defaultProps = {
+  tag: 'p',
```
eddywashere: I wasn't sure why this would ever happen but then I remembered the static text class. 👍

src/Input.jsx
```diff
@@ -0,0 +1,67 @@
...
+    tag,
+    'static': staticInput,
+    ...attributes,
+  } = props;
+
+  const checkInput = ['radio', 'checkbox'].indexOf(type) > -1;
+
+  const fileInput = type === 'file';
+  const textareaInput = type === 'textarea';
+  const selectInput = type === 'select';
```
eddywashere: Should `textarea` and `select` get moved into their own components? 

- I think it's super convenient wrapping this all in one component,
- but I also think it might be misleading to people since that's not a valid type attribute (yeah we get flexibility with react components but it's something to think about)

src/Input.jsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  const classes = classNames(
+    className,
+    state ? `form-control-${state}` : false,
+    size ? `form-control-${size}` : false,
+    formControlClass
+  );
+
+  if (!staticInput && !selectInput) {
```
eddywashere: if `textarea` and `select` stay within input, this should also check for textareaInput as to not add type to rendered dom.

src/Input.jsx
```diff
@@ -0,0 +1,67 @@
...
+  'static': PropTypes.bool,
+  className: PropTypes.string,
+};
+
+const defaultProps = {
+  tag: 'p',
+  type: 'text',
+};
+
+const Form = (props) => {
```
eddywashere: Can you update the exported component name from `Form` to `Input`?

src/FormFeedback.jsx
```diff
@@ -0,0 +1,34 @@
...
+  children: PropTypes.node,
+  tag: PropTypes.string,
+  className: PropTypes.string,
+};
+
+const defaultProps = {
+  tag: 'div',
+};
+
+const Form = (props) => {
```
eddywashere: `Form` to `FormFeedback`

src/FormGroup.jsx
```diff
@@ -0,0 +1,45 @@
...
+  tag: PropTypes.string,
+  color: PropTypes.string,
+  className: PropTypes.string,
+};
+
+const defaultProps = {
+  tag: 'div',
+};
+
+const Form = (props) => {
```
eddywashere: `Form` to `FormGroup`

src/FormText.jsx
```diff
@@ -0,0 +1,39 @@
...
+  tag: PropTypes.string,
+  color: PropTypes.string,
+  className: PropTypes.string,
+};
+
+const defaultProps = {
+  tag: 'small',
+};
+
+const Form = (props) => {
```
eddywashere: `Form` to `FormText`

src/Label.jsx
```diff
@@ -0,0 +1,86 @@
...
+  md: columnProps,
+  lg: columnProps,
+  xl: columnProps,
+};
+
+const defaultProps = {
+  tag: 'label',
+};
+
+const Form = (props) => {
```
eddywashere: `Form` to `Label`

eddywashere:
A few comments. Great work!

src/Label.jsx
```diff
@@ -0,0 +1,86 @@
...
+    className,
+    hidden ? 'sr-only' : false,
+    check ? `form-check-${inline ? 'inline' : 'label'}` : false,
+    size ? `col-form-label-${size}` : false,
+    colClasses,
+    colClasses.length ? 'col-form-label' : false
+  );
+
+  return (
+    <Tag {...attributes} className={classes} htmlFor={htmlFor} />
```
eddywashere: consider adding `{...attributes}` after `htmlFor={htmlFor}`, I could imagine someone not using `for` and sticking with `htmlFor`, in which case it would be overridden with a falsey attribute.

TheSharpieOne:
Sorry about the copy-pasta with the exported component name. I'll clean these up Monday.

As for taking select and textarea out, would it be acceptable to leave them in and create convenience components called `Select` and `Textarea` which simply wrap `Input` to set the type? I believe it will make this project's code a bit cleaner (imho, since then there will not be components which do the similar thing except for output a different tag) and more clear for the dev user (seeing `Select` instead of `Input type="select"`) while also making it easy if they Inputs are dynamically created (by the dev user via some config and custom things they set up (no need for a switch to determine which component to use, always Input).

eddywashere:
Hmmm. That's a fair point. There would still be a switch for type and props assigned.  In that case, no need to include Select & Textarea in this pr. If it's a big enough problem it can be addressed in a later pr without a breaking change. Removing it after it existed would be a breaking change.

eddywashere:
👍

PR #68, opened 2016-07-29
fix(docs): update path to allow direct loading of the tags doc page

TheSharpieOne:
When running localhost / dev webpack, the tags route cannot be navigated to directly (or reloaded). This is because I missed a reference when updating labels to tags. This PR fixes that.


eddywashere:
Thanks! Phantomjs seems flakey. Saw the test pass across v6. Maybe moving to chrome headless could make this more predictable.

PR #86, opened 2016-08-05
Use proper case to prevent Don't Call PropTypes warning

jsarman:
https://github.com/facebook/react/pull/7132


jsarman:
Sorry my editor took over. I will recreate.

PR #87, opened 2016-08-05
Use proper case to prevent Don't Call PropTypes warning

jsarman:
https://github.com/facebook/react/pull/7132

### DIFF
--- a/src/Button.jsx
+++ b/src/Button.jsx
@@ -65,7 +65,7 @@ class Button extends React.Component {
   }
 }
 
-Button.propTypes = propTypes;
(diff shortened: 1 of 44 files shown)


jsarman:
Sorry that test fixed my issue was wasnt proper it should be lower case.

TheSharpieOne:
https://facebook.github.io/react/warnings/dont-call-proptypes.html

> The normal usage of PropTypes is still supported:
```
 Button.propTypes = {
   highlighted: React.PropTypes.bool
 };
```
> Nothing changes here.

PR #89, opened 2016-08-06
Update eslint, deps & update src files

eddywashere:


src/Dropdown.jsx
```diff
@@ -1,3 +1,6 @@
+/* eslint react/no-find-dom-node: 0 */
+// https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-find-dom-node.md
```
eddywashere: > Facebook will eventually deprecate findDOMNode as it blocks certain improvements in React in the future.

Might be worth keeping track of this discussion at https://github.com/yannickcr/eslint-plugin-react/issues/678

eddywashere: Didn't update this just yet to use `ref={node => this.node = node}` because `node` for some reason is null sometimes. Specifically on dropdown toggle click. https://github.com/reactstrap/reactstrap/blob/0c1b5e48c208ac4542357534defa7681e467fbc9/src/Dropdown.jsx#L183

gaearon: If you see a bug in React, file an issue in React please. 😉

PR #91, opened 2016-08-06
feat(Card): fix image class

mking-clari:
Minor fix for card image border radius.

Note that a card image should have one of `card-img`, `card-img-top`, or `card-img-bottom` (this is a Bootstrap inconsistency).

Also, when a card image is in the middle of a card, a plain image is used in the Bootstrap example. We may still want to use CardImg here, but ensure no classes are applied.


eddywashere:
👍

PR #92, opened 2016-08-06
feat(Dropdowns): use split class for split dropdowns

mking-clari:
Minor fix for split dropdown. The padding around the caret should be smaller. We can make this adjustment using the `dropdown-toggle-split` class.


eddywashere:
Currently there's not a good pattern for disabling classes that are included by default. Otherwise I'd suggest adding this to caret prop. Seems safer to opt in for now. :+1:

PR #93, opened 2016-08-07
WIP: feat(Alerts): add Alerts component

cbfx:


src/Alert.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+const alertTypes = ['success', 'info', 'warning', 'danger'];
+
+const propTypes = {
+  color: React.PropTypes.oneOf(alertTypes),
```
TheSharpieOne: Nooo, please. I the one thing I really love if the fact that there is not extra BS restricting components to the basic/default BS. (BS means two different things there). 
React bootstrap does this and it is loaded with all of these constraints making it heavy. They eventually added even more weight by adding a way to add custom variants to individual components. 
I like this project because it doesn't do that. I can pass a color of "whatever" and it will output it in a class that makes sense given the specific component. I use my css to extend BS following their pattern (such as "alert-whatever" in this case). 
My point is, this is pointless unneeded validation which adds more weight total the project and ends up restricting what the user can do.

cbfx: @TheSharpieOne yo! thanks for the input. i'll talk to @eddywashere and see what we can do to take your use case into account. thanks a billion for the feedback. : )

cbfx:
yo @eddywashere - i'm going close this out and let someone else take it. i didn't have time to finish it today but perhaps tomorrow i can put some official time into it during the day. i'll re-open if i can get to it. sorry m8.

Issue #75, opened 2016-07-31
Media object


eddywashere:
http://v4-alpha.getbootstrap.com/layout/media-object/

- [x] Media

eddywashere:
Released in 2.4.0

PR #94, opened 2016-08-08
feat(Media): add component

mking-clari:
References #75

Notes
- The media components are fairly lightweight (almost single class). This is particularly true for `MediaHeading` and `MediaObject`.
- I'm using `MediaMarginal` for shared code between `MediaLeft` and `MediaRight`. It's not intended for direct use by users, but is exposed so we can test it.
- In `MediaList`, I'm applying `tag="li"` to the children. This way, the user of `MediaList` doesn't have to specify `li` on each `Media` item.
- I'm using [Holder](http://holderjs.com) to populate placeholder images, similar to the Bootstrap docs. Also, I had to rerun Holder on route update. Without this, if the user goes to `/components/media`, then `/components/buttons`, then hits back, the placeholder image isn't populated.

### DIFF
--- a/src/Media.jsx
+++ b/src/Media.jsx
@@ -1,0 +1,72 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+const propTypes = {
+  body: PropTypes.bool,
+  bottom: PropTypes.bool,
+  children: PropTypes.node,
+  className: PropTypes.string,
+  heading: PropTypes.bool,
+  left: PropTypes.bool,
+  list: PropTypes.bool,
+  middle: PropTypes.bool,
+  object: PropTypes.bool,
+  right: PropTypes.bool,
+  tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
+  top: PropTypes.bool,
+};
+
+const Media = (props) => {
+  const {
+    body,
+    bottom,
+    className,
+    heading,
+    left,
+    list,
+    middle,
+    object,
+    right,
+    tag,
+    top,
+    ...attributes,
+  } = props;
+
+  let defaultTag;
+  if (heading) {
+    defaultTag = 'h4';
+  } else if (left || right) {
+    defaultTag = 'a';
+  } else if (object) {
+    defaultTag = 'img';
+  } else if (list) {
+    defaultTag = 'ul';
+  } else {
+    defaultTag = 'div';
+  }
+  const Tag = tag || defaultTag;
+
+  const classes = classNames(
+    className,
+    {
+      'media-body': body,
+      'media-heading': heading,
+      'media-left': left,
+      'media-right': right,
+      'media-top': top,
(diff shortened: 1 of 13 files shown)


src/Media.jsx
```diff
@@ -0,0 +1,43 @@
...
+  } = props;
+
+  const classes = classNames(
+    className,
+    'media'
+  );
+
+  return (
+    <Tag {...attributes} className={classes}>
+      {left}
```
eddywashere: This kind of goes against one of the goals for the project. "content is expected to be composed via props.children rather than using named props to pass in Components." Mostly for consistency. Any reason you didn't use a generic Media component for left/right. Placement could be determined by the order it is presented in the markup.

mking-clari: Ok, I see. Would it look like this?

```
<Media>
  <MediaLeft ... />
  <MediaBody ... />
<Media>

<Media>
  <MediaBody ... />
  <MediaRight ... />
<Media>
```

eddywashere: What about `<Media><Media left/><MediaBody/></Media>` and `<Media right/>`

mking-clari:
@eddywashere I updated the code to use `<Media left>` and `<Media right>`.

eddywashere:
I think that makes sense `<Media body>`, `<Media heading>`, and `<Media object>`. What about `<Media top>`, `<Media middle>`, `<Media bottom>`? Is that too much?

mking-clari:
@eddywashere I agree, I think it makes the usage of the media component easier to remember (more consistent).

mking-clari:
@eddywashere I updated the code.

A couple of notes:
* The default `media` class is provided by negating all the boolean properties. It's an expression that will need to be updated if there are any new media properties.
* Currently I have separated `Media` (body/left/right/heading/object) from `MediaList`. The reason is that `MediaList` has special handling for children (setting the tag to be `li`).

src/MediaList.jsx
```diff
@@ -0,0 +1,45 @@
...
+  const classes = classNames(
+    className,
+    'media-list'
+  );
+
+  return (
+    <Tag {...attributes} className={classes}>
+      {React.Children.toArray(children).map((child, i) => {
+        return React.cloneElement(child, {
+          key: i,
```
eddywashere: Not sure about `childTag`. Would it be better to push people to add `<Media tag="li"/>`. This seems a little strict by cloning each element inside of MediaList.

eddywashere:
@mking-clari thanks a ton for this pr and all the updates. I wish I would have had time to do the feedback all at once. I have that one last comment about `childTag`, let me know what you think. otherwise this looks good to go.

mking-clari:
@eddywashere thanks for the feedback, I think you've been really helpful and responsive, especially compared to many other github maintainers. I'm happy to contribute a small part to an active repository. The code should be updated with the latest change.

eddywashere:
good call 👍

PR #95, opened 2016-08-09
Make sure size property of modal has effect.

tvogels:
Currently the size property of `Modal` does nothing. This adds modal-size to the className if a size is specified.


eddywashere:
Thanks @tvogels!

PR #99, opened 2016-08-16
feat(inputGroup): add InputGroup

### LINKED ISSUE #74: InputGroup
http://v4-alpha.getbootstrap.com/components/input-group/

- [ ] InputGroup
- [ ] InputGroupAddon
- [ ] InputGroupButton or alternative (ButtonDropdown?)

TheSharpieOne:
Resolves #74
Add `InputGroup` with `size` prop
Add `InputGroupAddon`
Change Input to add `addon` prop for radio/checkbox to not add `form-check-input` class
Add `InputGroupButton` with convinence shorthand when only a single string is passed
Add tests for all of the above
Add documention

### DIFF
--- a/src/InputGroupButton.jsx
+++ b/src/InputGroupButton.jsx
@@ -1,0 +1,53 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+import Button from './Button';
+
+const propTypes = {
+  tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
+  children: PropTypes.node,
+  groupClassName: PropTypes.any,
+  groupAttributes: PropTypes.object,
+  className: PropTypes.any
+};
+
+const defaultProps = {
+  tag: 'div'
+};
+
+const InputGroupButton = (props) => {
+  let {
+    className,
+    tag: Tag,
+    children,
+    groupClassName,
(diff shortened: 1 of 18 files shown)


src/InputGroupButton.jsx
```diff
@@ -0,0 +1,53 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+import Button from './Button';
+
+const propTypes = {
+  tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
+  children: PropTypes.node,
+  groupClassName: PropTypes.any,
```
eddywashere: Been thinking about these one off props. They seem fine for now. Do you think that something like `tags`, `props` and `classNames` would help define a common pattern for this kind of stuff? ex: `classNames={{group, button}}`, `props={{group: groupAttrs}}`.

eddywashere:
:shipit:

src/InputGroupButton.jsx
TheSharpieOne: Yeah, definitely. I like the `props` since `className` is just a prop, one could do `props={{group: {className: 'yo'}}}` to get that there without a special prop just for it.
I can change this in a future PR to match that pattern if you want to go that way.

PR #100, opened 2016-08-16
fix(tests): fix coverage for Modal tests

TheSharpieOne:
Fixes code branch coverage not testing 100%. Size tests were missing from #95.

Turns this:
image
into this:
image

(now all 100%)


eddywashere:
Good catch. Thanks for getting this back to 100%!

PR #101, opened 2016-08-16
chore(coverage): Update instrumenter

TheSharpieOne:
Replace isparta with istambul since isparta is no longer maintained
See: https://github.com/douglasduteil/isparta
Istambul now support es2015/6.
(Also added cross-env to allow it to work on windows)

This PR is not a big deal, just some maintenance, though I did notice the coverage output was _slightly_ more detailed with the babel istambul plugin this PR adds than the isparta loader it replaces.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,9 +44,7 @@
   },
   "homepage": "https://github.com/reactstrap/reactstrap#readme",
   "dependencies": {
-    "babel-cli": "^6.11.4",
     "classnames": "^2.2.3",
-    "cross-env": "^2.0.0",
     "lodash.isfunction": "^3.0.8",
     "lodash.omit": "^4.4.1",
     "tether": "^1.3.4"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
   "dependencies": {
     "babel-cli": "^6.11.4",
     "classnames": "^2.2.3",
+    "cross-env": "^2.0.0",
```
eddywashere: I know it's weird, but I was hoping to restrict `dependencies` to the core reactstrap library dependencies. I must have accidently added `babel-cli` to it.

eddywashere:
I think I should restrict tests to latest node version to avoid the double coverage reports.

TheSharpieOne:
Yeah, not too sure why it is not about to determine the changes to the coverage either and the branches are "**"... strange.

package.json
TheSharpieOne: will do

package.json
```diff
@@ -56,8 +55,10 @@
     "react-dom": "^0.14.0 || ^15.0.0"
   },
   "devDependencies": {
+    "babel-cli": "^6.11.4",
```
eddywashere: 🙌  thanks!

Issue #77, opened 2016-07-31
Pagination


eddywashere:
http://v4-alpha.getbootstrap.com/components/pagination/

- [x] Pagination
- [x] PaginationItem
- [x] PaginationLink

PR #108, opened 2016-08-20
feat(Pagination): add Pagination component

mking-clari:
References #77

* Add `Pagination`, `PaginationItem`, `PaginationLink`
* `Pagination` supports sizing
* `PaginationItem` supports active and disabled states
* `PaginationLink` supports special rendering for next and previous links

### DIFF
--- a/src/PaginationLink.jsx
+++ b/src/PaginationLink.jsx
@@ -1,0 +1,78 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+const propTypes = {
+  'aria-label': PropTypes.string,
+  children: PropTypes.node,
+  className: PropTypes.string,
+  next: PropTypes.bool,
+  previous: PropTypes.bool,
+  tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
+};
+
+const defaultProps = {
+  tag: 'a',
+};
+
+const PaginationLink = (props) => {
+  const {
+    className,
+    next,
+    previous,
+    tag: Tag,
+    ...attributes,
+  } = props;
+
+  const classes = classNames(
+    className,
+    'page-link'
+  );
+
+  let defaultAriaLabel;
+  if (previous) {
+    defaultAriaLabel = 'Previous';
+  } else if (next) {
+    defaultAriaLabel = 'Next';
+  }
+  const ariaLabel = props['aria-label'] || defaultAriaLabel;
+
+  let defaultCaret;
+  if (previous) {
+    defaultCaret = '\u00ab';
+  } else if (next) {
+    defaultCaret = '\u00bb';
+  }
+
+  let children = props.children;
+  if (previous || next) {
+    children = [
+      <span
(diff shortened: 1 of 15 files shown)


src/PaginationLink.jsx
```diff
@@ -0,0 +1,43 @@
...
+    ...attributes,
+  } = props;
+
+  const classes = classNames(
+    className,
+    'page-link'
+  );
+
+  return (
+    <Tag {...attributes} className={classes}>
```
eddywashere: Can you add the aria label, the aria hidden attr and sr-only label for accessibility?

```
<a class="page-link" href="#" aria-label="Next">
    <span aria-hidden="true">&raquo;</span>
    <span class="sr-only">Next</span>
</a>
```

eddywashere: For prev/next, I don't think it will ever require children unless someone wants to override the carets. Could you setup rendering to be: if prev/next, include spans and render children instead of carets vs rendering children before the prev/next content.

eddywashere:
A few comments, let me know what you think. Great work!

mking-clari:
@eddywashere Thanks for the feedback, I updated the code.

Some notes:
* If previous or next is set, the children prop replaces just the caret, not all the spans. If the user wants full customization of the spans, they can just not set previous or next.
* The `aria-label` prop is used for both the `aria-label` attribute and the `.sr-only` text.

src/PaginationLink.jsx
```diff
@@ -0,0 +1,67 @@
...
+  if (previous) {
+    defaultAriaLabel = 'Previous';
+  } else if (next) {
+    defaultAriaLabel = 'Next';
+  }
+  const ariaLabel = props['aria-label'] || defaultAriaLabel;
+
+  let defaultCaret;
+  if (previous) {
+    defaultCaret = <span>&laquo;</span>;
```
eddywashere: I don't think these extra spans are needed anymore since it's taken care of with `<span aria-hidden="true">{children || defaultCaret}</span>`.

src/PaginationLink.jsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  return (
+    <Tag
+      {...attributes}
+      className={classes}
+      aria-label={ariaLabel}
+    >
+      {(previous || next) ?
+        (<span>
```
eddywashere: extra span here too. you could push the child content to an array to get around the need for the wrapper.

src/PaginationLink.jsx
eddywashere: I don't remember if `&laquo;` as a string will render html, might need html entity.

mking-clari:
@eddywashere For the `&laquo;` issue, I replaced the html entity with the unicode character (`\u00ab`), which should [avoid the escaping issue](https://facebook.github.io/react/docs/jsx-gotchas.html). Also updated the test for consistency.

For the span issue, I added an array (keys required to avoid warning).

eddywashere:
👍

Issue #110, opened 2016-09-01
Namespace modal with custom css class for styling


ajainarayanan:
### Issue description
Have custom css class for modal-dialog to allow namespac'ing for modal styling.

- version `#2.6.0`
- components: `Modal`

Right now if a user needs to style modals (say lg modal) it has to be a global css styling, as the modal is attached to the body. 

User should be able to have a different styles to be applied for different modals. This could be achieved by enabling the user to provide `className` on `<Modal>` component. This way user should be able to style each modal in a different way if need be with a top level css class.

#### Existing DOM representation:
```
<div>
  <Button color="danger" onClick={this.toggle}>Launch Modal</Button>
  <Modal isOpen={this.state.modal} toggle={this.toggle}>
    <ModalHeader toggle={this.toggle}>Modal title</ModalHeader>
    <ModalBody>
      Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
    </ModalBody>
    <ModalFooter>
      <Button color="primary" onClick={this.toggle}>Do Something</Button>
      <Button color="secondary" onClick={this.toggle}>Cancel</Button>
    </ModalFooter>
  </Modal>
</div>
```

#### Suggested change:
```
<div>
  <Button color="danger" onClick={this.toggle} className="my-custom-modal">Launch Modal</Button>
  <Modal isOpen={this.state.modal} toggle={this.toggle}>
    <ModalHeader toggle={this.toggle}>Modal title</ModalHeader>
    <ModalBody>
      Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
    </ModalBody>
[...]

PR #111, opened 2016-09-01
feat(ModalClassName): Adds custom class name for modal-dialogs

ajainarayanan:
Related Issue: #110

- Users should be able to style modals and should be able to namespace the same.
- This will help us have styling specific for modals and the rest be generic.


eddywashere:
Thanks @ajainarayanan! will release this evening

PR #120, opened 2016-09-06
Button & DropdownToggle updates

### LINKED ISSUE #98: Curious behaviour with DropdownToggle
It seems that there's a [particular behaviour](https://github.com/reactstrap/reactstrap/blob/master/src/DropdownToggle.jsx#L65-L73) that tells the DropdownToggle component (and maybe others) to use the children as "master button" if there's a single one, instead of creating a new one. I feel like it's a mistake:

- It's not an intuitive behaviour - without looking at the source code, one might wonder what happens (I sure did)

- The `tag` prop already allows us to customize the way the button is rendered, and is much more idomatic than adding props to already-rendered components.

- It makes it quite harder to use icons inside a DropdownToggle (because the icon props are rewritten, so the classNames are lost). For example, the following doesn't work:

    ```js
    <DropdownToggle>
        <i className={`fa fa-user`} />
    </DropdownToggle>
    ```

    One has to write the following:

    ```js
    <DropdownToggle>
        <i className={`fa fa-user`} /> <span></span>
    </DropdownToggle>
    ```

    In order to bypass the check.

- Additional issues may arise (for example, without testing, I guess that such a pattern also breaks refs, and triggers more renders than required).

Would you consider removing this behaviour?

eddywashere:
This fixes some inconsistencies within button and dropdown components and removes some weird behavior found in #98.

- Button color defaults to `secondary` (previously default style in bs3)
- DropdownToggle: no longer clones each element in props.children when it’s an array, instead it renders props.children inside a single component (Button) with the onClick toggle context.

### DIFF
--- a/docs/lib/Components/DropdownsPage.js
+++ b/docs/lib/Components/DropdownsPage.js
@@ -1,5 +1,6 @@
 /* eslint react/no-multi-comp: 0, react/prop-types: 0 */
 import React from 'react';
+import { Link } from 'react-router';
 import { PrismCode } from 'react-prism';
 import Helmet from 'react-helmet';
 import {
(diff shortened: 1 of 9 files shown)


docs/lib/Components/DropdownsPage.js
```diff
@@ -32,6 +33,13 @@ export default class DropdownPage extends React.Component {
       <div>
         <Helmet title="Dropdowns" />
         <h3>Dropdowns</h3>
+        <p>
+          The <code>Dropdown</code> component is used to pass the <code>isOpen</code> & <code>toggle</code> props via context to the following components: <code>DropdownToggle</code>, <code>DropdownMenu</code>. The <code>DropdownToggle</code> uses the <code>Button</code> component internally, meaning it also accepts all the props the <Link to="/components/buttons/">Button component</Link> accepts.
```
eddywashere: Also added Dropdown description and info around advanced positioning.

eddywashere:
Besides the Button color deafult, this affects users who pass an array of children to `DropdownToggle`. 

Old example:

```jsx
<DropdownToggle>
  <Button color="primary">Click Me</Button>
  <Button color="primary">Or me!!</Button>
</DropdownToggle>
```

Updated syntax:

```jsx
<DropdownToggle color="primary">Click Me</DropdownToggle>
<DropdownToggle color="primary">Or me!!</DropdownToggle>
```

You could think of `DropdownTogggle` as an enhanced `Button` that has `onClick` set to use the toggle method passed to the parent `Dropdown` component (uses context).

eddywashere:
@arcanis thoughts?

The following will now work

```jsx
<DropdownToggle>
    <i className={`fa fa-user`} />
</DropdownToggle>
```

this too:

```jsx
<DropdownToggle tag="i" className={`fa fa-user`} />
```

arcanis:
Looks much better :+1:

PR #127, opened 2016-09-07
fix(modal): fix event listener for modal

### LINKED ISSUE #126: conditional modal footer closes modal
Is there a way to conditionally display/toggle the modal footer without causing the modal to close?

My use case establishes the modal footer as conditional when there are files selected i.e I display buttons to remove files. I'm running into the issue that while I can make the modal footer conditionally appear, the case in which the footer should disappear causes the entire modal to close.

Take a look at the following bin sample, toggle the modal and add a file. When you remove files the entire modal closes.
http://www.webpackbin.com/NkyVH5_jb

TheSharpieOne:
Fixes #126
Change the event listeners for click and keyup in modal
to set `useCapture` to true to trigger the handler before
any other handler.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -100,7 +100,7 @@ class Modal extends React.Component {
   }
 
   removeEvents() {
-    document.removeEventListener('click', this.handleBackdropClick, false);
+    document.removeEventListener('click', this.handleBackdropClick, true);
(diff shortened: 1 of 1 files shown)


eddywashere:
I did not now about useCapture other than it was safer to use false. Makes way more sense now. Thanks for looking into this and fixing it. 🙌

PR #128, opened 2016-09-07
fix(TetherContent): rerender when other props change

### LINKED ISSUE #125: PopoverContent not re-rendering
If you edit the props for PopverContent, and it's already open, it doesn't re-render the content until closed and opened again.

Take a look at the following codepen... just open one and let it sit for a few seconds... 
http://codepen.io/zeroasterisk/pen/zKxgrJ

You can see in the console.log that the state --> props are updating, but the UI is not.  If you close and re-open the popover, it shows the new/current props.

TheSharpieOne:
Fix #125
Let the TetherContent re-render when any props change while it is
showing. This allows the children to update dynamically without
needing to hide and show the content to force the update.

This also includes tests, but no additional docs as this is not a feature, but something that should just work this way.
If you would like me to add to the docs to document this, let me know and I will adjust this PR.

### DIFF
--- a/src/TetherContent.js
+++ b/src/TetherContent.js
@@ -32,6 +32,9 @@ class TetherContent extends React.Component {
   componentDidUpdate(prevProps) {
     if (this.props.isOpen !== prevProps.isOpen) {
       this.handleProps();
+    } else if (this._element) {
+      // rerender
+      this.renderIntoSubtree();
     }
   }
 
@@ -94,11 +97,7 @@ class TetherContent extends React.Component {
 
     this._element = document.createElement('div');
     document.body.appendChild(this._element);
-    ReactDOM.unstable_renderSubtreeIntoContainer(
-      this,
-      this.renderChildren(),
-      this._element
-    );
+    this.renderIntoSubtree();
 
     if (this.props.arrow) {
       const arrow = document.createElement('div');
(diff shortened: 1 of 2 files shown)


test/TetherContent.spec.js
```diff
@@ -81,14 +81,14 @@ describe('TetherContent', () => {
...
+    it('should be called on componentDidMount', () => {
       state = true;
-      spyOn(TetherContent.prototype, 'componentWillUnmount').and.callThrough();
+      spyOn(TetherContent.prototype, 'componentDidMount').and.callThrough();
       spyOn(TetherContent.prototype, 'show').and.callThrough();
       const wrapper = mount(<TetherContent tether={tetherConfig} isOpen={state} toggle={toggle}><p>Content</p></TetherContent>);
       const instance = wrapper.instance();
 
-      expect(TetherContent.prototype.componentWillUnmount.calls.count()).toBe(0);
+      expect(TetherContent.prototype.componentDidMount.calls.count()).toBe(1);
```
TheSharpieOne: I threw this in here, the test didn't make any sense saying `show` should be called on unmount. Seemed like a copy paste from the one above it. Now this tests to make sure the show is called when the component did mount.

eddywashere: Awesome!

src/TetherContent.js
```diff
@@ -119,6 +118,14 @@ class TetherContent extends React.Component {
     return this.props.toggle();
   }
 
+  renderIntoSubtree() {
```
eddywashere: 👌

eddywashere:
Thanks @TheSharpieOne

Issue #72, opened 2016-07-31
Tabs


eddywashere:
http://v4-alpha.getbootstrap.com/components/navs/#tabs

- [x] tab option to Nav
- [x] TabContent
- [x] TabPane

ajainarayanan:
@eddywashere Is there a semantic example you are following in implementing this component? Can you give an example to show how this component will eventually be used?

eddywashere:
Currently I've been getting by without tab specific components by using Nav and determining active content via route or other method.

```jsx
<div>
  <Nav tabs>
    <NavItem>
      <NavLink href="/1">Link</NavLink>
    </NavItem>
    <NavItem>
      <NavLink href="/2">Another Link</NavLink>
    </NavItem>
  </Nav>
 {this.props.children || activeContent}
</div>
```

Eventually I'd see a more controlled example with Tab helpers: TabContent and TabPane. A toggle method would update state to reflect active tab. TabContent would pass that through to TabPane. Here's a rough example:

```jsx
class Example extends React.Component {
  constructor(props) {
    super(props);

    this.toggle = this.toggle.bind(this);
    this.state = {
      activeTab: '1'
    };
  }

  toggle(tab) {
    if (this.state.activeTab !== tab) {
      this.setState({
        activeTab: tab
      });
    }
  }

  render() {
    return (
      <div>
        <Nav tabs>
          <NavItem>
            <NavLink onClick={() => {this.toggle('1')}}>Link</NavLink>
          </NavItem>
          <NavItem>
            <NavLink onClick={() => {this.toggle('2')}}>Another Link</NavLink>
          </NavItem>
        </Nav>
        <TabContent activeTab={this.state.activeTab}>
          <TabPane name="1">
            <p>tab content</p>
          </TabPane>
          <TabPane name="2">
            <p>tab content</p>
          </TabPane>
        </TabContent>
      </div>
    );
  }
}
```

eddywashere:
Shipped in 3.1.0

ukiyakimura:
@eddywashere : 
i got some problem, everytime i used <TabContent> and <TabPane>. it always show this error 

`warning.js:36Warning: React.createElement: type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components). Check the render method of `SetupPage`.`

[...]

PR #131, opened 2016-09-08
Feat: Initial implementation of Tabs component

ajainarayanan:
PR for #72 

##### Note:
- Not sure if I adhered to coding style. Please let me know if I missed something
- Will squash all the commits if the changes are ok
- Have added some comments, places where I thought needed some explanation. Please feel free to remove them if you think its not right.
- If someone can point me how to update the test coverage would make sure it is 100%

### DIFF
--- a/src/TabContent.js
+++ b/src/TabContent.js
@@ -1,9 +1,14 @@
 import React, { PropTypes, Component } from 'react';
+import classnames from 'classnames';
 
 const propTypes = {
   children: PropTypes.node,
   activeTab: PropTypes.any,
   className: PropTypes.string
+};
+
+const childContextTypes = {
+  activeTabId: PropTypes.any
 };
 
 export default class TabContent extends Component {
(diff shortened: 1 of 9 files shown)


eddywashere:
Couple of comments

I think this could use the .tab-content, .tab-pane classes and .active from http://v4-alpha.getbootstrap.com/components/navs/#using-data-attributes (fade option could be part of a different pr). I should have mentioned that in GitHub issue.

Also, instead of iterating through children and setting state, you could use context to simplify the TabContent component https://facebook.github.io/react/docs/context.html.

Ex: add getChildContext to TabContent, return {activeTab: this.props.activeTab}. In TabPane, it can accept a 2nd argument, context, from there you can compare props.tabId to context.activeTab and add the .active class if they are equal.

For coverage, there's a folder in test directory called coverage and somewhere inside is a root index.html. I usually open that in a browser and sort by missing coverage

Thanks for tackling this! feel free to ping me here or in slack if you need anything. The link to it is in contributing.md

ajainarayanan:
@eddywashere Will work on this weekend. Just one minor question, react docs mentions using context only when it is absolutely necessary and try to avoid it as much as possible. Is there any specific reason why we should use contextTypes for communicating with children? The props seemed to be simple. Is there any specific reason we should use contextTypes? (Sorry if my question is a little silly, I am relatively new to react :))

eddywashere:
Glad you brought that up. "Using context makes your components more coupled" I interpret that as a positive effect in this case. TabPanes are coupled to TabContent. It's partially to save time passing the same prop to child components. If the API changes we could always switch to cloning children(TabPanes) with the activeTab prop from TabContent. That wouldn't cause a breaking change and seems like an easy win for keeping things simple.

ajainarayanan:
@eddywashere That was super useful! Have updated the PR. Please review and let me know I have missed anything.

src/TabContent.js
```diff
@@ -0,0 +1,41 @@
...
+      activeTab: this.props.activeTab
+    };
+  }
+  getChildContext() {
+    return {
+      activeTabId: this.state.activeTab
+    };
+  }
+  componentWillReceiveProps(nextProps) {
+    this.setState({
```
eddywashere: you should check if `nextProps.activeTab !== this.props.activeTab` before using setState

src/TabContent.js
```diff
@@ -0,0 +1,41 @@
...
+      activeTab: nextProps.activeTab
+    });
+  }
+  render() {
+    return (
+      <div className={classnames('tab-content', this.props.className)}>
```
eddywashere: consider moving `classnames('tab-content', this.props.className)` to a const called classes. It's not across all components but trying to standardize that.

src/TabPane.js
```diff
@@ -0,0 +1,22 @@
...
+  className: PropTypes.string,
+  tabId: PropTypes.any
+};
+const contextTypes = {
+  activeTabId: PropTypes.any
+};
+
+export default function TabPane({ className, tabId, children }, context) {
+  return (
+    <div className={classnames('tab-pane', className, { active: tabId === context.activeTabId })}>
```
eddywashere: same classes const here

src/TabContent.js
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const childContextTypes = {
+  activeTabId: PropTypes.any
+};
+
+export default class TabContent extends Component {
+  constructor(props) {
+    super(props);
+    this.childRefs = {};
```
eddywashere: this can be removed now

src/TabPane.js
```diff
@@ -0,0 +1,22 @@
...
+const propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  tabId: PropTypes.any
+};
+const contextTypes = {
+  activeTabId: PropTypes.any
+};
+
+export default function TabPane({ className, tabId, children }, context) {
```
eddywashere: could you move these props into something like the pattern in other components, [example](https://github.com/reactstrap/reactstrap/blob/af874bc9001be3bf692a998714ebe7aa4f3d7bbb/src/Card.js#L18-L26).

This makes it easier to pass through attributes/props that aren't explicitly called out.

src/TabContent.js
ajainarayanan: Not sure why github is not collapsing or updating this comment but this is fixed

eddywashere:
thanks a ton for working through this @ajainarayanan! I'll squash and merge via github.

test/Tabs.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    tab1.find('.nav .nav-item .nav-link').at(1).simulate('click');
+    expect(tab1.find('.nav .nav-item .nav-link').at(0).hasClass('active')).toBe(false);
+    expect(tab1.find('.nav .nav-item .nav-link').at(1).hasClass('active')).toBe(true);
+    expect(tab1.find('.tab-content .tab-pane').at(1).hasClass('active')).toBe(true);
+    tab1.find('.nav .nav-item .nav-link').at(0).simulate('click');
+  });
+  it('should show no active tabs if active tab id is unknown', () => {
+    let tab1 = mount(<TabsExample />);
+    const instance = tab1.instance();
+    expect(instance instanceof TabsExample).toBe(true);
```
TheSharpieOne: are you testing the example?

ajainarayanan: The TabExample is a wrapper around the Tab. The aim in this test is to have a click initiated from the wrapper that toggles the TabContent (in this case an invalid case). The clicking and managing ActiveTab is not in the scope of the Tab component. The test still focuses on testing the Tab and not the TabsExample though.

TheSharpieOne: It's just this except tests the example directly, seems pointless.

PR #135, opened 2016-09-13
feat(modal): add backdrop and keyboard options

### LINKED ISSUE #134: Optionally hide modal backdrop in Modal
### Issue description
- While using Modal it would be nice if Modal has an option to hide the backdrop. Right now it is compulsorily shown and there is no way to configure the Modal to not show the backdrop.

Could be similar to [bootstrap v4 modal](http://v4-alpha.getbootstrap.com/components/modal/#options)
- version `3.1.0`
- components: `Modal`

TheSharpieOne:
Closes #134
Add backdrop prop to modal with the following options
 - `true` show backdrop, close on click (default)
 - `false` hide backdrop
 - `"static"` show backdrop, do not close on click

Add keyboard prop to modal with the following options
 - `true` close on escape (default)
 - `false` do not close on escape

Add tests and docs

### DIFF
--- a/src/TabContent.js
+++ b/src/TabContent.js
@@ -34,7 +34,7 @@ export default class TabContent extends Component {
     const classes = classnames('tab-content', this.props.className);
     return (
       <div className={classes}>
-        { this.props.children }
+        {this.props.children}
       </div>
     );
   }
--- a/docs/lib/Components/ModalsPage.js
+++ b/docs/lib/Components/ModalsPage.js
@@ -5,6 +5,9 @@ import Helmet from 'react-helmet';
 import ModalExample from '../examples/Modal';
 const ModalExampleSource = require('!!raw!../examples/Modal');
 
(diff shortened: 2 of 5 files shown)


src/TabContent.js
```diff
@@ -34,7 +34,7 @@ export default class TabContent extends Component {
     const classes = classnames('tab-content', this.props.className);
     return (
       <div className={classes}>
-        { this.props.children }
```
TheSharpieOne: This was given me a linter error when running the tests.

ajainarayanan:
@TheSharpieOne That was super quick! Thanks!

src/TabContent.js
ajainarayanan: For some reason it didn't throw me this error. Not sure if @eddywashere saw this when he release 3.1.0.

eddywashere: missed it, maybe it needs option to exit properly in build step

eddywashere:
great additions!

PR #136, opened 2016-09-13
fix(TabContent): fix TabContent test coverage

TheSharpieOne:
Adds additional test to bring coverage back up to 100%

Also, just throwing this in here. I do not like the way the tests were written for the Tab components. Testing is supposed to improve code confidence, but I do not feel these tests did that for me.

### DIFF
--- a/src/TabContent.js
+++ b/src/TabContent.js
@@ -34,7 +34,7 @@ export default class TabContent extends Component {
     const classes = classnames('tab-content', this.props.className);
     return (
       <div className={classes}>
-        { this.props.children }
+        {this.props.children}
       </div>
     );
   }
--- a/test/Tabs.spec.js
+++ b/test/Tabs.spec.js
@@ -1,10 +1,7 @@
 /* eslint react/no-multi-comp: 0, react/prop-types: 0 */
 import React from 'react';
-import ReactDOM from 'react-dom';
-
 import { mount } from 'enzyme';
-import { Nav, NavItem, NavLink, TabContent, TabPane } from 'reactstrap';
-import classnames from 'classnames';
+import { TabContent } from 'reactstrap';
 // Not sure if this is correct but didn't want to repeat a whole bunch of code.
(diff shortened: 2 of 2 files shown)


src/TabContent.js
```diff
@@ -34,7 +34,7 @@ export default class TabContent extends Component {
     const classes = classnames('tab-content', this.props.className);
     return (
       <div className={classes}>
-        { this.props.children }
```
TheSharpieOne: Had to make the linter happy in order to run test locally. Not sure why no one else is getting this (including Travis) when running the tests.

ajainarayanan:
@TheSharpieOne Can you elaborate on what you think is the right way to do the same? Just trying to find what I am missing here, did I miss any scenario? When I had merged the PR for Tab component the coverage was 100% for me.

TheSharpieOne:
I would test the individual components, mocking out everything else. Right now, just running the example in the test will give 99.5% test coverage, even though nothing is tested. By mocking out everything you are not testing, it ensures that the thing you are testing work directly. It also means coverage only hits on the stuff that is tested and not things that just happen to run.
It also helps so that things do not get in the way. I couldn't use the example to test the piece that was missing. I had to use the TabContent directly, and it wasn't hard or any more work really.

ajainarayanan:
@TheSharpieOne I see. Ok. Will fix the tests probably tonight. Sorry about the change in coding style.

test/Tabs.spec.js
```diff
@@ -1,10 +1,7 @@
 /* eslint react/no-multi-comp: 0, react/prop-types: 0 */
 import React from 'react';
-import ReactDOM from 'react-dom';
-
 import { mount } from 'enzyme';
-import { Nav, NavItem, NavLink, TabContent, TabPane } from 'reactstrap';
```
eddywashere: ah, totally missed the unused deps in these tests. Currently eslint only lints the main src directory. Thanks!

PR #138, opened 2016-09-13
feat(Modal): allow for multiple modals

### LINKED ISSUE #137: Attach clickHandler on backdrop element instead of document
### Issue description
Bumped onto a scenario where I need to show one Modal on top of another one and on click on second Modal(anywhere) the first Modal gets closed.

The reason for this is the Modal component attaches the click event to document instead of backdrop element.

- version `#3.1.0`
- components: `Modal`

### Steps to reproduce issue
- Open a Modal
- Have a button in the Modal that will open Another Modal
- Click anywhere in the Second Modal (opened on top of the first one).

CodePen: http://codepen.io/anon/pen/amdaGw

#### Note:
In my case the first Modal gets closed and the second Modal becomes un-responsive. However I couldn't reproduce this in codepen.

EDIT: Removed an error screenshot. Could be based on my usage.

TheSharpieOne:
Resolves #137
Moves the click and keyup listeners from the document to the
`.modal` div, which wraps the modal-dialog. This allows for each
modal to be individually triggered to close.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -88,7 +88,6 @@ class Modal extends React.Component {
 
   destroy() {
     const classes = document.body.className.replace('modal-open', '');
-    this.removeEvents();
 
     if (this._element) {
       ReactDOM.unmountComponentAtNode(this._element);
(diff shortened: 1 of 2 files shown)


ajainarayanan:
Tested it locally. Worked as expected 👍

src/Modal.js
```diff
@@ -151,6 +142,8 @@ class Modal extends React.Component {
             transitionAppearTimeout={300}
             transitionEnterTimeout={300}
             transitionLeaveTimeout={300}
+            onClickCapture={this.handleBackdropClick}
```
eddywashere: TIL about onClickCapture 👍

eddywashere:
:shipit:

PR #141, opened 2016-09-14
fix(TabContent,TabPane): Fixes TabContent & Tabpane tests

ajainarayanan:
- Removes `<TabsExample />` (which requires no test) and mocks everything possible.
- Tests TabContent & TabPane components alone and nothing else.

@TheSharpieOne Let me know if I misunderstood anything wrong from yesterday. Will be happy to fix it.

### DIFF
--- a/test/Tabs.spec.js
+++ b/test/Tabs.spec.js
@@ -1,35 +1,53 @@
 import React from 'react';
 import { mount } from 'enzyme';
-import { TabContent } from 'reactstrap';
-// Not sure if this is correct but didn't want to repeat a whole bunch of code.
-import TabsExample from '../docs/lib/examples/Tabs';
+import { TabContent, TabPane } from 'reactstrap';
 
+let activeTab = '1';
 describe('Tabs', () => {
   it('should render', () => {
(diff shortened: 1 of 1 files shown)


TheSharpieOne:
Coverage went down: [this line](https://coveralls.io/builds/7884799/source?filename=src%2FTabContent.js#L28) never gets hit, when means in all of the tests, the if statement is always false, and thus the true path is never taken/tested.

I was add a new test similar to [the one I added](https://github.com/reactstrap/reactstrap/blob/master/test/Tabs.spec.js#L36-L42), but instead of not changing the active tab and expecting the spy to have not been called. Change the active tab using setProps and expect the spy to have been called.

ajainarayanan:
@TheSharpieOne ok cool. Got it. Will do sometime today.

ajainarayanan:
@TheSharpieOne Sorry was a little busy at work place. Updated the PR with appropriate test case (coverage is now 100% I believe as far as I checked it locally). Let me know if I missed anything

eddywashere:
👍

PR #143, opened 2016-09-15
feat: add delay prop to Tooltip

### LINKED ISSUE #115: Tooltip timeout option
The current **Tooltip** component has some fixed timeout which decides how long the tooltip remains visible after a mouseout event.

https://github.com/reactstrap/reactstrap/blob/master/src/Tooltip.jsx#L63

It would be nice to have the option of changing this via an option prop.

I think the usual use case (also the case for me) is not wanting any delay at all - i.e. timeout: 0.
Unless there is some specific reason why the fixed timeout exists?

alisd23:
Previous delay was a harcoded value of 250ms for the *hide*, and no delay for the
*show*.
New prop `delay` allows for either an object of form:
`{ show: 100, hide: 200 }` or simply a number to set these delays.
Default is `{ show: 0, hide: 250 }`

Closes #115

### DIFF
--- a/docs/lib/Components/TooltipsPage.js
+++ b/docs/lib/Components/TooltipsPage.js
@@ -34,6 +34,8 @@ export default class TooltipsPage extends React.Component {
   // target div ID, popover is attached to this element
   tether: PropTypes.oneOfType([PropTypes.object, PropTypes.bool]),
   // optionally overide tether config http://tether.io/#options
+  delay: PropTypes.oneOfType([PropTypes.object, PropTypes.number]),
+  // optionally override show/hide delays - default { show: 0, hide: 250 }
   placement: PropTypes.oneOf([
     'top',
     'bottom',
--- a/docs/lib/Components/TooltipsPage.js
+++ b/docs/lib/Components/TooltipsPage.js
@@ -34,7 +34,10 @@ export default class TooltipsPage extends React.Component {
   // target div ID, popover is attached to this element
   tether: PropTypes.oneOfType([PropTypes.object, PropTypes.bool]),
   // optionally overide tether config http://tether.io/#options
-  delay: PropTypes.oneOfType([PropTypes.object, PropTypes.number]),
+  delay: PropTypes.oneOfType([
+    PropTypes.shape({ show: PropTypes.number.isRequired, hide: PropTypes.number.isRequired }),
+    PropTypes.number
+  ]),
   // optionally override show/hide delays - default { show: 0, hide: 250 }
   placement: PropTypes.oneOf([
     'top',
--- a/docs/lib/Components/TooltipsPage.js
+++ b/docs/lib/Components/TooltipsPage.js
@@ -35,7 +35,7 @@ export default class TooltipsPage extends React.Component {
   tether: PropTypes.oneOfType([PropTypes.object, PropTypes.bool]),
   // optionally overide tether config http://tether.io/#options
   delay: PropTypes.oneOfType([
-    PropTypes.shape({ show: PropTypes.number.isRequired, hide: PropTypes.number.isRequired }),
+    PropTypes.shape({ show: PropTypes.number, hide: PropTypes.number }),
     PropTypes.number
   ]),
   // optionally override show/hide delays - default { show: 0, hide: 250 }
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -10,7 +10,10 @@ const propTypes = {
   tether: PropTypes.object,
   toggle: PropTypes.func,
   children: PropTypes.node,
-  delay: PropTypes.oneOfType([PropTypes.object, PropTypes.number])
+  delay: PropTypes.oneOfType([
+    PropTypes.shape({ show: PropTypes.number.isRequired, hide: PropTypes.number.isRequired }),
+    PropTypes.number
+  ])
 };
 
 const DEFAULT_DELAYS = {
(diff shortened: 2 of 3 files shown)


src/Tooltip.js
```diff
@@ -9,12 +9,19 @@ const propTypes = {
   disabled: PropTypes.bool,
   tether: PropTypes.object,
   toggle: PropTypes.func,
-  children: PropTypes.node
+  children: PropTypes.node,
+  delay: PropTypes.oneOfType([PropTypes.object, PropTypes.number])
```
TheSharpieOne: Why not use PropTypes.shape to define the object instead of just the generic object since you are mandating that show and hide have to be a number in `getDelay`. It avoids the `NaN` check and would give developers a specific message when they don't use the right shape.

src/Tooltip.js
```diff
@@ -50,23 +58,25 @@ class Tooltip extends React.Component {
...
+    if (this._showTimeout) {
+      clearTimeout(this._showTimeout);
+    }
+    this._hideTimeout = setTimeout(this.hide, this.getDelay('hide'));
   }
 
-  onTimeout() {
-    if (this.props.isOpen) {
-      this.toggle();
+  getDelay(key) {
```
TheSharpieOne: Instead of running this on every mouse over and mouse leave, why not do this when it mounts and each time props update. It will help avoid overhead.

src/Tooltip.js
```diff
@@ -79,10 +89,24 @@ class Tooltip extends React.Component {
...
+
   handleDocumentClick(e) {
     if (e.target === this._target || this._target.contains(e.target)) {
-      if (this._hoverTimeout) {
-        clearTimeout(this._hoverTimeout);
+      if (this._hideTimeout) {
+        clearTimeout(this._hideTimeout);
+      }
+      if (this._showTimeout) {
+        clearTimeout(this._showTimeout);
```
TheSharpieOne: You need to clear out `this.showTimeout` and `this.hideTimeout`. You can simply set them to `undefined`. Without clearing them out, they will still be truthy and thus, every click after the have been set will trigger the `clearTimeout`, even if the timeout has already been cleared.

TheSharpieOne: If fact, everywhere you use `clearTimeout` you will have this problem. It needs to clear the timeout and unset the var that contains the timeout id (`this._showTimeout` and `this.hideTimeout`).
Also, when the timeout triggers the function (`show`/`hide`) they need to clear the var containing their respective timeout id (`this._showTimeout` and `this.hideTimeout` respectfully)

src/Tooltip.js
alisd23: It's true that that would reduce some overhead, but I feel the tiny performance increase would not be worth complicating the code with extra lifecycle logic.

It's only a couple of direct accesses from the props and an if statement, and `mouseover` and `mouseleave` events aren't realistically going to happen very often unless the user is flailing the mouse around.

src/Tooltip.js
alisd23: Yeah good idea, I'll use shape instead. And I can add a couple of tests to check that the `hide` or `show` defaults are picked up :+1:

src/Tooltip.js
alisd23: Yeah that's true, it would run clearTimeout unnecessarily. That must have already been an issue before this PR though, because there was no unsetting of the timeout property either. But I may as well include that change in this PR just to tidy it up.

TheSharpieOne:
When `delay={{ show: 200 }}`, hide would be `undefined` from `getDelay`. This is probably not wanted.
Testing this is the test I was previously talking about.

alisd23:
Well now the propType for delay is `shape({ show: number, jhideL number })`, there will be a warning if the user doesn't supply both. Is that not enough? I think you were right before, it's better to tell the dev their input is wrong than hide it and choose a value for them I think.

TheSharpieOne:
I think the shape will ensure that there are no extra properties, but it will not ensure that the properties are there. You may have to add isRequired to ensure they are there. Something like http://stackoverflow.com/a/26924696/1873485

alisd23:
@TheSharpieOne Ok I think this is ready

eddywashere:
If `{ show: 0, hide: 250 }` is the default, I would assume setting `{hide: 0}` would result in: `{ show: 0, hide: 0 }`. It doesn't need to be part of this pr since adding it wouldn't introduce a breaking change.

test/Tooltip.spec.js
```diff
@@ -176,7 +176,41 @@ describe('Tooltip', () => {
...
+      );
+      const instance = wrapper.instance();
+
+      instance.onMouseLeaveTooltip();
+      expect(isOpen).toBe(true);
+      jasmine.clock().tick(200);
+      expect(isOpen).toBe(false);
+    });
+  });
+
```
eddywashere: [this comment](https://github.com/reactstrap/reactstrap/pull/143#issuecomment-247635912) mentions adding a test here for partial delay object. It's fine to add a test case that expects an error, and then fix in another pr or grab the value from the default object which I'd like to eventually be the way this works.

eddywashere:
thanks for working through this @alisd23 & @TheSharpieOne

test/Tooltip.spec.js
alisd23: Yep OK fair enough, it should be a quick fix so I'll just get it in this PR.

docs/lib/Components/TooltipsPage.js
```diff
@@ -34,6 +34,11 @@ export default class TooltipsPage extends React.Component {
   // target div ID, popover is attached to this element
   tether: PropTypes.oneOfType([PropTypes.object, PropTypes.bool]),
   // optionally overide tether config http://tether.io/#options
+  delay: PropTypes.oneOfType([
+    PropTypes.shape({ show: PropTypes.number.isRequired, hide: PropTypes.number.isRequired }),
```
TheSharpieOne: Tiny thing. Docs still showing it was required.

alisd23: Damn good catch

PR #149, opened 2016-09-19
Tooltip autohide prop

eddywashere:
Adds autohide prop to tooltip. When autohide is false, hovering over tooltip content will not close the tooltip.

example use case: UI needs to allow users a chance to copy content from tooltip content.

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -10,6 +10,7 @@ const propTypes = {
   tether: PropTypes.object,
   toggle: PropTypes.func,
   children: PropTypes.node,
+  autohide: PropTypes.bool,
   delay: PropTypes.oneOfType([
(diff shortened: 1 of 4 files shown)


eddywashere:
Thought this already existed when I commented over [here](https://github.com/reactstrap/reactstrap/issues/115#issuecomment-244806080).

src/Tooltip.js
```diff
@@ -154,14 +180,16 @@ class Tooltip extends React.Component {
 
     return (
       <TetherContent
-        onMouseOver={this.onMouseOverTooltip}
```
eddywashere: extra props are not part of the clone children stuff that goes on in TetherContent. I think it makes more sense to apply these event handlers to the children directly.

PR #157, opened 2016-09-26
chore(tests): remove karma

### LINKED ISSUE #154: Tests run out of memory
Watch task is basically dead at the moment. Noticed that this issue is happening way more often then it used to. @TheSharpieOne mentioned moving off of karma all together and I am +1000 to that. Keeping this issue open to track the status of it. 

also, I had some experiment with enzyme and jest in https://github.com/reactstrap/component-template and was pleasantly surprised by the watch, coverage and just general experience of using it.

TheSharpieOne:
Closes #154
As they say, karma's a b!%$#. so this removes that.
Uses react-script to make it similar to the component template
This also sorta kinda removes Jasmine in favor of Jest.. which uses
the Jasmines API... so not really :smile:


eddywashere:
@TheSharpieOne 🙌

PR #158, opened 2016-09-27
build(tests): remove istanbul

eddywashere:
This was slowing down the tests


.travis.yml
```diff
@@ -4,10 +4,12 @@ node_js:
 cache:
   directories:
     - node_modules
+before_install:
+  - npm install -g npm@latest
 before_script:
   - chmod +x scripts/docs
 script:
-  - npm run cover
+  - CI=true npm run cover
```
TheSharpieOne: Travis already has this predefined

.travis.yml
```diff
@@ -4,10 +4,12 @@ node_js:
 cache:
   directories:
     - node_modules
+before_install:
+  - npm install -g npm@latest
```
TheSharpieOne: Why is this needed? Travis will have node 6 with npm 3 since node v6 is in this config. Adding this extra steps makes travis take longer and can potentially break things when npm has a new major version release.

eddywashere: Was checking if the npm version mattered, I'll remove

.travis.yml
eddywashere: It does. I couldn't find it in their docs. I was really hoping the slow tests were caused by something obvious but nothing came up. Tests run so fast locally. I think this is a better scenario than it was before but I definitely want to figure that out. I'll create an issue.

TheSharpieOne: https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
The suits themselves are running slow. It ends up showing the timing for each suit to indicate that a particular suit took longer than expected to run. That indicates that the tests are taking a long time and not that it is hanging waiting for user input.
It runs pretty quick locally for me as well. I am think it might be because I have a powerful machine and travis is running in a VM with little processing power. idk.

eddywashere: Thanks for the direct link to that section, not sure how I missed it. I'll check out a limited docker environment to see if I can reproduce it.

PR #159, opened 2016-09-27
fix(label): add disabled class when form-check-inline and disabled

TheSharpieOne:
Adds the disabled class when the label has both the inline and
check props. This is the way bootstrap disables inline radio
and checkboxes since there is no wrapping element.

I must have missed this use-case before when creating the forms and labels. The whole form-check classes are new to bootstrap 4.

Note: there is currently an open issue with bootstrap (twbs/bootstrap#20799) to correct the color of the disabled label's text (and to clarify how to denote disabled inline radio/checkboxes in the docs as it is not there, I had to infer it from [the styles](https://github.com/twbs/bootstrap/blob/v4.0.0-alpha.4/scss/_forms.scss#L241-L243))

### DIFF
--- a/src/Label.js
+++ b/src/Label.js
@@ -21,6 +21,7 @@ const propTypes = {
   hidden: PropTypes.bool,
   check: PropTypes.bool,
   inline: PropTypes.bool,
+  disabled: PropTypes.bool,
   size: PropTypes.string,
   for: PropTypes.string,
   tag: PropTypes.string,
@@ -43,6 +44,7 @@ const Label = (props) => {
     tag: Tag,
     check,
     inline,
+    disabled,
     size,
     for: htmlFor,
     ...attributes,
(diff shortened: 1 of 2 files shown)


eddywashere:
Nice work getting that other issue into the alpha v4 release. lgtm!

src/Label.js
```diff
@@ -43,6 +44,7 @@ const Label = (props) => {
     tag: Tag,
     check,
     inline,
+    disabled,
```
eddywashere: should `disabled` be passed into the rendered component? I don't think so in this context, but it's usually something that would get passed through in something like an input.

eddywashere: merging this as is, just curious about that

TheSharpieOne: For this use case, the disabled class is needed on the label. You'd end up passing it to both the label and the Input. Most of the time, the label does not wrap the Input. When not inline, you pass disabled to the FormGroup and the Input.
The validation library I create does these thing for the developer using contexts.

eddywashere:
published in 3.3.1 💫

PR #162, opened 2016-09-28
Add Alert Component

aaronpanch:
Hello! I was working on a small project using reactstrap, made an alert, and thought it would be cool to clean things up and make a PR.  In summary, this simply adds a Bootstrap 4 alert, and a documentation page with some examples.

I would like to note a few things for discussion, in addition to any comments or suggestions you might have!
- I've added a peer dependency on `react-addons-css-transition-group`.  It wasn't my favorite thing to do; however, it made implementing the dismissible fade-out easier.
- I didn't know the order for the documentation pages, so I've added it to the end.
- Should there be a link to the Bootstrap docs describing how to use `.alert-heading` and `.alert-link`?  I only ask to maintain some feature parity with Bootstrap 4.
- In issue #6, there props outlined include 'isOpen' and 'toggle', but those didn't seem to match the bootstrap behavior (from my understanding).  It would be cool to add though...apologies for not asking before submitting a PR--it's easy to change and modify!
- The accessibility attributes are also included.

Thanks!

### DIFF
--- a/docs/lib/Components/index.js
+++ b/docs/lib/Components/index.js
@@ -106,8 +106,8 @@ class Components extends React.Component {
           to: '/components/jumbotron/'
         },
         {
-          name: 'Alert',
-          to: '/components/alert/'
+          name: 'Alerts',
+          to: '/components/alerts/'
         }
       ]
     };
--- a/docs/lib/Components/index.js
+++ b/docs/lib/Components/index.js
@@ -104,6 +104,10 @@ class Components extends React.Component {
         {
           name: 'Jumbotron',
           to: '/components/jumbotron/'
+        },
+        {
+          name: 'Alert',
+          to: '/components/alert/'
         }
       ]
     };
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -1,0 +1,96 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+
+const propTypes = {
+  className: PropTypes.any,
+  color: PropTypes.string,
+  onDismiss: PropTypes.func,
+  tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string])
+};
+
+const defaultProps = {
+  color: 'warning',
+  tag: 'div'
+};
+
+export const Alert = (props) => {
+  const {
+    className,
+    tag: Tag,
+    color,
+    onDismiss,
+    ...attributes
+  } = props;
+
+  const classes = classNames(
+    className,
+    'alert',
+    `alert-${color}`,
+    { 'alert-dismissible': onDismiss }
+  );
+
+  return (
+    <Tag {...attributes} className={classes} role="alert">
+      { onDismiss ?
+        <button type="button" className="close" aria-label="Close" onClick={onDismiss}>
+          <span aria-hidden="true">&times;</span>
+        </button>
+        : null }
+      {props.children}
+    </Tag>
+  );
+};
+
+Alert.propTypes = propTypes;
+Alert.defaultProps = defaultProps;
+
+const wrapperPropTypes = {
+  dismissible: PropTypes.bool
+}
+
+const wrapperDefaultPropTypes = {
+  dismissible: false
+}
+
+class AlertWrapper extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      visible: true
+    }
+  }
+
+  onDismiss = () => {
+    this.setState({ visible: false });
+  }
+
+  render() {
+    const { dismissible, ...props } = this.props
+        , alert = <Alert {...props}
+                    key="alert"
+                    onDismiss={dismissible ? this.onDismiss : undefined} />
+
+    if (dismissible) {
+      return (
+        <ReactCSSTransitionGroup
+          transitionName={{
+            leave: 'fade',
+            leaveActive: 'out'
+          }}
+          transitionEnter={false}
+          transitionLeaveTimeout={300}>
+          {this.state.visible ? alert : null}
+        </ReactCSSTransitionGroup>
+      )
+    } else {
+      return alert;
+    }
+  }
+}
+
+AlertWrapper.propTypes = wrapperPropTypes;
+AlertWrapper.defaultProps = wrapperDefaultPropTypes;
+
+export default AlertWrapper;
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -11,13 +11,19 @@ const propTypes = {
   color: PropTypes.string,
   isOpen: PropTypes.bool,
   toggle: PropTypes.func,
-  tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string])
+  tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
+  transitionAppearTimeout: PropTypes.number,
+  transitionEnterTimeout: PropTypes.number,
+  transitionLeaveTimeout: PropTypes.number
 };
 
 const defaultProps = {
   color: 'success',
   isOpen: true,
-  tag: 'div'
+  tag: 'div',
+  transitionAppearTimeout: 150,
+  transitionEnterTimeout: 150,
+  transitionLeaveTimeout: 150
 };
 
 const Alert = (props) => {
@@ -28,6 +34,9 @@ const Alert = (props) => {
     isOpen,
     toggle,
     children,
+    transitionAppearTimeout,
+    transitionEnterTimeout,
+    transitionLeaveTimeout,
     ...attributes
   } = props;
 
@@ -53,11 +62,20 @@ const Alert = (props) => {
     <ReactCSSTransitionGroup
       component={FirstChild}
       transitionName={{
+        appear: 'fade',
+        appearActive: 'in',
+        enter: 'fade',
+        enterActive: 'in',
         leave: 'fade',
         leaveActive: 'out'
       }}
-      transitionEnter={false}
-      transitionLeaveTimeout={300}>
+      transitionAppear={transitionAppearTimeout > 0}
+      transitionAppearTimeout={transitionAppearTimeout}
+      transitionEnter={transitionEnterTimeout > 0}
+      transitionEnterTimeout={transitionEnterTimeout}
+      transitionLeave={transitionLeaveTimeout > 0}
+      transitionLeaveTimeout={transitionLeaveTimeout}
+      >
       {isOpen ? alert : null}
     </ReactCSSTransitionGroup>
   );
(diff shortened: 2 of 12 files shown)


src/Alert.js
```diff
@@ -0,0 +1,96 @@
...
+
+const wrapperPropTypes = {
+  dismissible: PropTypes.bool
+}
+
+const wrapperDefaultPropTypes = {
+  dismissible: false
+}
+
+class AlertWrapper extends React.Component {
```
eddywashere: This should be in it's own file

src/Alert.js
```diff
@@ -0,0 +1,96 @@
...
+
+  render() {
+    const { dismissible, ...props } = this.props
+        , alert = <Alert {...props}
+                    key="alert"
+                    onDismiss={dismissible ? this.onDismiss : undefined} />
+
+    if (dismissible) {
+      return (
+        <ReactCSSTransitionGroup
```
eddywashere: Consider using the [Render a single child](https://facebook.github.io/react/docs/animation.html#rendering-a-single-child) method to avoid the extra element created by `ReactCSSTransitionGroup`. I've never done this before but it seems like just the right thing to use.

src/Alert.js
```diff
@@ -0,0 +1,96 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      visible: true
+    }
+  }
+
+  onDismiss = () => {
+    this.setState({ visible: false });
```
eddywashere: I don't think this needs to be an uncontrolled component. Users should control state. I think this could accept the `isOpen` prop for toggling visibility and use that logic with [single child method](https://facebook.github.io/react/docs/animation.html#rendering-a-single-child) to control how it's shown. `onDismiss` would be replaced by the `toggle` callback.

docs/lib/Components/index.js
```diff
@@ -104,6 +104,10 @@ class Components extends React.Component {
         {
           name: 'Jumbotron',
           to: '/components/jumbotron/'
+        },
+        {
+          name: 'Alert',
+          to: '/components/alert/'
```
eddywashere: I would update title and url to be pluralized to closer match v4 docs.

src/Alert.js
```diff
@@ -0,0 +1,96 @@
...
+
+const propTypes = {
+  className: PropTypes.any,
+  color: PropTypes.string,
+  onDismiss: PropTypes.func,
+  tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string])
+};
+
+const defaultProps = {
+  color: 'warning',
```
eddywashere: can you change this to `success` for default

src/Alert.js
eddywashere: it might just be one component if you move the `ReactCSSTransitionGroup` up to `Alert`. Also, you could pass `tag` through as the `component` prop of [ReactCSSTransitionGroup](https://facebook.github.io/react/docs/animation.html#rendering-a-different-component).

src/Alert.js
```diff
@@ -0,0 +1,96 @@
...
+                    onDismiss={dismissible ? this.onDismiss : undefined} />
+
+    if (dismissible) {
+      return (
+        <ReactCSSTransitionGroup
+          transitionName={{
+            leave: 'fade',
+            leaveActive: 'out'
+          }}
+          transitionEnter={false}
```
eddywashere: I think it should be possible to transition on enter, not sure what that involves. Any reason not to include it?

src/Alert.js
```diff
@@ -0,0 +1,96 @@
...
+          transitionLeaveTimeout={300}>
```
eddywashere: I would make this configurable for leave and onenter.

eddywashere:
Did you look at using the `Fade` component? It might help avoid the peer dependency. I think I avoided `ReactCSSTransitionGroup` because I didn't have as much control as I did with `ReactTransitionGroup`. There's definitely room for improvement in `Fade`. Thanks for working on this!

aaronpanch:
@eddywashere  Thanks for the comments!

One question about procedure: for the changes, should I add separate commits, squash them into the originals, or separate and squash later?

Also, I did take a quick look at the `Fade` component but it didn't seem to support a transition out (yet), and looked similar to the CSS transition group. I can take a deeper look today.

src/Alert.js
aaronpanch: Yeah, I totally agree--especially if we support a fade-in animation and do the single child thing, there isn't really a need for a separate wrapping component.

src/Alert.js
aaronpanch: That's a great suggestion. I only omitted it for attempted parity with Bootstrap, but it's an easy 2-liner addition.  I suppose we should also support the `transitionAppear` for initial mounting too.

src/Alert.js
aaronpanch: For sure!

docs/lib/Components/index.js
aaronpanch: Woops...Good call

src/Alert.js
aaronpanch: I agree with the notion that this doesn't need to be an uncontrolled component but for simplicity perhaps theres another way to achieve the same goal?  Which is better, someone rendering an long-lasting alert and toggles its visibility or mounting and un-mounting the component.

src/Alert.js
aaronpanch: I haven't either, but it seems cool!

eddywashere:
Don't worry about squashing things. I prefer detailed commits. I squash and merge via github once it's all good to go. I'll need some time to better describe the limitations I thought I found in css transition group. But feel free to look into it as well. I think it was regarding unique class names on transition life cycle events, maybe something specific to "while transitioning". which bootstrap applies classes to.

src/Alert.js
eddywashere: It could be a controlled and uncontrolled component. But I think that leads to inconsistency (updating a prop vs updating via setState) and extra docs. Preferring props over internal state is a general rule for this library. I built [component-template](https://github.com/reactstrap/component-template) with this sort of thing in mind. Someone could build `reactstrap-controlled-components`, which wrap these components and adds a controlled version of `Alert`. It's not the most convenient thing to do but it definitely enables more prescriptive ways of doing things. Even though that prescription is simplicity.

aaronpanch: Yeah, I hear ya.  In terms of development, a controlled component is real simple so I'll go with that. Are you using the `isOpen` and `toggle` terminology to be consistent throughout the library?  I suppose `isVisible` matches what's going on semantically but comes at a cost of inconsistency.

aaronpanch: Also, to clarify, I wasn't suggesting a mix of controlled and uncontrolled, rather an approach like a dropdown where it internally controlled its own state.  All in all, a controlled component is in the works.

src/Alert.js
```diff
@@ -0,0 +1,69 @@
...
+        </button>
+        : null }
+      { children }
+    </Tag>
+  );
+
+  return (
+    <ReactCSSTransitionGroup
+      component={FirstChild}
+      transitionName={{
```
eddywashere: experimented with this, the following example seems to apply the right classes and works with the default opacity animation of `.15s`. 

```jsx
<ReactCSSTransitionGroup
  component={FirstChild}
  transitionName={{
    enter: 'fade',
    enterActive: 'in',
    leave: 'fade',
    leaveActive: 'out',
    appear: 'fade',
    appearActive: 'in'
  }}
  transitionEnterTimeout={150}
  transitionLeaveTimeout={150}
>
  {isOpen ? alert : null}
</ReactCSSTransitionGroup>
```

src/Alert.js
eddywashere: The Dropdown examples do advocate set state but that's in the example files vs the src. The only things I see that use state in the `src` files are `Fade` and `TabContent`. I think `isOpen` is not the best term but it's not way off from describing this.

eddywashere:
Notes in animation docs

> When using ReactCSSTransitionGroup, there's no way for your components to be notified when a transition has ended or to perform any more complex logic around animation. If you want more fine-grained control, you can use the lower-level ReactTransitionGroup API which provides the hooks you need to do custom transitions.

That's why I went with it for `Fade`. Totally forgot I added hooks for onEnter and onLeave, based on the `ReactTransitionGroup` lifecycle callbacks. I used `onLeave` to teardown `Modal`. I created a gist with the changes I explored to `Fade` to support `Alert`. It works for `Alert` but it breaks `Modal`. I don't know what it would take to update Modal to use it. [Check out the changes here](https://gist.github.com/eddywashere/52c148968ff3d8cb5cbea7d3135b6c94). It seems more involved. I'd be ok with you sticking with `ReactCSSTransitionGroup` for the quick win and spending more time on making `Fade`.

If you stick with `ReactCSSTransitionGroup`, just be sure to add it to `npm install` portion of docs. Going this route, we don't have to introduce the `onEnter`, `onLeave` callbacks until `Fade` is updated.  Also free to add `AlertHeading` and `AlertLink` 👍 .

Thanks again for working on this. Can't wait to use it! Let me know what you think about direction for this.

aaronpanch:
Sorry for the delay -- I just moved to a new apartment!  I agree with the direction: lets stick with `ReactCSSTransitionGroup` for the easy win. I'll add it to the docs also.  I see what you're saying for the `Fade` component.  I think I have some time to look at it.  Also, I'm still adding the parameterized transition timeouts.  Almost there!

src/Alert.js
aaronpanch: I'm cool with `isOpen`.  The fact that people can probably guess most of the property names in this library is awesome.

src/Alert.js
aaronpanch: See commit 462f6cd.  That adds the timeout props to `Alert` and 1dc4e6d documents it. Setting each to `0` will cancel the respective animation.  Is that cool?

eddywashere: Awesome! That works for me.

eddywashere:
@aaronpanch no rush, I see commits added after your comment. Could you let me know when this is good to review?

aaronpanch:
@eddywashere I think they're good to review when you have a chance! Thanks for all the feedback!

eddywashere:
thanks @aaronpanch! :shipit:

IndifferentDisdain:
I'm getting the following error after this change; any thoughts? It's causing webpack build to fail.

ERROR in ./~/reactstrap/lib/Alert.js
Module not found: Error: Cannot resolve module 'react-addons-css-transition-group' in N:\workspaces\credentialing\node_modules\reactstrap\lib
 @ ./~/reactstrap/lib/Alert.js 17:37-81

eddywashere:
@IndifferentDisdain  A new peer dependency was added. I was considering moving it to dependencies to avoid problems like this. I'm not sure if that would cause more problems.

For now, I would recommend saving this dep to your project.

IndifferentDisdain:
@eddywashere Thanks, I just rtfm'd and noticed the additional dependency. :smile_cat:

PR #163, opened 2016-09-29
Making every URL like https://webpack.github.io/docs/webpack-dev-serv…

andrerpena:
…er.html to work when you hit F5 on dev mode.

Today.. if you type `npm start` and navigate to, let's say, `https://webpack.github.io/docs/webpack-dev-server.html`, and then you hit F5, the page doesn't display.

Now it does.


eddywashere:
Nice fix @andrerpena!

Issue #164, opened 2016-10-01
Modal call blocks and removes scrolls on parent window


wzup:
An example is your modal demo https://reactstrap.github.io/components/modals/
Call modal (click "Launch Modal" button) , then close the modal (click "Do Something" or "Cancel" buttons within opened modal) and you won't see any scrollbars on the page. The page is blocked. You cannot scroll it.

andrerpena:
I've noticed it too. Additionally, the fact that the parent window's width changes makes it to "shake" when you open the Modal.

eddywashere:
I think I know what's going on. Just so I have all the info, can you tell me the following: what browser, browser version and OS version are you using? Do you see any of this happen in the same portion of bootstrap docs? Thanks for reporting! I'll look into it in the meantime.

andrerpena:
![demo](http://i.imgur.com/86q4Yx9.gif)

eddywashere:
Got it. To reproduce on osx, I set `show scroll bars` to `always` in general > settings. I see the issue. In bootstrap docs, the `modal-open` class is applied to the body on open. I thought that was enough but apparently they also add padding-right to accommodate for scrollbar when present. You won't see that if your OS autohides scrollbar. I'll look into that.

The other issue in the original comment:

> The page is blocked. You cannot scroll it.

This should be the desired behavior when the modal is open. At least this is how it appears in the [bootstrap v4 modal docs](http://v4-alpha.getbootstrap.com/components/modal/) (the `modal-open` class applies `overflow:hidden` to body). 

Are you saying scrolling is blocked after it the modal is closed?

eddywashere:
@andrerpena the issue you described is fixed in `3.3.2`. @wzup I wasn't able to reproduce what you described exactly. Found something similar and opened up this issue https://github.com/reactstrap/reactstrap/issues/166. If that's what you're talking about, lets close this issue and track it over there.

TheSharpieOne:
Closing this issue as suggested above since it has not been active for several months.

PR #165, opened 2016-10-01
fix(Modal): Account for body padding when scrollbar is present

eddywashere:
Currently, body and fixed navbars get pushed because of the modal-open class. This pr only fixes the regular body content. Need to look at how bootstrap does this for the [fixed content](https://github.com/twbs/bootstrap/blob/v4.0.0-alpha.4/js/src/modal.js#L70).

Fixes part of #164.

Need to look at [adjusting](https://github.com/twbs/bootstrap/blob/v4.0.0-alpha.4/js/src/modal.js#L395-L406) modal dialog as well. 

Also I had a hard time getting tests in for this. Going leave those commented out for now.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -3,6 +3,11 @@ import ReactDOM from 'react-dom';
 import classNames from 'classnames';
 import TransitionGroup from 'react-addons-transition-group';
 import Fade from './Fade';
+import {
+  getOriginalBodyPadding,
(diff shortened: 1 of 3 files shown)


eddywashere:
Opened up an issue upstream to track the fixed content shift. https://github.com/twbs/bootstrap/issues/20824

PR #171, opened 2016-10-08
Add nested modal example

jonjaques:
First of all, awesome work.

On the version I was using, I couldn't seem to make nested modals work - so I thought I was going to have to hack it in; seeing as it's not supported in vanilla BS.

Turns out once I put an example together it already worked! Much kudos 😄 

So, just thought I'd throw this up here in case you wanted to showcase that.


eddywashere:
@jonjaques thanks! Also, thanks for trying that use case out and submitting the example. 👍

Issue #169, opened 2016-10-05
Flexbox grid support


madisvain:
Currently Col component does not support flexbox as there is no way to define a <Col> without it having the column number.

http://v4-alpha.getbootstrap.com/layout/flexbox-grid/
_Unlike the default grid system where a grid column starts as full-width in the xs tier, flexbox requires a .col-{breakpoint} class for each tier._

Flexbox would require to have the possibility of defining `<div class="col-xs">` but currently the <Col> element will add col-xs-12 by default.

eddywashere:
Thanks for filing the issue! I could imagine leveraging the size props, and adding a boolean option that would only output the `col-${size}` class. It would look like this:

```jsx
<Row>
  <Col xs>col-xs</Col>
  <Col xs="6">col-xs-6</Col>
  <Col xs>col-xs</Col>
</Row>
```

Are the push/pull/offset classes supported in flexbox grid? If so, maybe `flex` is a prop or the size props accept `flex` or `auto` as a way to output the auto layout class.

```jsx
<Row>
  <Col xs="flex">col-xs</Col>
  <Col xs="6">col-xs-6</Col>
  <Col xs={{offset: 1, size: 'flex'}}>col-xs</Col>
</Row>
```

```jsx
<Row>
  <Col xs="auto">col-xs</Col>
  <Col xs="6">col-xs-6</Col>
  <Col xs={{offset: 1, size: 'auto'}}>col-xs</Col>
</Row>
```

ronnyhaase:
- [x] Add flexbox auto size ability to Col component (PR #175)
- [ ] Add flexbox alignment support to Col & Row component
- [ ] Use a Bootstrap release with flexbox enabled for Docs

ronnyhaase:
Hey guys!

I'm almost done with the last task for full flexbox support: Support flex alignment classes of Bootstrap.

I'd like to ask you for feedback regarding **naming** and **types** for the necessary properties:

Context: http://v4-alpha.getbootstrap.com/layout/flexbox-grid/

My current implementation looks like:

`<Col first="xs" first={['xs', 'sm']} first={{ xs: true, sm: true, /* ... */ }} first />`

Where a boolean `first` could may resolve to `['xs', 'sm', 'md', 'lg', 'xl']` and/or accept `"all"`?

What do you think about it? And also, do you think the props should be prefixed, like `order-first`. Or is it better to just have one prop `order` which expects `first`, `last` and `unordered` as attributes?

[...]

PR #175, opened 2016-10-13
Add flex-support for Col component #169

ronnyhaase:
Hi!

As requested in issue #169 I implemented support for dropping the size for the Col component, respectively flexbox support.

I added support for all suggestions by @eddywashere, meaning a value can be dropped at all, or set to "flex" and "auto", like:

```html
<Col xs />
<Col xs="auto" />
<Col xs="flex" />

<Col
      xs={{ size: true, push: 2, pull: 2, offset: 2 }}
      sm={{ size: 'auto', push: 2, pull: 2, offset: 2 }}
      md={{ size: 'flex', push: 2, pull: 2, offset: 2 }}
/>
```

If the object's size shall accept boolean as well, and both "flex" and "auto" should be supported may needs some discussion?

Corresponding unit tests were added as well.

Looking forward to your feedback
Ronny

### DIFF
--- a/docs/lib/Components/LayoutPage.js
+++ b/docs/lib/Components/LayoutPage.js
@@ -41,6 +41,7 @@ export default class LayoutsPage extends React.Component {
 const columnProps = PropTypes.oneOfType([
   PropTypes.string,
   PropTypes.number,
+  PropTypes.bool,
   PropTypes.shape({
     size: stringOrNumberProp,
     push: stringOrNumberProp,
--- a/docs/lib/examples/Layout.js
+++ b/docs/lib/examples/Layout.js
@@ -13,6 +13,10 @@ export default class Example extends React.Component {
           <Col xs="6">col-xs-6</Col>
         </Row>
         <Row>
+          <Col xs>.col-xs (flexbox)</Col>
+          <Col xs>.col-xs (flexbox)</Col>
+        </Row>
+        <Row>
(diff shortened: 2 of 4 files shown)


eddywashere:
@ronnyhaase nice work! I meant to finalize the options. After thinking about it, I don't think `flex` fits in as a descriptive value for the size attribute. Could you remove the `flex` value support and could you also update the docs? Everything else looks good. Thanks!!!

ronnyhaase:
I agree regarding `flex`. What's your opinion about accepting boolean for size in the configuration object @eddywashere? I don't like it as well.

Also allow me to ask if you prefer a rebase for removing `flex` or a separate commit.

I will make the changes and add documentation later on.

Thank you!

eddywashere:
Re: Boolean, I was originally cool with it. I'm a little hesitant to commit to it. Just doing the auto seems safer to maintain and is explicit. 

Not super strict but I like to rebase. Either way I squash into a single commit when merging pr in github.

ronnyhaase:
Hej! So I removed "flex" as valid option, and bool as valid type for size of config object.
Also I added two examples to the docs.

But unfortunately these examples are not working, since the docs use the Bootstrap release from NPM which has flex disabled by default. 
How should I proceed from here?

eddywashere:
that's a really good question. react-helmet might support this. If not, we might need to hack in a separate html template for flexbox pages. If you need this out asap. I can merge it without the docs. let me know what you think.

PR #177, opened 2016-10-16
docs(components): fix typo in components page menu

polmauri:
Change menu item incorrectly spelled as "Breadscrumbs" to the correct form "Breadcrumbs".


eddywashere:
Thanks @polmauri!

PR #181, opened 2016-10-17
feat(tether): add ability to get tether ref

### LINKED ISSUE #174: No way to access tether object to trigger a manual position()
There is currently no way for a Tooltip to trigger a manual "position()" to re-position the tooltip.

One way would be to add a ref on the tether object and make a function getTetherInstance on the Tooltip.

```
    getTetherInstance(){
        if (this.tetherContent && this.tetherContent._tether){
            return this.tetherContent._tether;
        }
        return null;
    }

    <TetherContent
         ref={(ref) => this.tetherContent = ref}
    >
```

What do you think?

TheSharpieOne:
Fixes #174
Add `tetherRef` to TetherContent which works similarly to `ref`

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react';
+import omit from 'lodash.omit';
 import TetherContent from './TetherContent';
 import { getTetherAttachments, tetherAttachements } from './utils';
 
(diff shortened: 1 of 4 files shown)


src/Tooltip.js
```diff
@@ -178,24 +180,27 @@ class Tooltip extends React.Component {
...
-        {...attributes}
         arrow="tooltip"
         tether={tetherConfig}
+        tetherRef={this.props.tetherRef}
         isOpen={this.props.isOpen}
         toggle={this.toggle}
       >
         <div
-          className="tooltip-inner"
+          {...attributes}
```
eddywashere: good call!

PR #182, opened 2016-10-17
fix(tooltip): fix when toggle is not provided

TheSharpieOne:
Before, when props.toggle was not provided, this.toggle would throw
Essentially when someone just provides the `isOpen` prop and wants
to manually control the hiding/showing.

`onMouseOverTooltip` would trigger `show`, which would trigger `toggle` when the tooltip was not being shown. Without `toggle` prop, it would error.

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -26,7 +26,8 @@ const defaultProps = {
   isOpen: false,
   placement: 'bottom',
   delay: DEFAULT_DELAYS,
-  autohide: true
+  autohide: true,
+  toggle: function () {}
 };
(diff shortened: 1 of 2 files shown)


ronnyhaase:
👍

eddywashere:
Should this be added to other components?

TheSharpieOne:
Which other components? I looked at pop over and it did not work the same way. I mean, we could default some of the callback functions, but at a certain point, they are not explicitly used like it was here.

eddywashere:
That's right! Nothing to see here :)

PR #186, opened 2016-10-18
chore(docs): button spacing

TheSharpieOne:
This removes the special button margin which is in the docs and
replaces it with `{' '}` to add an actual space in the HTML markup.
This makes the button spacing work like bootstrap does and documents
`{' '}` being used.

I am addressing this because I feel that it should be addressed. I have seen many developers not have spacing between buttons because they followed/copy&paste the docs here (which did not have a space, only special css) and it makes it look bad. Bootstrap intended there to be a space separating the buttons. Documenting the space saves headaches and prevents users from adding special CSS to add the space... and not have the space in some places (such as the CSS the docs were using)

### DIFF
--- a/docs/lib/Components/ButtonsPage.js
+++ b/docs/lib/Components/ButtonsPage.js
@@ -53,22 +53,22 @@ export default class ButtonsPage extends React.Component {
         </pre>
         <h3>Sizes</h3>
         <div className="docs-example">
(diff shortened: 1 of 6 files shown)


eddywashere:
I'm of the same mindset. I think I avoided this at first because react placed a span instead of a text node. Thank you for picking this up!

Issue #79, opened 2016-07-31
Collapse


eddywashere:
http://v4-alpha.getbootstrap.com/components/collapse/
- [x] Collapse
  - consider using refactoring [Fade component](https://github.com/reactstrap/reactstrap/blob/master/src/Fade.jsx)

eddywashere:
Todo: list transition states and classes. I looked at this for a bit and I think there's something more to just adding and removing classes during these states. I believe there is a height inline style applied before showing the content.

LiJinyao:
I'd like to build this component😊,  can you assign it to me?

lvauvillier:
Humm, I didn't see this issue before sending a PR (#190) for the navbar.
Maybe I can help here and update it to do a more generic one with a Collapse component

LiJinyao:
Ops, My work almost finished. Maybe you can left it to me.

eddywashere:
@lvauvillier I saw the pr. I was going to suggest that you use or look into the work @LiJinyao is doing. Thanks for the update @LiJinyao. Let me know if you need anything. Component is looking good. I'll try it out as soon as I get some free time. Traveling this week until Monday.

lvauvillier:
@LiJinyao thanks for your work, I'll try it with the navbar toggler.
@eddywashere If all works fine, I'll update my PR using the new `Collapse` component. (My PR will finally just be a doc update for the navbar toogler use case)

lvauvillier:
@LiJinyao's, your Collapse component is working fine
I created a branch with a sample mixing Collapse and the navbarToggler here : https://github.com/lvauvillier/reactstrap/tree/NavbarToggler-sample

Ready to PR when your component will officially be available.

LiJinyao:
@lvauvillier Good job! I'll make a PR toady.

eddywashere:
released in 3.8.0, thanks @LiJinyao!

gthomas-appfolio:
FYI bootstrap v5 alpha now uses `collapse active` instead of `collapse in`:

https://github.com/twbs/bootstrap/pull/21019

What's the proper way to follow up with these changes in the reactstrap codebase?  PR as they are made in BS4 or you wait for certain milestones?

eddywashere:
[...]

PR #190, opened 2016-10-20
feat(NavbarCollapse) : add a new element to handle collapsable zone in Navbar

lvauvillier:
Add a new NavbarCollapse element.
Update documentation with an example


lvauvillier:
This behavior should use the future generic Collapse component.
See #79

PR #193, opened 2016-10-21
fix(docs): Update Bootstrap to alpha.5 and switch to flexbox enabled …

ronnyhaase:
Updated to Bootstrap v4-alpha.5 which contains a precompiled CSS file with flex enabled, and updated Webpack alias to use this one.

Now the Layout example in the docs is fine.


eddywashere:
@ronnyhaase woot!

PR #196, opened 2016-10-22
docs(Navs): Update nav dropdown markup

eddywashere:
Noticed an issue where the dropdown in nav (pills and tabs) were looking weird. It turns out it was using incorrect markup. Instead of nesting nav-link in the dropdown toggle, it should be combined. I’m not sure if this was part of alpha v5, or if the markup and classes were always off.

I’m hesitant to add a new component to simplify this or to add a prop to DropdownToggle to enable this common usage. For now, I’m updating the docs to use the built in flexibility of the existing dropdown toggle component.


eddywashere:
Looks like this was part of #120 - specifically [this](https://github.com/reactstrap/reactstrap/pull/120/files#diff-1d08ef71b40cb51f7bc32df77833d1e7L65).

PR #200, opened 2016-10-24
fix(NavbarToggler) remove unnecessary menu char

lvauvillier:
remove unnecessary char in NavbarToggler component


eddywashere:
I was wondering why this was ever there. It looks like ☰ was removed in alpha v5. Thanks for the pr @lvauvillier :shipit:

PR #201, opened 2016-10-24
add Collapse component #79

LiJinyao:


src/Collapse.js
```diff
@@ -0,0 +1,101 @@
...
+    const collapse = this.state.collapse;
+
+    if (willOpen && collapse === HIDDEN) {
+      // will open
+      this.setState({ collapse: SHOW }, () => {
+        // start transition
+        this.element.style.height = `${this.getHeight()}px`;
+      });
+      this.transitionTag = setTimeout(() => {
+        this.setState({ collapse: SHOWN });
```
eddywashere: I believe the bootstrap js removes the height inline style after it has been shown. This would likely prevent any issues when added content and then collapsing the content again.

LiJinyao: Oh Sorry! I found the problem. I'll fix it.

src/__tests__/Collapse.spec.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { shallow, mount } from 'enzyme';
+import Collapse from '../Collapse';
+
+describe('Collapse', () => {
+  it('should render children', () => {
```
eddywashere: Test coverage went down. Could you setup `jasmine.clock` here since you're using `setTimeout`.

```js
  beforeEach(() => {
    jasmine.clock().install();
  });

  afterEach(() => {
    jasmine.clock().uninstall();
  });
```

After prop changes, I believe you'll need to then use, `jasmine.clock().tick(400);` to move the virtual clock forward and test for the changes.

webpack.dev.config.js
```diff
@@ -35,6 +35,7 @@ var paths = [
   '/components/tabs/',
   '/components/jumbotron/',
   '/components/alerts/',
+  '/components/collapse',
```
eddywashere: can you add a trailing `/` to generate the correct route path. I noticed a console warning regarding the active class applied to the nav .

src/__tests__/Collapse.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    const wrapper = shallow(<Collapse><p>hello</p></Collapse>).find('p');
+    expect(wrapper.text()).toBe('hello');
+  });
+
+  it('should have default isOpen value', () => {
+    const wrapper = shallow(<Collapse />);
+    expect(wrapper.instance().props.isOpen).toEqual(false);
+  });
+
+  it('should render with calss "collapse"', () => {
```
eddywashere: couple of typos here, `calss` -> `class`

src/Collapse.js
```diff
@@ -0,0 +1,104 @@
...
+      // HIDDEN
+        collapseClass = 'collapse';
+    }
+
+    const classes = classNames(
+      className,
+      collapseClass
+    );
+    return (
+      <div
```
eddywashere: Could you update this to use the `Tag (tag)` prop pattern that other components use?

src/Collapse.js
```diff
@@ -0,0 +1,104 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    const willOpen = nextProps.isOpen;
+    const collapse = this.state.collapse;
+
+    if (willOpen && collapse === HIDDEN) {
+      // will open
+      this.setState({ collapse: SHOW }, () => {
+        // start transition
+        this.element.style.height = `${this.getHeight()}px`;
```
LiJinyao: I thought the timing will be off at first, now I am trying to move these dom attribute changes to react. Let's see if it will work fine.

src/Collapse.js
```diff
@@ -0,0 +1,120 @@
...
+  isOpen: false,
+  tag: 'div'
+};
+
+class Collapse extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      collapse: props.isOpen ? SHOWN : HIDDEN,
+      height: 0,
```
eddywashere: Should height be set to null isOpen is true?

LiJinyao: Ops, Yes it is. Thanks for your patient review!

PR #202, opened 2016-10-24
fix(events): use capture events

TheSharpieOne:
Similar to a previous issue where an event was triggered AFTER some DOM manipulation which caused the event logic to run after things changed. This really only impacted when the thing that triggered the event was no longer in the DOM which changed the flow and triggered things to happen.

It is probably best to _globally_ do the same thing, which is to handle the event as soon as possible using `useCapture`.

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -105,11 +105,11 @@ class Dropdown extends React.Component {
   }
 
   addEvents() {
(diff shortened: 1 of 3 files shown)


eddywashere:
+1. I thought I went through the few areas we're using events outside of SyntheticEvent to switch to `useCapture`. I should have made an issue to track. Thanks for knocking this out!

PR #206, opened 2016-10-26
feat(cssModule): add support for css module

### LINKED ISSUE #205: css-module support?
Hi, have a use case for isolating bootstrap classes from rest of the page, and curious if there is a way to use css-modules with reactstrap so that Bootstrap does not need to be loaded as global classes?

[react-fontawesome](https://github.com/danawoodman/react-fontawesome/blob/master/api.md) has an example of how this could be specified.

TheSharpieOne:
Add cssModule prop to every component which produces a css classname
Add mapToCssModule util will can turn a normal css className string
into module className if the module object/mapping is provided.
Closes #205

### DIFF
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -1,5 +1,6 @@
 import React, { PropTypes } from 'react';
 import classNames from 'classnames';
+import { mapToCssModules } from './utils';
 import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
 
 const FirstChild = ({ children }) => (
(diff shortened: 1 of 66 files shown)


TheSharpieOne:
.....This feature should probably be documented somewhere. It's more of a global feature, similar to the `tag` prop. I'm thinking there might need to be a new page in the docs which explains these things rather than mentioning it on every component.

eddywashere:
nice! I haven't used css-modules before. I thought something like this would be more involved. +1 to docs, maybe a core concepts page or advanced usage.

:shipit:

kuzvac:
Hi @TheSharpieOne! Can you, please suggest how use this Css module feature? I can't find any documentation about this awesome feature :(

TheSharpieOne:
When you import a css/sass/less etc file it will result in a mapping object, simply pass that object to the cssModule prop. 
```
import styles from './styles.scss';
//... 
<Card color="primary" outline cssModule={styles}>
  Stuff
</Card>
//... 
```

PR #210, opened 2016-10-29
Fix eslint errors, add lint to Travis checks.

threehams:
There were a few existing eslint errors, mostly around import order. I've fixed those, then set up Travis to run `npm run lint` after the tests to avoid these in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ cache:
 before_script:
   - chmod +x scripts/docs
 script:
+  - npm run lint
   - npm run cover
 after_script:
   - npm run report-coverage
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -1,7 +1,7 @@
 import React, { PropTypes } from 'react';
 import classNames from 'classnames';
-import { mapToCssModules } from './utils';
 import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+import { mapToCssModules } from './utils';
 
 const FirstChild = ({ children }) => (
   React.Children.toArray(children)[0] || null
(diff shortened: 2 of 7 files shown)


threehams:
Going to intentionally trigger an eslint error to test this.

threehams:
Failed! https://travis-ci.org/reactstrap/reactstrap/builds/171675216

Squashing to a single commit.

.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 /dist
 /build
 /lib
+.vscode
```
TheSharpieOne: This file is attached on part of the project and is specific to your setup. You should add this in to your global git ignore.

.travis.yml
```diff
@@ -7,7 +7,7 @@ cache:
 before_script:
   - chmod +x scripts/docs
 script:
-  - npm run cover
+  - npm run cover && npm run lint
```
TheSharpieOne: I would lint before test so it will fail early. Also I think you and add this as another task and do not need to combine them in the task/line.

.gitignore
threehams: Done. I didn't know there was a setting for global .gitignore - that's really useful.

.travis.yml
threehams: I'll add them on separate lines. I think it's best to run both lint and tests on CI, so contributors get feedback on all errors (rather than needing to fix one at a time to see the results on the other).

If this works out, I'll rebase commits to match contribution guidelines... which I should have read earlier.

eddywashere: Sweet! I didn't know an extra script line would still be run if the previous script failed. Best of both experiences.

PR #211, opened 2016-10-29
Standardize on string for className props.

### LINKED ISSUE #209: className propTypes are a mix of 'any' and 'node'.
While working on some TypeScript type definitions, I noticed the propTypes for className throughout the project are a mix of PropTypes.any and PropTypes.node.

Is there a reason to diverge from React's default `string` type for this? If not, I can change them to be consistent throughout the project - it'll definitely simplify the typing work.

- version `3.7.0`
- components: all

threehams:
Fixes #209.

This is possibly a minor breaking change (dev only). Although React doesn't support className props other than strings, it's possible to extend render methods to accept other types (Radium does this), although I think it's then that library's responsibility to change the propTypes validation.

### DIFF
--- a/docs/lib/Components/AlertsPage.js
+++ b/docs/lib/Components/AlertsPage.js
@@ -30,7 +30,7 @@ export default class AlertsPage extends React.Component {
         <pre>
           <PrismCode className="language-jsx">
 {`Alert.propTypes = {
-  className: PropTypes.any,
+  className: PropTypes.string,
   color: PropTypes.string, // default: 'success'
   isOpen: PropTypes.bool,  // default: true
   toggle: PropTypes.func,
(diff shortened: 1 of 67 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 /dist
 /build
 /lib
+.vscode
```
eddywashere: Could you remove this? I think local `.gitignores` are a nice way to highlight untracked folders that are generated by scripts in the repo and .vscode seems like a platform specific thing.

eddywashere:
Good call out that this is a dev build breaking change. I think warnings in dev shouldn't warrant a bump in major version. Thanks for the updates!

.gitignore
threehams: 👍

Issue #129, opened 2016-09-07
<NavbarToggler> usage


rpassis:
Hi there ,
Would it be possible to include a simple example on how to implement a collapsible `Navbar` component? I can see the `<NavbarToggler>` is available but haven't been able to make it work.
Many thanks,
Rog

eddywashere:
Definitely need to add that. Thanks for reporting!

Kepro:
@eddywashere i looking into code, and there is nothing about onClick or something, can you add some easy example here?

eddywashere:
@Kepro @rpassis this is now documented thansk to @edgji in #216. thanks!

PR #216, opened 2016-11-01
add NavbarToggler example

edgji:
adding a collapsible Navbar component example with NavbarToggler per #129

### DIFF
--- a/docs/lib/examples/NavbarToggler.js
+++ b/docs/lib/examples/NavbarToggler.js
@@ -1,6 +1,5 @@
 import React from 'react';
-import { Navbar, NavbarToggler, NavbarBrand, Nav, NavItem, NavLink } from 'reactstrap';
-import classnames from 'classnames';
+import { Collapse, Navbar, NavbarToggler, NavbarBrand, Nav, NavItem, NavLink } from 'reactstrap';
 
 export default class Example extends React.Component {
   constructor(props) {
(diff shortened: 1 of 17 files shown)


docs/lib/examples/NavbarToggler.js
```diff
@@ -21,18 +20,18 @@ export default class Example extends React.Component {
     return (
       <div>
         <Navbar color="faded" light>
-          <NavbarBrand href="/">reactstrap</NavbarBrand>
           <NavbarToggler className="float-sm-right hidden-lg-up collapsed" onClick={this.toggleNavbar} />
-          <div className={classnames('collapse', 'navbar-toggleable-md', {in: !this.state.collapsed})}>
-            <Nav className="float-sm-left float-md-right" navbar>
+          <Collapse className="navbar-toggleable-md" isOpen={!this.state.collapsed}>
+            <NavbarBrand href="/">reactstrap</NavbarBrand>
```
eddywashere: It's too bad BS4 doesn't support this being outside of the collapsible section. I imagine the toggle icon on one side and the brand on the other side. 👍

edgji: I think it would support it with additional styling but I implemented it exactly like the sample on BS4's own docs for simplicity.

eddywashere: For sure. I'm holding off on documenting custom styles but I might do something like [this](http://codepen.io/eddywashere/pen/wGVgoN) for the docs site navbar. Thanks again for documenting this!

Issue #172, opened 2016-10-10
Referencing issue


madisvain:
Using refs for components. I have not tested other components but it seems using refs on for example Input component does not work and the input is not made available. If I replace this with a regular input then everything works as expected.

https://facebook.github.io/react/docs/more-about-refs.html

I used the input like this:

```
<Input
  ref={ (ref) => this._input = ref }
  { ...input }
  type="text"
  onChange={ this.handleInputChange }
  onFocus={ this.handleInputFocus }
  onBlur={ this.handleInputBlur }
/>
```

With the datepicker overylay example from here:
http://react-day-picker.js.org/examples/?overlay

madisvain:
_Not to be confused with the render() function that you define on your component (and which returns a virtual DOM element), ReactDOM.render() will return a reference to your component's backing instance (or null for stateless components)._

So they will not work for stateless components. Ref's are not usually needed but the use case especially for inputs is surely there.

madisvain:
Actually one more note:
https://github.com/facebook/react/issues/4936#issuecomment-179917137

It would be quite simple to add support for them to stateless components also without making them have state.

TheSharpieOne:
I'm guessing that you want the `ref` in order to access the value of the input when the `onChange`, `onFocus`, and `onBlur` events are called?

If that is the case, save the trouble and don't use `ref`. The event, `e` in the timepicker example you linked, contains a ref to the input, `event.target` (`e.target`, in the example)
The whole `this.input.blur()` and `this.input.focus()` can be more implicit by using `e.target.blur()` and `e.target.focus()` in that example. Plus, doing it this way means if you have multiple `Input`s, you don't have to manage multiple refs.
Personally, I built a whole validation library around the inputs reactstrap exposes and did not need any `ref`s. I was also able to hook up a different datepicker (https://github.com/YouCanBookMe/react-datetime) (though that datepicker has it's [own issues](YouCanBookMe/react-datetime#44)).

As you mentioned, ref's are not usually needed... But there probably should be a way to get a ref (for other reason)... Most of the components are stateless, which poses an issue. The comment in the react repo with an example is mostly showing how to get a ref for children DOM, plus the whole "getRef" style prop seems dirty. @eddywashere would have to chime in if there should be a way to access refs and, if so, which way (via "prop-in-the-middle" or make components not stateless).

eddywashere:
[...]

PR #218, opened 2016-11-01
feat(*): add way to get ref to focuable components

TheSharpieOne:
The main use-case for refs is the ability to set focus on an element, because of this, I have added the ability to get a ref to components which generate focusable items (inputs, buttons, links/anchor).

I just want to note, using a ref to get an input's value is not the correct way. Using a callback, such `onChange` or `onBlur`, you can use the supplied event to get the value without needing a ref.

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -9,6 +9,7 @@ const propTypes = {
   disabled: PropTypes.bool,
   outline: PropTypes.bool,
   tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
+  getRef: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
   onClick: PropTypes.func,
   size: PropTypes.string,
   children: PropTypes.node,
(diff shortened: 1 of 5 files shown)


eddywashere:
Could you add lodash noop as the default prop.

TheSharpieOne:
They default to `undefined` and when `undefined` is passed to `ref`, it is the same as not having a `ref` prop which I'd argue is probably better than triggering an extra function (even if the function does nothing).

ajainarayanan:
@TheSharpieOne I agree that refs could be used to set focus on an input element but wouldn't it be more straight forward by accepting a focus(or autoFocus) attribute for focus alone? We could also have a getRef attribute that will return the underlying input element's ref and can be used independently. 

Its more of getting the element ref and focusing it vs setting the attribute of element and make itself focus.

eddywashere:
@TheSharpieOne great point.  I didn't think about that. 👍 

@ajainarayanan good discussion, moving this to #172 since it's out of scope for this pr.

PR #220, opened 2016-11-04
DropdownToggle unknown prop warning

bskimball:
React throws unknown prop warning when using the tag prop. This was due to the tag prop being set twice in the render method. I added componentWillMount in order to evaluate the tag prop before the component is rendered. I also removed the de-structured props variable.

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -23,7 +23,7 @@ const defaultProps = {
   'aria-haspopup': true,
   color: 'secondary',
   tag: null,
-  ariaLabel: "Dropdown"
+  ariaLabel: 'Dropdown'
 };
 
 const contextTypes = {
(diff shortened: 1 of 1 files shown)


TheSharpieOne:
First, thanks for find this!
The problem is that while the [`tag` is pulled out of the props](https://github.com/reactstrap/reactstrap/blob/master/src/DropdownToggle.js#L57) during the destructuring, it is later [assigned back into the props when `nav` is `true`](https://github.com/reactstrap/reactstrap/blob/master/src/DropdownToggle.js#L70-L73)
The fix here is to not set `tag` in/on `props` when `nav` is `true`, but instead to change the `Tag` variable which was defined in the destructuring. This means that the destructuring assignment cannot be to a const variable.

bskimball:
I see what you mean. The easy fix would be to change const to let on props de-structure. 

Would that be preferred to evaluating the tag prop in componentWillMount method?

I see you are already fixing the issue. Should I stop on my pull request?

TheSharpieOne:
Yes, you can close this. The change was simple (1 liner + a test to ensure there is no regression) after further analyzing the code and issue to determine the intent of setting the tag that way.

eddywashere:
fix released in 3.9.0

PR #221, opened 2016-11-04
fix(DropdownToggle): only set tag prop if the Tag is default button

TheSharpieOne:
Supersedes #220 
This is one of those interesting 'defaults' which in the case of `nav`, it would tell the Button to become an anchor tag, but if the tag was already being override via DropdownToggle's `tag` prop, it would try to set the `tag` prop on that thing, which may not support the `tag` prop.

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -41,7 +41,7 @@ class DropdownToggle extends React.Component {
       return;
     }
 
-    if (this.props.nav) {
+    if (this.props.nav && !this.props.tag) {
       e.preventDefault();
     }
 
@@ -53,7 +53,7 @@ class DropdownToggle extends React.Component {
   }
 
   render() {
-    const { className, cssModule, caret, split, nav, ...props } = this.props;
+    const { className, cssModule, caret, split, nav, tag, ...props } = this.props;
     const ariaLabel = props['aria-label'] || 'Toggle Dropdown';
     const classes = mapToCssModules(classNames(
       className,
@@ -66,13 +66,19 @@ class DropdownToggle extends React.Component {
     ), cssModule);
     const children = props.children || <span className="sr-only">{ariaLabel}</span>;
 
-    if (nav && !props.tag) {
-      props.tag = 'a';
+    let Tag;
+
+    if (nav && !tag) {
+      Tag = 'a';
       props.href = '#';
+    } else if (!tag) {
+      Tag = Button;
+    } else {
+      Tag = tag;
     }
 
     return (
-      <Button
+      <Tag
         {...props}
         className={classes}
         onClick={this.onClick}
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -21,7 +21,6 @@ const defaultProps = {
   'data-toggle': 'dropdown',
   'aria-haspopup': true,
   color: 'secondary',
-  tag: Button
 };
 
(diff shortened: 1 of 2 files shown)


src/DropdownToggle.js
```diff
@@ -67,7 +67,7 @@ class DropdownToggle extends React.Component {
     ), cssModule);
     const children = props.children || <span className="sr-only">{ariaLabel}</span>;
 
-    if (nav) {
+    if (nav && Tag === Button) {
```
TheSharpieOne: Checking if the tag is the default... but what if set by the user and was another reactstrap component which could accept `tag`... should this code block be ran?
I was thinking about checking if `Tag` was a function and whether or not `Tag.propTypes.tag` was defined or something to see if it could accept this prop.
Let me know if that is something that is desired or if just checking that `Tag` is the default is good enough.

eddywashere: I would eventually like to use an env var similar to react to remove proptypes in production code since they aren't used. 

> I was thinking about checking if Tag was a function and whether or not Tag.propTypes.tag was defined or something to see if it could accept this prop.

To avoid the need for that check I think we should switch the `<Tag ... />` to `<Button ... tag={UserTag} />` 

For components that kind of extend an existing component, like this one does with `Button`, I think the render method should always use the extended component and pass the tag prop along to that component. This way we never pass a reactstrap specific prop to a non reactstrap component.

```jsx
// user renders:
<DropdownToggler color="danger" tag="span" />
// DropdownToggler render method
<Button {...extraStuff} color="danger" tag="span"  />
// Button render method
<span className="..." />
```

I think on paper it's good but I haven't tried it just yet. If it seems reasonable could you update this component in that fashion?

eddywashere:
@bskimball just curious, what was your usecase for passing a tag prop to DropdownToggler? Custom component? A work around for the time being could be a custom component that removes the tag prop in it's (final) render method.

TheSharpieOne:
Alright, I updated the PR to always use `Button`, which will render with the tag that is provided by the user/dev on the `DropdownToggle` or it's default. There is a special consideration for the case where no `tag` is provided on `DropdownToggle` and `nav` is true, in which case the `tag` will be set to `'a'` making `Button` render an anchor tag.

bskimball:
@eddywashere I was using the dropdown in the navbar and I didn't want it to look like a button. I used just the nav prop at first, but it showed as a button. I then added tag='a' which looked good and worked well, except it threw React's unknown prop error.

eddywashere:
ran this over the weekend, now I see what the problem is. Thanks for context @bskimball. My bad for the confusion @TheSharpieOne, I think I misread earlier bootstrap docs re: button in nav and maybe button works but definitely the dropdown links should not have `.btn` classes. I'll need to review and follow up later this evening.

eddywashere:
Updated with what was probably in the original pr.

src/DropdownToggle.js
```diff
@@ -42,7 +41,7 @@ class DropdownToggle extends React.Component {
       return;
     }
 
-    if (this.props.nav) {
+    if (this.props.nav && !this.props.tag) {
```
eddywashere: added this to align with what was in the `Tag` logic.

eddywashere:
@TheSharpieOne let me know if this lgtm to you

TheSharpieOne:
LGTM

Issue #224, opened 2016-11-13
Upcoming Bootstrap alpha 6 breaking changes


eddywashere:
From the alpha 6 ship list https://github.com/twbs/bootstrap/issues/20939

- [x] New flexbox column classes - (conflict with `auto` prop value) https://github.com/twbs/bootstrap/pull/20963
- [x] `Tag` renamed back to `Badge` https://github.com/twbs/bootstrap/pull/21020
- [x] `.in` and `.active` to `.show` - https://github.com/twbs/bootstrap/pull/20982/files https://github.com/twbs/bootstrap/pull/21036
  - affects the following:  Alert, Collapse, Dropdown, Modal, Popover, Tooltips
  - <strike>now could be a good time to switch `isOpen` to `isActive` across the library</strike>
  - ^ it might be possible to support both props in the next version vs doing that in current version
- [x] .row `.no-gutters` https://github.com/twbs/bootstrap/pull/21211
- [x] Progress bar rewritten https://github.com/twbs/bootstrap/pull/21473
- [x] Fix navbars and navs

Also, I think it's time to ship alpha releases that keep up with changes in `twbs/bootstrap#master` for people keeping up with latest. Also need to get #195 out to help with this.

## reactstrap v4 prerelease

We'll be using the `v4` branch to prioritize bootstrap breaking changes. Prerelease versions are now published to npm under the `next` tag and their respective prerelease version.

ronnyhaase:
I'll take care of the new column classes.

LiJinyao:
I'll take care of the Collapse component.

eddywashere:
Released `4.0.0-alpha.2` to address the new `active` class https://github.com/reactstrap/reactstrap/blob/v4/CHANGELOG.md#400-alpha2-2016-11-28

eddywashere:
More changes to look out for, https://github.com/twbs/bootstrap/pull/21036 `.in` -> `.active` -> `.show`.

eddywashere:
Released [4.0.0-alpha.6](https://github.com/reactstrap/reactstrap/blob/v4/CHANGELOG.md#400-alpha6-2017-01-03) with a few fixes. Navbar and navs seem to be very broken. I don't think `.container-fluid` still works inside of `.navbar`. Might be a bug. Need to dig more into that. `Col` also needs to be updated.

eddywashere:
[...]

PR #226, opened 2016-11-14
Tag to Badge

TheSharpieOne:
part of #224


TheSharpieOne:
It may be worth it to alias Tag and Badge, so it will not be a breaking change. People could still use `Tag`, it will output the badge classes as well as Badge, which matches bootstraps naming. I only ask because bootstrap's breaking changes do not necessarily mean a reactstrap breaking change.
Also we could deprecate `Tag` while doing that to eventually get rid of it.
Just food for thought.

eddywashere:
> bootstrap's breaking changes do not necessarily mean a reactstrap breaking change.

The current rule of thumb is, latest version of reactstrap should support the latest version of bootstrap released. If a breaking change happens in bootstrap, (classes/component renamed), it means a major version bump because consumers should have some sort of guarantee that classes won't change. I think jest snapshot tests would be helpful here for us and them. If a consumer had their own snapshot tests, they should only have failing tests when a) something that was already incorrect was fixed or b) there was a major version change.

It's probably safer not to do unrelated reactstrap api changes alongside the bootstrap breaking changes. This should hopefully make transitioning easier. Also, I'd prefer not to have master in a weird state with conditional classes, props (ex: `isOpen || isActive`), or aliases.

I've setup a `v4` branch to merge things that aren't ready for 3.x.x.

eddywashere:
Merged via #230. We can chat more about this in #224.

Issue #185, opened 2016-10-18
Add docs for ListGroup and add ListGroupItem, heading, text


eddywashere:
Following up on the original ListGroup component work. 

http://v4-alpha.getbootstrap.com/components/list-group/

ListGroupItem should support
- active
- disabled
- action
- color

Also, add the following components ListGroupItemHeading ListGroupItemText. Defaulting to the tags used in the docs.

LiJinyao:
I'll do that!

eddywashere:
@LiJinyao released these components in `3.9.0`

PR #236, opened 2016-11-24
ListGroup doc #185

LiJinyao:
#185  Add ListGroup doc.

**API Change in ListGroupItem component:**
When I write doc, I found that using `color="success"` to set contextual class is not a good idea.

The `color` prop is no longer support, when you want to set a contextual class, just pass the name (one of “success”, “info”, “warning”, “danger”) to the component.

Since this component is not exists in doc yet and just released a few days, I think the change is acceptable.

### DIFF
--- a/docs/lib/Components/ListGroupPage.js
+++ b/docs/lib/Components/ListGroupPage.js
@@ -81,17 +81,6 @@ export default class ListGroupPage extends React.Component {
             {ListGroupCustomContentExampleSource}
           </PrismCode>
         </pre>
-
-
-        <h4>Container Properties</h4>
-        <pre>
-          <PrismCode className="language-jsx">
-{`Container.propTypes = {
-  fluid:  PropTypes.bool
-  // applies .container-fluid class
-}`}
-          </PrismCode>
-        </pre>
       </div>
     );
   }
--- a/docs/lib/Components/ListGroupPage.js
+++ b/docs/lib/Components/ListGroupPage.js
@@ -3,7 +3,18 @@ import React from 'react';
 import { PrismCode } from 'react-prism';
 import Helmet from 'react-helmet';
 import ListGroupExample from '../examples/ListGroup';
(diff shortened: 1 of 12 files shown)


eddywashere:
@LiJinyao why remove `color` prop when that's a standard prop reactstrap uses for applying color classes? It's a good approach for a single choice based prop. We normally use boolean props to apply one off classes, classes that can be combined or used to apply a safe default.

LiJinyao:
@eddywashere Oh I haven't notice that it is a standard prop. I'll change it back.
Why I make this change is `color="info"`  needs more type then just a `info` on off prop. But after your review I found that it is not convenient for generating color props.

docs/lib/Components/ListGroupPage.js
```diff
@@ -0,0 +1,98 @@
...
+          <ListGroupCustomContentExample />
+        </div>
+        <pre>
+          <PrismCode className="language-jsx">
+            {ListGroupCustomContentExampleSource}
+          </PrismCode>
+        </pre>
+
+
+        <h4>Container Properties</h4>
```
eddywashere: Can you remove the `container` docs from this

src/ListGroupItem.js
```diff
@@ -33,6 +33,12 @@ const ListGroupItem = (props) => {
     'list-group-item'
   );
 
+  // Prevent click event when disabled.
+  if (disabled) {
```
eddywashere: I was going to ask to convert this to a normal component to make use of instance methods like [onClick in Button](https://github.com/reactstrap/reactstrap/blob/master/src/Button.js#L32-L41) but this is basically the same thing. Neat! I believe a new function is assigned each time it is rendered as disabled. Can you move this logic to a `const handleDisabledOnClick` that's assigned outside of the `ListGroupItem` function.

LiJinyao: OK，I'll do it as soon as possible :)

PR #244, opened 2016-12-01
feature(modal): make z-index configurable

### LINKED ISSUE #243: Hardcoded `zIndex: 1000`on Modal
### Issue description

- version `#3.9.1`
- components: `Modal.js`

In the following [commit](https://github.com/reactstrap/reactstrap/commit/9ddeb8a0be17eddddac278490f9b278096494781)  a hardcoded `zindex = 1000` was added to the root element of the modal that is created in `Modal.js`. This can be a problem for users that have elements above `z-index 1000` since the modal itself and the backdrop will be underneath these elements. We have solved this temporarily by using the `3.9.0` release, but it would be nice if the hardcoded `1000` on line `129` in `Modal.js` can be made configurable or the issue can be resolved in some other way.

TheSharpieOne:
Resolves #243

### DIFF
--- a/docs/lib/Components/ModalsPage.js
+++ b/docs/lib/Components/ModalsPage.js
@@ -49,7 +49,12 @@ export default class ModalsPage extends React.Component {
     PropTypes.bool,
     PropTypes.oneOf(['static'])
   ]),
-  keyboard: PropTypes.bool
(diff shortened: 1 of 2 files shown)


eddywashere:
Thanks!

PR #254, opened 2016-12-17
Pass through unused props in Modal

DanielHuisman:
Currently it's not possible to for example change the size (excluding default Bootstrap sizes) of modals. This change allows users to style the modal dialog.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -1,6 +1,7 @@
 import React, { PropTypes } from 'react';
 import ReactDOM from 'react-dom';
 import classNames from 'classnames';
+import omit from 'lodash.omit';
 import TransitionGroup from 'react-addons-transition-group';
 import Fade from './Fade';
 import {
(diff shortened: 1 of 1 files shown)


TheSharpieOne:
My 2 cents on this:
There is a size prop. Bootstrap comes with default sizes, you can override these. You can make your own custom sizes. Make a modal-my-size class and make it the size you want then set the size prop to my-size. There isn't really much need for the styles since there is bootstrap css framework  and if you are not using bootstrap then you may want to find another library.

DanielHuisman:
This patch has more than just that specific use case. I think it's important for a library, such as reacststrap, to support both CSS and inline styles. Almost all other components support passing style elements (some through `...otherProps`), so why not support it here?

TheSharpieOne:
So... Apply the other props here as well

DanielHuisman:
Changed it.

eddywashere:
This is a great improvement, thanks @DanielHuisman!

eddywashere:
released in `3.9.5` and  `4.0.0-alpha.5`

PR #271, opened 2017-01-06
adding missing and cleaning up cssModule

skoranga:
Some files where missed the update during https://github.com/reactstrap/reactstrap/pull/206. Patch for fixing the cssModule logic for the same.


eddywashere:
thanks!

PR #281, opened 2017-01-08
Update ModalHeader close button

gthomas-appfolio:
Recent changes to Bootstrap alpha use flexbox to position ModalHeader
title and close button, which cause the current code to incorrectly
position title and close on the opposite and wrong side.


eddywashere:
released in `reactstrap@4.0.0`

PR #283, opened 2017-01-11
Col - fix disabled width props from being passed to div

hosmelq:
Sometimes you don't want the `col` class by default, and setting xs={false} works but i got this error

`
Unknown prop 'xs' on <div> tag. Remove this prop from the element.
`

This prop is not removed from attributes.

https://github.com/reactstrap/reactstrap/blob/master/src/Col.js#L63

### DIFF
--- a/src/Col.js
+++ b/src/Col.js
@@ -29,7 +29,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-    xs: true
+  xs: true
 };
 
 const getColumnSizeClass = (isXs, colWidth, colSize) => {
--- a/src/Col.js
+++ b/src/Col.js
@@ -79,7 +79,7 @@ const Col = (props) => {
   });
 
   if (attributes['xs']) {
-      delete attributes['xs'];
+    delete attributes['xs'];
   }
 
   const classes = mapToCssModules(classNames(
--- a/src/Col.js
+++ b/src/Col.js
@@ -53,14 +53,14 @@ const Col = (props) => {
   colWidths.forEach(colWidth => {
     const columnProp = props[colWidth];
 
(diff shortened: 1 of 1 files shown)


src/Col.js
```diff
@@ -28,10 +28,6 @@ const propTypes = {
   cssModule: PropTypes.object,
 };
 
-const defaultProps = {
```
TheSharpieOne: `defaultProps` is used below

hosmelq: My bad, thanks.

src/Col.js
```diff
@@ -78,6 +78,10 @@ const Col = (props) => {
     }
   });
 
+  if (attributes['xs']) {
```
eddywashere: I'd be happy to add back support for `false` as I think `Col` should have some sort of default that takes care of the typical use case. I recently and foolishly moved `delete attributes[colWidth];` after the prop check, it should instead revert back to the following in `colWidths.forEach(colWidth => {`:

```
// always clear out width props from attributes
delete attributes[colWidth];

if (!columnProp) {
  return;
}
```

hosmelq: Do you already moved this ?

eddywashere: It was previously before the check, (see commit](https://github.com/reactstrap/reactstrap/pull/267/files#diff-771c8561c9896cbffb7d4b7f17506333)

image

src/Col.js
```diff
@@ -29,7 +29,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  xs: true
+    xs: true
```
eddywashere: nit: can you remove the extra spaces?

Issue #278, opened 2017-01-07
Proper classes not added for fixed navbar


corydeppen:
It looks like the `navbar-fixed-[bottom|top]` class names were shortened to just `fixed-[top|bottom]` in a recent Bootstrap commit ([see docs example](https://github.com/twbs/bootstrap/commit/4dc6f44796c2b73d40f72fff0f809cc3f87b7343#diff-454f6074dd45826f6593b7353c9d1eb4)). The `Navbar` component is still generating the old class names when the `fixed` prop is `top` or `bottom`.

I also wondered what your thoughts were on modifying the PropType for `fixed` to use the more specific `oneOf(['bottom', 'top'])` instead of `string`.

I'd be happy to submit a PR if you'd like.

- version `#4.0.0-alpha.8`
- components: `Navbar`

TheSharpieOne:
For the oneOf thing, what about left and right? You may be thinking "those values are not in bootstrap" but bootstrap is just a framework which lets you customize it in many different ways, building upon their common defaults. Restricting the props restricts developers which use this library. React-bootstrap restricts the props (or at least used to, they may have changed, but unlikely) and forces the developer to register new/different variants which is silly and painful, not to mention a waste of space/code when trying to keep packages small. Unless you want your site to obviously use bootstrap, you customize the colors. Some add more custom variants, adding variants rather than just changing predefined ones. For some companies 5 colors are not enough others may want a left nav that is fixed. We shouldn't add more code to restrict others?

corydeppen:
That's a fair point. I just look at `propTypes` as a way of documenting components and guiding devs when a prop is missing or used incorrectly.

Would the following be a reasonable alternative that would both document the default Bootstrap options as well as allow for custom positions?

```js
oneOfType([
  PropTypes.oneOf(['bottom', 'top']),
  PropTypes.string
])
```

TheSharpieOne:
Bootstrap does a good job of documenting itself. Most people probably look there first. It's not like people are going to be using a navbar, and know there is a fixed property as well as what the property does without know what values to supply. If people do not know, then bootstrap is where they should look. The docs for this library are also pretty good, though I am not sure if they cover ever variant like bootstrap does.
My point is it's more likely that people will not know what is possible than not know how to do something and this extra code isn't going to solve that.

eddywashere:
[...]

PR #289, opened 2017-01-11
Navbar fixed + sticky classes

eddywashere:
Navbar

- update fixed class name
- add support for sticky class


eddywashere:
fixes #278

Issue #277, opened 2017-01-07
Events for Collapse component


j-francisco:
It would be useful if the Collapse component could accept props like onOpened (or onShown), onClosed (or onHidden), etc. An example of needing this is I'm building a feature in my app where when a Collapse component is opened, I want to scroll the header of the Collapse to the top of the screen. This causes some trouble because I need to perform the scroll after the Collapse has finished opening, otherwise the scroll position would be off.

I took a crack at implementing an onOpened prop [here](https://github.com/j-francisco/reactstrap/tree/collapse-on-opened-callback) but because I don't know if you guys have greater plans for how to handle events across the library, I figured it's not worth working on a PR right now.

eddywashere:
We haven't really had to expose lifecycle events but Collapse is a unique component that could really use this. I'd expect props to be name `on${Event}` for the different transition states. Could you rebase the pr for v4 from master?

j-francisco:
I'll try to get to it soon. Should there be 4 new props: onShow, onShown, onHide, onHidden?

eddywashere:
@j-francisco ah good point, I should have said only `onOpened` and `onClosed`, since they are the only events not currently exposed. `onOpen` and `onClose` are the basic state transition events that can be checked in toggle method. If you find the latter critical to expose, feel free to do it. Thanks for your patience and help!

j-francisco:
@eddywashere should the onOpened and onClosed events be triggered only on updates, or should they also be triggered on the initial render based on the isOpen prop? I'm leaning towards only on updates, since the main reason for these events is because of the delay, which only happens on updates.

eddywashere:
@j-francisco haven't published this just yet but will update this thread when it gets published.

iamkratos:
Was this added? Any chance we could props like this to the Modal component?

j-francisco:
This was added to the Collapse component. The props have since been renamed to `onEntered` and `onExited`.  I believe Modal has `onOpened` and `onClosed`.

PR #301, opened 2017-01-20
feat(Collapse): Add onOpened and onClosed events (#277)

j-francisco:
Allow passing in onOpened and onClosed props as functions that are triggered after the Collapse component has finished opening or closing. Useful for performing an action that relies on the new height of the page after a Collapse component has opened or closed (such as scrolling).

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -31,14 +31,13 @@ const defaultProps = {
   isOpen: false,
   tag: 'div',
   delay: DEFAULT_DELAYS,
+  onOpened: () => {},
+  onClosed: () => {},
 };
 
 class Collapse extends Component {
   constructor(props) {
     super(props);
-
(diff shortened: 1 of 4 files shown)


src/Collapse.js
```diff
@@ -78,10 +84,36 @@ class Collapse extends Component {
...
+    }
+  }
+
   componentWillUnmount() {
     clearTimeout(this.transitionTag);
   }
 
+  onOpened() {
+    if (this.props.onOpened) {
+      this.props.onOpened();
```
TheSharpieOne: What do you think about defaulting these props to noop? (`() => {}`) then there would not be a need for these helper function nor the need to bind them in the constructor.

j-francisco: Makes sense, I just pushed the change.

PR #303, opened 2017-01-22
Return event to toggle function

SteveByerly:
Return the click event to passed-in toggle function so that state can be updated dynamically. This would allow you to set a property (say href) on the component and update state by the target's attribute.

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -121,7 +121,7 @@ class Dropdown extends React.Component {
       return;
     }
 
-    this.toggle();
+    this.toggle(e);
   }
 
   handleProps() {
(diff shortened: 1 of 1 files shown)


TheSharpieOne:
Ran into this myself, while this change is needed, it is probably a good it enough. The event isnt passed https://github.com/reactstrap/reactstrap/blob/master/src/DropdownToggle.js#L52
My workaround without a code change was to put use onClick on the the dropdown toggle to get to the event.

eddywashere:
I don't think that toggle is always called with an event object. At least the external toggle methods passed in. Programmatically controlling components doesn't require it, but I could see it being useful. If this is critical, or other components don't have this onClick work around, I'd suggest updating the pr to add the support to the other components for consistency.

bherila:
> set a property (say href) on the component

Where would you put the `href` attribute exactly?

SteveByerly:
@eddywashere @TheSharpieOne thanks for looking. I'll update my PR accordingly.

eddywashere:
closing this out. feel free to reopen when needed

PR #307, opened 2017-01-26
feat(Uncontrolled): add various uncontrolled components

TheSharpieOne:
`Alert` (dismissible), `ButtonDropdown`, `Dropdown`, `NavDropdown`, and `Tooltip` now have uncontrolled counterparts `UncontrolledAlert`, `UncontrolledButtonDropdown`, `UncontrolledDropdown`, `UncontrolledNavDropdown`, and `UncontrolledTooltip` respectfully.
These uncontrolled components can be used without the user/developer needing to control the state. They will manage their own state of when to show/hide based on the common use-case. The original components are still there for the more advanced use and non-common use-cases.
The goal of these new components is to make it easier for out-of-the-box development.

Documentation for the various components have been added to the end of each of their respective pages.

### DIFF
--- a/src/Uncontrolled.js
+++ b/src/Uncontrolled.js
@@ -1,0 +1,55 @@
+import React, { Component } from 'react';
+import Alert from './Alert';
+import ButtonDropdown from './ButtonDropdown';
+import Dropdown from './Dropdown';
+import NavDropdown from './NavDropdown';
+import Tooltip from './Tooltip';
+
+const components = {
+  UncontrolledAlert: Alert,
+  UncontrolledButtonDropdown: ButtonDropdown,
+  UncontrolledDropdown: Dropdown,
+  UncontrolledNavDropdown: NavDropdown,
+  UncontrolledTooltip: Tooltip,
+};
+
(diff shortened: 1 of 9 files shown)


src/Uncontrolled.js
```diff
@@ -0,0 +1,55 @@
...
+    toggle() {
+      this.setState({ isOpen: !this.state.isOpen });
+    }
+
+    render() {
+      return <Tag isOpen={this.state.isOpen} toggle={this.toggle} {...this.props} />;
+    }
+  }
+
+  SimpleComponent.displayName = key;
```
eddywashere: nice

eddywashere:
Thanks! It kind of highlights the api differences between the components added and ones that could be added, Popover & Modal. I don't think it would be practical to have Tooltip require onMouseEnter, onMouseLeave + using toggle and isOpen on the target component. Maybe Tooltip could have a prop to enable current auto hover events and Popover could have a similar prop to enable auto hover handlers. I'll create an issue to track/discuss it. Thanks again!

PR #309, opened 2017-01-27
feat(Col): add custom widths ability to Col

### LINKED ISSUE #297: Add props for custom widths
I use the _custom.scss file to add additional widths, e.g. xxl. It would be nice if all the components supported this, e.g. `<Col xxl="4"></Col>`

This will mean supporting dynamic props names. Not sure how this is achieved or even if it's possible. I'll do some research.

TheSharpieOne:
closes #297

### DIFF
--- a/src/Col.js
+++ b/src/Col.js
@@ -26,10 +26,11 @@ const propTypes = {
   xl: columnProps,
   className: PropTypes.string,
   cssModule: PropTypes.object,
+  widths: PropTypes.array,
 };
 
 const defaultProps = {
-  xs: true
+  widths: colWidths,
 };
 
 const getColumnSizeClass = (isXs, colWidth, colSize) => {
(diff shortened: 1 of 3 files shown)


src/__tests__/Col.spec.js
```diff
@@ -22,6 +22,23 @@ describe('Col', () => {
...
 
+  it('should allow custom columns to be defined', () => {
+    const wrapper = shallow(<Col widths={['base', 'jumbo']} base="4" jumbo="6" />);
+
+    expect(wrapper.hasClass('col-4')).toBe(true);
+    expect(wrapper.hasClass('col-jumbo-6')).toBe(true);
+  });
+
+  it('should allow custom columns to be defined with objects', () => {
+    const wrapper = shallow(<Col widths={['base', 'jumbo', 'wtf']} wtf={{ size: 1, push: 2, pull: 3, offset: 4 }} />);
```
eddywashere: 😆 just happy to have tests for this

src/Col.js
```diff
@@ -46,20 +47,25 @@ const Col = (props) => {
...
+    }
 
     delete attributes[colWidth];
 
     if (!columnProp) {
       return;
     }
 
-    const isXs = colWidth === 'xs';
+    const isXs = !i;
```
eddywashere: very clever!

TheSharpieOne: Yeah, but I probably should have just went with `i === 0` or something to be more clear. That's my bad.

PR #321, opened 2017-02-01
fix(Modal): unmounting nested modals

hongrich:
Unmounting nested modals should remove the matching number of
`modal-open` classes from the body node. Previously, the updated class
names were computed before its children are unmounted and only updated
in the DOM afterwards. This can cause `modal-open` class name to be left
on the body if any of its children is also a modal (which is also trying
to remove those `modal-open` classes).

Here's an demo of the issue: http://www.webpackbin.com/4J7BkUsPf
1. Click on `Example 2`
2. Click on `Show Nested Model`  <- typo in the example!
3. Click on `All Done` which unmounts the example 2 component
4. Note that there is still a `modal-open` class left on the body node

### DIFF
--- a/src/__tests__/Modal.spec.js
+++ b/src/__tests__/Modal.spec.js
@@ -1,14 +1,21 @@
 import React from 'react';
 import { mount } from 'enzyme';
-import { Modal } from '../';
+import { Modal, ModalBody } from '../';
 
 describe('Modal', () => {
   let isOpen;
   let toggle;
 
+  let isOpenNested;
+  let toggleNested;
+
   beforeEach(() => {
     isOpen = false;
     toggle = () => { isOpen = !isOpen; };
+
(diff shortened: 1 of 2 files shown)


src/__tests__/Modal.spec.js
```diff
@@ -396,4 +403,27 @@ describe('Modal', () => {
...
+            Yo!
+          </Modal>
+        </ModalBody>
+      </Modal>
+    );
+
+    jasmine.clock().tick(300);
+    expect(wrapper.children().length).toBe(0);
+    expect(document.getElementsByClassName('modal-dialog').length).toBe(2);
+    expect(document.body.className).toBe('modal-open modal-open');
```
eddywashere: will need to follow up with a proper fix for duplicate classes by either tracking active modals or using a prop as a flag for managing that body class.

Issue #323, opened 2017-02-08
Why is the toggle prop required on Modal?


bherila:
Just wondering why the toggle prop is marked as required on the Modal component? Any objections to a PR which would make this functionality optional?

TheSharpieOne:
It's for when the user clicks on the backdrop and/or uses the keyboard (esc) to close the modal. But yes, those things can be turned off / disabled, in which case the toggle prop should not be required.
Also if toggle is not passed, it should be the same as if the close on keyboard and close on backdrop click there false.

bherila:
Ok thanks @TheSharpieOne , I've sent a PR.

PR #325, opened 2017-02-08
Make toggle prop optional on Modal

bherila:
re #323


bherila:
@eddywashere what do you think?

eddywashere:
+1 to this specifically. Will hold off on merging until I can ship a release and avoid any confusion with what's in master and what latest on npm

bherila:
Good call @TheSharpieOne , I fixed that.

PR #331, opened 2017-02-12
feat(Popover): closeOnClick property

cjvirtucio87:
Added closeOnClick property to Popover component:
- closeOnClick affects Popover behavior when toggler is clicked
- if true, Popover closes when toggler is clicked
- old behavior is the default


TheSharpieOne:
Just FYI, you can push additional changes to an existing PR, there is no need close and create new ones.

docs/lib/examples/Popover.js
```diff
@@ -24,7 +24,7 @@ export default class Example extends React.Component {
         <Button id="Popover1" onClick={this.toggle}>
           Launch Popover
         </Button>
-        <Popover placement="bottom" isOpen={this.state.popoverOpen} target="Popover1" toggle={this.toggle}>
+        <Popover placement="bottom" isOpen={this.state.popoverOpen} closeOnClick={ true } target="Popover1" toggle={this.toggle}>
```
TheSharpieOne: For simplicity and best practices, just adding the `closeOnClick` prop with no attribute/value will work just the same.

src/TetherContent.js
```diff
@@ -26,7 +28,10 @@ class TetherContent extends React.Component {
     super(props);
 
     this.handleDocumentClick = this.handleDocumentClick.bind(this);
+    this.checkTogglerTarget = this.checkTogglerTarget.bind(this);
```
TheSharpieOne: This does not appear to be needed since `this.checkTogglerTarget` only appears to be called from `this.handleDocumentClick` which will always have the correct context of `this`

docs/lib/examples/Popover.js
cjvirtucio87: It works even without passing a boolean. Edited the example to show this more clearly.

src/TetherContent.js
cjvirtucio87: Good catch. Removed.

cjvirtucio87:
@TheSharpieOne Not sure how else I can have the Popover remember that it's already been clicked. Basically, the issue I had was that handleDocumentClick would fire after having clicked the toggler to close the Popover. So the toggler has to remember that it's been clicked to prevent handleDocumentClick from firing again.

I'll write a better test.

TheSharpieOne:
It doesn't need to manage state in order to do what you are trying to do. In fact, I believe managing the state in the way you are doing it causes a defect. The defect is: when you click on the popover [content] itself and then on the toggle/button it will not close.
I was able to remove the need from state altogether, which also fixed that defect. You can see this working: http://www.webpackbin.com/4JFv0e2OM

src/TetherContent.js
```diff
@@ -66,15 +71,34 @@ class TetherContent extends React.Component {
     return config;
   }
 
+  checkTogglerTarget(togglerID, clickTarget) {
+    if (clickTarget && clickTarget.id) return clickTarget.id === togglerID.slice(1);
```
TheSharpieOne: Also, I would use `substring` as it is slightly more performant than `slice`

cjvirtucio87:
Ah, that's much simpler. Broke it out into separate methods.

src/TetherContent.js
```diff
@@ -28,8 +28,9 @@ class TetherContent extends React.Component {
     super(props);
 
     this.handleDocumentClick = this.handleDocumentClick.bind(this);
-    this.checkTogglerTarget = this.checkTogglerTarget.bind(this);
     this.toggle = this.toggle.bind(this);
+    this.checkTogglerTarget = this.checkTogglerTarget.bind(this);
```
cjvirtucio87: Needed to bind these methods, otherwise they'd get called multiple times in the tests.

TheSharpieOne: They shouldn't need to be bound. Maybe there is a way to fix the test instead of changing the code?

cjvirtucio87: I just removed the tests that check how many times shouldToggle gets called. Without the bindings, shouldToggle gets called 11 times throughout the specs.

eddywashere:
It looks like we need to standardize and abstract the `getTarget` functionality into utils. Right now popover/tooltip have different ways of getting target. To future proof `getTarget` as much as possible it should support a css selector (string), a dom reference or a function that returns a dom reference. Once that's in it will be easier to check if target matches click target. going to close this out until that happens.

PR #338, opened 2017-02-23
fix(Nav): update valid properties

### LINKED ISSUE #295: Fix Nav stacked option
### Issue description

- version `#4.0.1`
- components: `Nav`

### Steps to reproduce issue
Just create a `<Nav stacked>...</Nav>` component => items are not stacked.

In code we see it results in `nav-stacked` class being added: https://github.com/reactstrap/reactstrap/blob/master/src/Nav.js#L41

This produces no desired affect as current BS4@alpah6 documentation suggests adding a general 'flex-column' utility class:
http://v4-alpha.getbootstrap.com/components/navs/#vertical

jebarjonet:
Looking at the doc of Bootstrap v4 alpha.6, `Nav` needed some adjustments:

- Remove properties `inline`, `stacked` and `disabled` (they don't exist in either `nav` or `navbar`)
- Add property `vertical`
- Update doc accordingly
- Fix #295 

screen shot 2017-02-22 at 21 36 53

### DIFF
--- a/docs/lib/Components/NavsPage.js
+++ b/docs/lib/Components/NavsPage.js
@@ -4,8 +4,8 @@ import { PrismCode } from 'react-prism';
 import Helmet from 'react-helmet';
 import NavsExample from '../examples/Navs';
 const NavsExampleSource = require('!!raw!../examples/Navs');
-import NavInlineExample from '../examples/NavInline';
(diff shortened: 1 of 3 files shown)


eddywashere:
thanks!

Issue #168, opened 2016-10-03
How to disable Fade on modal?


lawlmart:
Did I miss this in the docs? Didn't see it anywhere.

eddywashere:
It's not possible at the moment. It should be added. I'll keep this open to track that.

nndung179:
Could I take this enhancement @eddywashere ?

eddywashere:
@nndung179 all yours 👍

sidekickstudio:
Can i change the Modal animation when i press close button? some example? thanks

mrchess:
+1. I'm also trying to remove the animation. Anyone know how to do it programmatically?

cristian-sima:
Why does it take so long to add this feature?

kheang:
+1.  It looks like https://github.com/reactstrap/reactstrap/pull/339 by @annejohnson addresses this issue nicely.

eddywashere:
added in 4.6

PR #339, opened 2017-02-24
feat(Modal): Support fade and timeout props in the Modal component to allow configuring + disabling of the fade effect

annejohnson:
This PR adds support to the `Modal` component for a `fade` prop of type boolean (default value: true). It also adds support for modal+backdrop appear/enter/leave timeout props.
It addresses the following issues: https://github.com/reactstrap/reactstrap/issues/168 & https://github.com/reactstrap/reactstrap/issues/352

If a value of false is passed for `fade`, the Bootstrap modal appears instantaneously, without a fade transition. Likewise when the modal closes.

The timeout props control the speed of the fade effect for the modal and the backdrop.

Also included in this PR are test cases and an example in the documentation.


src/Modal.js
```diff
@@ -166,59 +198,70 @@ class Modal extends React.Component {
...
+    const modalAttributes = {
+      onClickCapture: this.handleBackdropClick,
+      onKeyUp: this.handleEscape,
+      style: { display: 'block' },
+      tabIndex: '-1'
+    };
+
+    if (this.hasTransition()) {
+      const modalTransitionTimeout = 300;
+      const backdropTransitionTimeout = 150;
```
TheSharpieOne: Should these times be configurable, or should that be another PR. At a certain point we could make the transition timing variable and when it is 0 disabled the transition animation.

annejohnson: I thought that was a good idea. I added that in this commit: https://github.com/reactstrap/reactstrap/pull/339/commits/a2fb9154d7da44b9540d888472193a591da7ed6d

Let me know what you think. Thank you for taking a look!

annejohnson: I've pushed another commit after seeing issue https://github.com/reactstrap/reactstrap/issues/352. That person wants the ability to separately configure appear vs. enter vs. leave timeouts.

It's a lot of new props, but all optional ones. What do you think?
https://github.com/reactstrap/reactstrap/pull/339/commits/ab93b7dd1ec26c0ab29fc6b6a0d03dbc6ca7e01c

theduke:
Could this be merged?

I really need this.

eddywashere:
@annejohnson I unfortunately messed up by not merging this first. A different pr merge caused a conflict with this branch. Can you resolve the conflict and commit that change? Thanks and sorry about that!

annejohnson:
@eddywashere I resolved the conflict. Sorry for the delay. Thanks for taking a look at these updates!

vstlouis:
Can we merge this please? Seems like its solid.

PR #346, opened 2017-02-28
Add toggle switch to DropdownItem to allow conditional toggle

CoderJoeIO:
In the current master the dropdown toggle is called whenever a DropdownItem is clicked. This interferes with using the Dropdown for multiselect purposes. My proposal adds a toggle switch that defaults to true, so that the current default behaviour stays the same, but an optional toggle prop can be set to {false} which then skips the call to this.context.toggle and allows the menu to stay open on DropdownItem onClick.


src/DropdownItem.js
```diff
@@ -40,7 +43,9 @@ class DropdownItem extends React.Component {
       this.props.onClick(e);
     }
 
-    this.context.toggle();
+    if (this.props.toggle) {
+      this.context.toggle();
```
eddywashere: Rethinking this, bool seems fine for preventing toggle. But, if a function is passed to `toggle` prop, we should probably call that instead of context. That can be an enhancement pr that should be backwards compatible.

CoderJoeIO: I completely agree, I'll do this and create a new PR as soon as I have some time.

Issue #337, opened 2017-02-21
Tooltip `target` requires string, but Tether takes object? 


mawburn:
version: 4.1.1
components: Tooltip 

I'm not sure if this is intended functionality or not, I might be overlooking it, but I can't find anything in the issues concerning this. 

But when using the tooltip it requires a target ID **string**, however Tether can take a DOM Object, ID, etc. I can get around this by giving it a junk ID that I threw on some html in the top level component, then setting the target directly inside the tether prop using the `event.target` coming from a child component. 

eg: 

```JSX
<Tooltip
    isOpen={this.state.tooltipOpen}
    target="junkidcuzwhousesidsinreact"
    tether={{target: this.state.tooltipTarget}}
    toggle={this.toggle}
>
    I'm a tooltip.
</Tooltip>
```

This seems very hacky solution to me and it feels like target should accept the same proptypes that Tether can. 

>target: The DOM element, jQuery element, or a selector string of an element which the element will be attached to

http://tether.io/#options

eddywashere:
Nice find! I agree that's how it should work. Unfortunately I simplified that target prop too much by coercing it into an id selector. Changing it would be a breaking api change that would affect everyone using those components (Tooltip & Popover). I would like to switch out tether for popper.js soon and would rather use that as a reason for a breaking change. I think the new prop would be `reference` but not sure just yet.

For now, I'd happily accept a pr that makes the target prop optional.

mawburn:
You think it would be a breaking change? 

I did this and all your tests still passed. I haven't dug much deeper than this or tested it against my code yet, though. 

```JavaScript
target: PropTypes.oneOfType([
  PropTypes.object,
  PropTypes.string
]).isRequired,
```

```JavaScript
getTetherConfig() {
  const attachments = getTetherAttachments(this.props.placement);
  const target = (typeof this.props.target === 'object') ? this.props.target : `#${this.props.target}`;

  return {
    ...defaultTetherConfig,
    ...attachments,
    target,
    ...this.props.tether
  };
}
```

eddywashere:
> however Tether can take a DOM Object, ID, etc.

[...]

PR #356, opened 2017-03-14
feat(Tooltip): Tooltip target element option

nlrowe:
Added the option to pass a DOM element to the tooltip target. The 'getTarget' function was also used for the tether config target to maintain post component mounting behavior. If this is not a concern, it can just reference this._target that is created on mount.

Addresses [#337](https://github.com/reactstrap/reactstrap/issues/337)


mawburn:
This is the same concern. @nlrowe and I work together.

src/Tooltip.js
```diff
@@ -113,12 +116,20 @@ class Tooltip extends React.Component {
...
+    return document.getElementById(target);
+  }
+
   getTetherConfig() {
     const attachments = getTetherAttachments(this.props.placement);
     return {
       ...defaultTetherConfig,
       ...attachments,
-      target: '#' + this.props.target,
+      target: this.getTarget(),
```
eddywashere: I forgot that I used the string selector because I wasn't sure if the target element would be available at the time getTetherConfig is called. Tether also accepts a function as target. This might be safer as:

```
target: this.getTarget
```

`this._target = this.getTarget()` in `componentDidMount` should be fine since it's mounted.

eddywashere:
Good callout for `this._target`. It should be fine to just reference that 👍

nlrowe:
If we change `getTetherConfig` to use `this._target`, the component fails if it is initialized with `isOpen` set to true as the `render` actually executes with an undefined `this._target` when it is called before the `componentDidMount`. 

The check in `render` can be modified to handle for that:

```JS
    if (!(this.props.isOpen && this._target)) {
      return null;
    }
```

Or, we can use the function and it will go get the element on its own.

Which would you prefer good sir?

eddywashere:
@nlrowe ah I was hoping the timing/lifecycle would be fine. Good to know, I think we should pass in the function instead. Thanks for working through this!

PR #361, opened 2017-03-15
style(Badge): Standardising tag assignment name

mesiu:
Just a minor tweak for consistency. :)


TheSharpieOne:
Note: This was just a carryover from when `Tag` was the name of the component/function and thus was not available for the variable name.

PR #364, opened 2017-03-16
Limit babelrc

balloob:
While working on the Rollup stuff I noticed that Reactstrap uses Babel stage-0 preset but none of the options are in use.

From my experience I can recommend to not use any Babel presets that are still very much in flux like stage-0. The next release of Babel might have different stage-0 features or the stage-0 APIs introduced breaking changes.

By looking at the code I realized that the only non-standard JavaScript feature that is being used is object spread. So this PR removes stage-0 and replaces it with just object-spread.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,6 @@
 {
-  "presets": ["es2015", "stage-0", "react"]
+  "presets": ["es2015", "react"],
+  "plugins": [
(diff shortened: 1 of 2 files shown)


eddywashere:
Nice! Was about to ask if there was any impact to non-modern browsers and I just realized the 2015 preset takes care that. lgtm

PR #370, opened 2017-03-23
Fix error when running npm start

miguelc1221:
Webpack was failing to compile when running `npm start`. The error was being caused by an arrow function class property in a component. Class properties are stage-2, so I added `babel-plugin-transform-class-properties` to dev dependencies. 

I also went ahead and upgraded `babel-preset-react-app` which removes the following deprecated warning
```
npm WARN deprecated babel-preset-latest@6.14.0: 💥  preset-latest accomplishes the same task as babel-preset-env. 🙏  Please install it with 'npm install babel-preset-env --save-dev'. '{ "presets": ["latest"] }' to '{ "presets": ["env"] }'. For more info, please check the docs: http://babeljs.io/docs/plugins/preset-env 👌. And let us know how you're liking Babel at @babeljs on 🐦
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,7 @@
 {
   "presets": ["es2015", "react"],
   "plugins": [
-    "transform-object-rest-spread"
+    "transform-object-rest-spread",
+    "transform-class-properties"
   ]
 }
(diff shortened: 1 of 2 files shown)


balloob:
Sorry, that was my bad. I have put in a PR to just get rid of the syntax as it's a 1 line change: https://github.com/reactstrap/reactstrap/pull/371

miguelc1221:
@balloob no problem

PR #373, opened 2017-03-29
pass in cssModule to Fade in Modal

jalbinson:
Modals would not work when a cssModule was passed in because the Fade component's styles wouldn't be pushed through `mapToCssModules` correctly.


eddywashere:
Thanks for the fix!

PR #379, opened 2017-03-30
[DO NOT MERGE] Experiment: Use cherry picked lodash modules

balloob:
So as a quick experiment to see if I could reduce the bundle size even more, I tried to replace the lodash dependencies with cherry picked functions from the lodash lib. The idea was that we would be able to dedupe overlapping code between the individual lodash packages.

Result: did not reduce bundle size.

Leaving PR here so people can learn from the results.

It is caused because Rollup can't 100% decide if some code is side-effect free. It will caution on the safe side and end up including more then is needed. [See Rollup wiki](https://github.com/rollup/rollup/wiki/Troubleshooting#tree-shaking-doesnt-seem-to-be-working)

| Type | Size |
| -- | -- |
| Current (function per package) | 111kB
| Import from `lodash/func-name` | 127kB
| Import named from `lodash-es/func-name` | 124kB
| Import named from `lodash` | Forgot, but was > 200kB

screen shot 2017-03-30 at 2 26 48 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,10 +48,12 @@
   "homepage": "https://github.com/reactstrap/reactstrap#readme",
   "dependencies": {
     "classnames": "^2.2.3",
+    "lodash-es": "^4.17.4",
     "lodash.isfunction": "^3.0.8",
     "lodash.isobject": "^3.0.2",
     "lodash.omit": "^4.4.1",
     "lodash.tonumber": "^4.0.3",
+    "source-map-explorer": "^1.3.3",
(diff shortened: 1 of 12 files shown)


TheSharpieOne:
Funny, I was noticing that and thinking the same thing. At my job we recently did the same thing... with the help of [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) though I am honest surprised that "Import named from `lodash-es/func-name`" did not make it smaller.

Issue #382, opened 2017-03-31
<Button> type attribute should default to 'submit', not 'button'


mikeatlas:
Per this recent commit last night https://github.com/reactstrap/reactstrap/commit/ebbeba75f6a05881ba6f5b34bca15c88f71eddf4 by  @TheSharpieOne ...

This caused a regression preventing us from logging into our own application. We were using a single `<Button>` element within a form for a login page, and the default behavior for a button in a form is `submit`. 

Once this change went in, our form stopped ... submitting! Changing our button to `type='submit'` does indeed fix things, but this is different from the [HTML5 button spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button):

> **type**
> The type of the button. Possible values are:
> **submit**: The button submits the form data to the server. This is the default if the attribute is not specified, or if the attribute is dynamically changed to an empty or invalid value.

I realize `Button` in reactstrap vs `button` in HTML5 are not the same, but it would be advisable to align with the HTM5 spec in this case.

h/t @j-francisco (thanks!)

TheSharpieOne:
[Actual spec](https://www.w3.org/TR/html5/forms.html#attr-button-type) which does explicitly say the default is submit.... but the default type for Internet Explorer is "button" (of course)

My only argument against having submit be the default is that is not the most common use-case for the Button in this library. With bootstrap and javascript/react, most people are using them outside of forms and when they are within a form #366 happens where the intent was not to trigger submission.

This was original brought up in slack: https://reactstrap.slack.com/archives/C1WQP1G2C/p1487185000000442
> ![image](https://cloud.githubusercontent.com/assets/664714/24569451/281bf1b0-1634-11e7-993d-6ea8b9428511.png)

The best advice is to always explicitly specify the type attribute for the button, because again, IE it does not default to submit.

Perhaps some middle ground would be only defaulting to button if there is an `onClick`, which indicates that the intent is for the event to be handled by react

> button: The button has no default behavior. It can have client-side scripts associated with the element's events, which are triggered when the events occur.

mikeatlas:
Hm, since you have ability to detect if a reactstrap `Button` is missing a `type` attribute, can you emit a console warning? That might be a suitable compromise.

TheSharpieOne:
[...]

PR #383, opened 2017-03-31
fix(Button): only default to button when click is handled

TheSharpieOne:
fixes: #382

`<Button>Go!</Button>` -> `<button class="btn btn-secondary">Go!</button>`
`<Button onClick={this.myHandler}>Go!</Button>` -> `<button type="button" class="btn btn-secondary">Go!</button>`
`<Button type="submit" onClick={this.myHandler}>Go!</Button>` -> `<button type="submit" class="btn btn-secondary">Go!</button>`

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -70,7 +70,7 @@ class Button extends React.Component {
 
     return (
       <Tag
-        type={Tag === 'button' ? 'button' : undefined}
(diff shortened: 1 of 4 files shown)


eddywashere:
very reasonable fix, thanks!

PR #384, opened 2017-04-02
Fix import/export not getting transpiled in lib dir

balloob:
Rollup conversion introduced a bug where Babel wouldn't transpile import/export to CommonJS syntax in the lib dir. This PR fixes it.

Report: https://github.com/reactstrap/reactstrap/pull/359#issuecomment-290928268

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,6 +7,9 @@
     "transform-object-rest-spread"
   ],
   "env": {
+    "dist-dir": {
+      "plugins": ["transform-es2015-modules-commonjs"]
+    },
     "webpack": {
       "plugins": ["transform-es2015-modules-commonjs"]
     },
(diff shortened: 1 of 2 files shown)


TheSharpieOne:
I would also point main in the package.json to the lib index.js as this would allow for better source maps and such.

balloob:
Nah, the source maps are already off since some transpilation happens. Instead I have added source maps back to the UMD output in https://github.com/reactstrap/reactstrap/pull/386

balloob:
Another issue with the lib dir is that if you import from it, you will also have to add the dependencies of Reactstrap to your own `package.json` (ie `reactstrap-tether`).

TheSharpieOne:
You shouldn't need to, reactstrap-tether is a dependency, not a peer or devDep. It will be downloaded when reactstrap is installed and  things know how to resolve it.

balloob:
@TheSharpieOne ah yes, you're right.

PR #386, opened 2017-04-02
Add UMD builds to Rollup

balloob:
I incorrectly assumed that Webpack was exporting CommonJS syntax. Instead, it was exporting UMD format (which is CommonJS compatible).

So, this updates the Rollup config, also making it backwards compatible with Reactstrap 4.3.0 to address https://github.com/reactstrap/reactstrap/pull/359#issuecomment-290926692.

As extra bonus, the UMD bundles are able to be minified to 95kB! Looking at the source code, it seems like minifiers can do a better job minifying since everything is inside a function.

Changes:

 - Dropped `reactstrap.cjs.js` output
 - Added UMD bundle `reactstrap.min.js` (compatible with CommonJS and equivalent to output in Reactstrap 4.3)
 - Added source maps back

image

### Testing changes

Point all the test files at the UMD bundle:

```bash
sed -i "" "s/\.\.\/\'/\.\.\/\.\.\/dist\/reactstrap.min'/" src/__tests__/*.spec.js
```

Run tests as usual: `npm test`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "reactstrap",
   "version": "4.4.0",
   "description": "React Bootstrap 4 components",
-  "main": "dist/reactstrap.cjs.js",
+  "main": "dist/reactstrap.min.js",
   "jsnext:main": "dist/reactstrap.es.js",
   "module": "dist/reactstrap.es.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


virgofx:
Thanks, looking forward to next release to test 👍

virgofx:
4.5.0 works great and existing CDN reference on main site can remain as is.  Thanks

Issue #335, opened 2017-02-18
Autofocus not working inside Reactstrap Modal 


indapublic:
Autofocus not working if element inside Reactstrap Modal

- version `4.2.0`
- components: `Modal`

### Example

```
import {Modal} from 'reactstrap'

render () {
    const {handleSubmit} = this.props
    return (
      <Modal isOpen={true}>
        <form onSubmit={handleSubmit}>
          <fieldset>
            <input name="email" type="email" label="Email Address" />
            <input name="password" type="password" label="Password" autoFocus={true} />
            <button type="submit">Press me</button>
          </fieldset>
        </form>
      </Modal>
    )
  }
```

TheSharpieOne:
This issue has been resolved in the 4.6.0. Add `autoFocus={false}` to your modal to opt-out of the modal's focus management.
See this example: https://www.webpackbin.com/bins/-Kpuo0BUJpkb53Cy3poH

PR #389, opened 2017-04-04
fix(Modal): Made focus optional

miguelc1221:
This PR addresses #335.

It seems that the Modal has a default autoFocus which is why any input inside the modal won't be autoFocused. I exposed a `isFocused` prop which defaults to true. This way if a user wants to autofocus an input, he/she can remove the autofocus from the modal first.


TheSharpieOne:
My two cents, isFocused is probably not the best name, something like autoFocus would be better.

miguelc1221:
yea you're right, fixed.

oliverasistemas:
What is the status on this, is it going to be merged?

eddywashere:
need to re-review and will ship it

PR #395, opened 2017-04-11
fix(*): import PropTypes from prop-types

apaatsio:
Using the PropTypes via the react package is deprecated as of React
15.5.0.

Use the PropTypes from the prop-types package as recommended.

See https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html


eddywashere:
Awesome! will check this out later

julien-f:
https://github.com/reactjs/prop-types#how-to-depend-on-this-package

Just saying :)

virgofx:
Bump the dependency to minimum of 15.5.7 per guides or better yet even 15.5.8 was released yesterday.

ufocoder:
@apaatsio By the way you `destruct` types like the following:
```
import {string, bool} from 'prop-types';
  		  
const propTypes = {
   color: string,
   pill: bool
}
```

apaatsio:
I updated the dependencies to follow the instructions at https://github.com/reactjs/prop-types#how-to-depend-on-this-package. Thanks @julien-f 

@ufocoder, I can see some benefits in readability but I'm not gonna go for that search-and-replace exercise. If someone wants to do it, go ahead.

apaatsio:
In the peer dependencies we still have `"react": "^0.14.0 || ^15.0.0"`. I'm not confident at all that this will work with such old versions of React. When was the last time someone tested reactstrap with React 0.14.0??

TheSharpieOne:
reactstrap doesn't do complex things, I would be surprised it didn't work with 0.14.0. prop-types on the other hand does not work with 0.14.0, the minimum version is 0.14.9. I would suggest changing the peerDep to `"react": "^0.14.9 || ^15.3.0"`. If someone uses an older version, they will get a warning during npm install (unmeet peer dependency), but it will still allow them to build and use reactstrap. They will also get warning from react as mentioned in the prop-types docs.

apaatsio:
@TheSharpieOne Thanks, I updated the dependencies.

nathancahill:
While we're doing this, we should also update from `react-addons-transition-group` and `react-addons-css-transition-group` to [reactjs/react-transition-group](https://github.com/reactjs/react-transition-group). See note here: https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#discontinuing-support-for-react-addons

nathancahill:
Nevermind, just saw #399 which is open for this.

HaNdTriX:
What's missing? 

@eddywashere can you merge this?

eddywashere:
Thanks @apaatsio!

PR #402, opened 2017-04-23
fix(DropdownToggle): add missing proptype (color)

apaatsio:


virgofx:
This PR is invalid. It breaks HTML and is not BS4 compatible. Color needs to be removed.

apaatsio:
@virgofx Can you elaborate what exactly is breaking? It seems to be working fine for me.

E.g.
```html
<DropdownToggle color="success" caret>Hello</DropdownToggle>
```
generates
```html
<button type="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" class="dropdown-toggle btn btn-success">Hello</button>
```

Moreover, if we remove the color property, how are we supposed to change the color of the DropdownToggle?

virgofx:
Ah, I see this distinction. You were using color for button. It only works when not omitting the tag property. Your fix broke issues in my application where we are not using button :)  So we need a solution that only passes color into dynamic tag attribute **if and only if** the tag is the button.

Issue #80, opened 2016-07-31
Carousel


eddywashere:
http://v4-alpha.getbootstrap.com/components/carousel/
- [ ] Carousel
- [ ] CarouselItem

raavanan:
Hey @eddywashere , I want to do help build this Component.

raavanan:
calling Dibs on this. I have started work.

raavanan:
@eddywashere  I have joined your slack team. I have forked the repo, and am studying the codebase. please ping me on starting work on this.

raavanan:
@eddywashere So here is the breakdown for the carousel component wise.

- [ ] Carousel (container)
- [ ] CarouselIndicators
- [ ] CarouselItem
- [ ] CarouselCaption
- [ ] CarouselControl (left and right)

eddywashere:
Looks good to start to me. I assume the methods would live on the wrapping component and get passed via state or context. Same for options via props?

raavanan:
Yup, that's the plan.

chefnobody:
@raavanan Any chance this Component will be wrapped up soon? I could make use of it asap 😃

raavanan:
@chefnobody I am done with most of the logic. been busy with work lately.. will knock this off this weekend

chefnobody:
@raavanan You're a good man! I'll keep an eye out for a PR, then. 👍

TheSharpieOne:
@raavanan Any update?

nielk:
Hey, can someone finish this component? 😃

louisscruz:
Are there any updates with this? Very much looking forward to a carousel!

Gribbs:
Boooooo! No carousel still

slimeygecko:
@raavanan, It looks like you've come a long way on the carousel. I'd be happy to finish it up for you. Do you have a fork / branch that has the code in it?

mpartipilo:
Please update #13 with this. Thanks :+1:

PR #407, opened 2017-05-01
Carousel implementation

mkalish:
Implementation of the bootstrap 4 carousel for reactstrap

Addresses: #80


mkalish:
Let me know if you would like to re-open this PR with the commits squashed

mkalish:
@TheSharpieOne 
I've made the small changes, but wanted to confirm the structural changes I was going to make before getting to far down that path.  I had originally gone with something like this if this was you were thinking of?


```
class Example extends Component {
  constructor(props) {
    super(props);
    this.handleNext = this.handleNext.bind(this);
    this.handlePrev = this.handlePrev.bind(this);
    this.handleGotoIndex = this.handleGotoIndex.bind(this);
    this.state = { activeIndex: 0, direction: 'right' };
  }

  handleNext() {
    this.setState({activeIndex: this.state.activeIndex + 1, direction: 'right'});
  }

 handlePrev() {
    this.setState({activeIndex: this.state.activeIndex - 1, direction: 'left'});
  }

  handleGotoIndex(idx) {
    this.setState({activeIndex: idx});
  }

  render() {
    const items = [
      {
        src: 'src1',
        altText: 'Slide 1',
        caption: 'Slide 1'
      },
      {
        src: 'src2',
        altText: 'Slide 2',
        caption: 'Slide 2'
      },
      {
        src: 'src3',
        altText: 'Slide 3',
        caption: 'Slide 3'
      }
    ];

  const carouselItems = items.map((item, idx) => {
      return (<CarouselItem
        direction={direction}
        key={idx}
        {...item}
        cssModule={cssModule}
      />);
    });

   const { activeIndex } = this.state;
    return (
     <Carousel>
       <CarouselIndicators  activeIndex={activeIndex} onClickHandler={this.handleGotoIndex} />
       {items[0]}
       <CarouselControl direction="prev" directionText="Previous" onClickHandler={this.handlePrev} />
       <CarouselControl direction="next" directionText="Next" onClickHandler={this.handleNext} />
     </Carousel>

      <Carousel items={items} />
    );
  }
}
```

I can go back to this approach, but it seemed like that this would force a user to have to implement the wrap functionality themselves, since that needs to be handle in the previous and next functions?

TheSharpieOne:
I would defer to @eddywashere, but we could have them wrap, or the component could do a mod (`%`). We can make the children be the slides/items and count the children to determine the number of items. `this.props.index % numOfChildren`. You can also determine the whole direction thing when the props update, comparing the current index to the next index. This would also help with the slide change direction when the index change by more than just 1/-1. Looks like your `handleGotoIndex` doesn't determine the direction like your `handleNext` and `handlePrev` do

mkalish:
@TheSharpieOne  @eddywashere 
I have an alternative implementation in this PR #412 that has the consumer implement much more of rotation including the wrapping functionality.

mkalish:
Closing this as #412 is the preferred implementation

PR #410, opened 2017-05-05
fix(Modal): fix bug where closing modal removed wrong modal-open string in class

markogresak:
Initially, I wanted to just report the issue, but while writing it, I found out that it was an easy fix so I did it myself.
Here is what I wanted to report:

### Issue description

- version: tested on`#4.5.0` but this issue can be found in the master branch
- components: `Modal`

### Steps to reproduce issue

1. add a class to the body which contains`modal-open`, e.g.: `my-modal-opened`
2. add a [Modal](https://reactstrap.github.io/components/modals/) element
3. show the modal, which will add the `modal-open` class to the body, the body class property is now `my-modal-opened modal-open`
4. hide the modal, expecting to return body class to `my-modal-opened`, but instead, the class is now `my-ed modal-open`

Demo: http://codepen.io/anon/pen/pPWVyP

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -122,7 +122,8 @@ class Modal extends React.Component {
       this._element = null;
     }
 
-    const classes = document.body.className.replace('modal-open', '');
+    // Use regex to prevent matching `modal-open` as part of a different class, e.g. `my-modal-opened`
+    const classes = document.body.className.replace(/(^| )modal-open( |$)/, ' ');
     document.body.className = mapToCssModules(classNames(classes).trim(), this.props.cssModule);
     setScrollbarWidth(this.originalBodyPadding);
   }
--- a/src/__tests__/Modal.spec.js
+++ b/src/__tests__/Modal.spec.js
@@ -482,4 +482,47 @@ describe('Modal', () => {
     expect(document.getElementsByClassName('modal-dialog').length).toBe(0);
     expect(document.body.className).toBe('');
   });
+
+  it('should remove exactly modal-open class from body', () => {
+    // set a body class which includes modal-open
+    document.body.className = 'my-modal-opened';
+
+    const wrapper = mount(
+      <Modal isOpen={isOpen} toggle={toggle}>
+        Yo!
+      </Modal>
+    );
+
+    // assert that the modal is closed and the body class is what was set initially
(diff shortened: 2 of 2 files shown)


src/Modal.js
```diff
@@ -122,7 +122,8 @@ class Modal extends React.Component {
       this._element = null;
     }
 
-    const classes = document.body.className.replace('modal-open', '');
+    // Use regex to prevent matching `modal-open` as part of a different class, e.g. `my-modal-opened`
+    const classes = document.body.className.replace(/(^| )modal-open( |$)/, '');
```
TheSharpieOne: Best to use word boundaries `\b` because this will turn `my-modal-opened modal-open modal-opened` into `my-modal-openedmodal-opened` which is not wanted. This is because it removes the space from both sides.
`/\bmodal-open\b/` would result in `my-modal-opened  modal-opened`. There is double space, but that isn't going to break anything (technically, the current implementation would result in double spaces as well).

markogresak: Great catch!
This was adapted from [Element.classList polyfill](https://developer.mozilla.org/en/docs/Web/API/Element/classList#Polyfill).
I guess I could also replace with `' '` instead of `''`. 
If you would opt for using `\b`, I would at least follow the `.replace(/\bmodal-open\b/, '')` with `.replace(/\s{2,}/g, ' ')`, i.e. replace all occurrences of 2 or more spaces with a single space. Otherwise, it looks weird and could lead to a broken test case if someone would expect single space and would get 2 spaces.

I would also suggest using `document.body.classList` in the future. Bootstrap 4 supports only IE10+ and classList is already supported in IE10. With `classList.remove`, there would be no such problem in the first place 😄

markogresak:
I pushed a change to fix the problem caught by @TheSharpieOne. I had also left a comment, but GitHub decided to hide it after the changes were pushed.

Instead of using word boundary match, I opted in for replacing current regex's match with a single space. If the match is at start/end, the next line uses `classes.trim()` so any extra spaces are going to be removed. 
I had also suggested using [`Element.classList`](https://developer.mozilla.org/en/docs/Web/API/Element/classList), but that would require a larger change and will not be necessary after this is merged.

eddywashere:
thanks @markogresak!

PR #413, opened 2017-05-08
feat(DropdownMenu):Add tag

oktapodia:
Add custom tag property to DropdownMenu component, div by default (for bootstrap 3 **only**)


TheSharpieOne:
I don't think travis picked up on the destination branch change from master to v3. You may want to "kick" it but closing and reopening the PR. This should get travis to run the CI against the right destination.

PR #414, opened 2017-05-08
feat(DropdownMenu):Add tag

oktapodia:
Add custom tag property to DropdownMenu component, div by default (for bootstrap 3 **only**)


oktapodia:
Following https://github.com/reactstrap/reactstrap/pull/413

oktapodia:
Travis does not trigger it as well, I'll try to do with another naming branch

PR #418, opened 2017-05-08
chore(lint): fix linter error

TheSharpieOne:
A new linter rule lines up with new browser spec prohibits a trailing comma after the rest syntax.
Since there can be no variables defined after the rest, it is pointless to have a comma dangle there.


eddywashere:
good to know!

PR #425, opened 2017-05-14
fix(ButtonDropdown) group={false} did nothing

tankorsmash:
The underlying Object.assign needs to have the overrides to the right it seems like

```function example(){
  let props = {group: "FROM PROPERTIES" };
  let propsOverwritten = Object.assign({}, props, {'group': "group from defaults"});
  console.log("props first, groups overwritten", propsOverwritten);
  
  let notOverwritten = Object.assign({}, {'group': "group from defaults"}, props);
  console.log("groups first, overwritten by props as intended", notOverwritten);
}


example()
```

ES6 example https://es6console.com/j2paoelc/

Not crazy sure, but I changed the order in `reactstrap.es.js` of `g,{group:!0}` and it worked, so this seems like it would have the appropriate affect.

### DIFF
--- a/src/ButtonDropdown.js
+++ b/src/ButtonDropdown.js
@@ -8,7 +8,7 @@ const propTypes = {
 
 const ButtonDropdown = (props) => {
   return (
-    <Dropdown group  {...props}/>
+    <Dropdown group {...props}/>
   );
 };
 
--- a/src/ButtonDropdown.js
+++ b/src/ButtonDropdown.js
@@ -8,7 +8,7 @@ const propTypes = {
 
 const ButtonDropdown = (props) => {
   return (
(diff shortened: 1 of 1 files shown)


tankorsmash:
Thanks for catching the space error, I missed that one too!

eddywashere:
@tankorsmash if you're looking to conditionally change the ButtonDropdown to a normal dropdown, I would recommend importing Dropdown and doing the check there. Also, this component could likely go away in a future release to simplify component list.

tankorsmash:
@eddywashere  No worries, I was purely trying to get rid of the group class because I have the button alone in a div. If you need to prune the component that's totally fine on my end, thanks!

Issue #432, opened 2017-05-15
React's "minified copy of the development build" warning after upgrading to 4.6.1


jameschao:
### Issue description

After upgrading to 4.6.1 from 4.5.0, I'm seeing the following warning message when running my tests in Jest:

```
    console.error node_modules/reactstrap/dist/reactstrap.min.js:1
      Warning: It looks like you\'re using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See https://fb.me/react-minification for more details.
```
I get this simply by importing `Button` (or other components) from `reactstrap`, even without actually using it in my JSX code.

Related changes I made on my end is to use `react-transition-group` and no longer `react-addons-css-transition-group` and `react-addons-transition-group`.

Was there a built related change that might've caused this?

- version `#4.6.1`
- components: `Button`

### Steps to reproduce issue
- Upgrade to 4.6.1.
- import a component like `Button`.

th3fallen:
same here

Skaronator:
same as #429

jameschao:
This duplicates #430 so I'll close this issue.

PR #433, opened 2017-05-15
chore(dist): fix release

### LINKED ISSUE #430: [4.6.1] Pointing to minified react causes warning
### Issue description

Upon upgrading from 4.5.0 to 4.6.1, a new warning pops up in console:

> reactstrap.min.js:1 Warning: It looks like you\'re using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See https://fb.me/react-minification for more details.

I was not able to fork the suggested pen to reproduce the issue, likely due to #429.

TheSharpieOne:
fixes #430 and #432. This uses the generated lib files when requiring/importing from npm module (not not using treeshaking (webpack1)). **This does NOT fix the dist/ files (such as the UMD build) being generated incorrectly.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "reactstrap",
   "version": "4.6.1",
   "description": "React Bootstrap 4 components",
-  "main": "dist/reactstrap.min.js",
+  "main": "lib/index.js",
   "jsnext:main": "dist/reactstrap.es.js",
   "module": "dist/reactstrap.es.js",
   "scripts": {


th3fallen:
👍

will-stone:
Hi, when will this fix be merged/released?

th3fallen:
@eddywashere is this going to be tagged as a release?

eddywashere:
@will-stone @th3fallen 4.6.2

thanks again @TheSharpieOne!

th3fallen:
@eddywashere thanks for that, have you received this error since the release?    Cannot find module 'react-transition-group/TransitionGroup' from 'Modal.js'

eddywashere:
@th3fallen not seeing that output in tests. the peer deps recently changed. Can you make sure `react-transition-group` is installed? New install instructions are: `npm install --save reactstrap react-transition-group react react-dom`. feel free to update changelog with this upgrade info if it helps.

th3fallen:
yeah, that was it manually installing fixed, it i was literally on the way here to post that but you beat me to it.

balloob:
This fix is incorrect. By pointing it at the lib, you're now having people include a 300kB version of Reactstrap instead of 91kB. Each individual lib file contains a copy of the React helpers, resulting in their size being ~3x as big.

#474 is the right fix. I will include a revert of this PR in that fix.

TheSharpieOne:
@balloob that is incorrect, as reference please see every other library ever made. For those using non es modules, this is the way it is done. There is another property in the package.json for es modules.

Issue #306, opened 2017-01-26
Run modal's OnExit function after the exit animation


asdine:
Using the Modal component with React-Router is difficult because when changing the route to a route that doesn't contain the Modal, the exit animation doesn't have the time to finish before the Modal is removed from the dom.
The only solution i found until now is to close the modal manually and run `setTimeout(hashHistory.push('/whatever'), 500)`, to let the modal quit properly before being removed from the dom, but it's a pretty ugly solution.

It would be better if we could run the `OnExit` function (or a new one) at the end of the exit animation so we could have more control on what's happening.

eddywashere:
+1. A similar pr was added for the collapse component - https://github.com/reactstrap/reactstrap/pull/301. I think the method names should be the same.

PR #434, opened 2017-05-15
add onOpened and onClosed callbacks for modals

SteveByerly:
Addresses #306 

Also re-ordered the propTypes in the example to match the order of the source code.


SteveByerly:
@eddywashere @virgofx would you be able to review / comment on this?

Issue #429, opened 2017-05-15
[4.6.1] Reactstrap dist minified builds broken


virgofx:
The minified versions are broken in 4.6.1. Somehow the build does not produce proper UMD output with latest Rollup.  See output:

```
Uncaught ReferenceError: process is not defined
    at invariant.js:26
    at index.js:330
    at index.js:330
```

**invariant.js:26**
```
if (process.env.NODE_ENV !== 'production') {
  validateFormat = function validateFormat(format) {
    if (format === undefined) {
      throw new Error('invariant requires an error message argument');
    }
  };
}

eddywashere:
Thanks for reporting! Sounds like this affects people that point to a cdn hosted version or point to minified file in build tooling.

Was this working in the previous release or has this been broken since rollup integration?

I'll leave this up for grabs, if you work on it please call it here or slack. Otherwise I'll look at it this evening. Thanks!

virgofx:
Was working as of 4.5.x.

eddywashere:
Thanks! It's likely part of the roll up plugin changes I did. https://github.com/reactstrap/reactstrap/commit/8b379a41dd06d840c35779493a970124f953a0d7

Should be safe to revert or look at what broke between major versions.

TheSharpieOne:
https://github.com/reactstrap/reactstrap/blob/master/package.json#L5 should not be pointing to the dist, but the lib.

virgofx:
@TheSharpieOne what about the other values that point to dist as well ... `module` `jsnext:main` ?

th3fallen:
Should be fixed in https://github.com/reactstrap/reactstrap/pull/433

TheSharpieOne:
#433 will not fix the dist builds being development builds (for things like UMD, as mentioned here). It only fixes when importing from node_modules.

TheSharpieOne:
@vigrofx for es modules, the dist works. Extra components will be removed via tree shaking

eddywashere:
should be good now with 4.6.2. Upgrade notes: make sure all peer deps are installed with `npm install --save reactstrap react-transition-group react react-dom`.

PR #435, opened 2017-05-16
Replace process.env in builds (#429)

nathancahill:
The Rollup recommended way to handle this with the [replace plugin](https://github.com/rollup/rollup-plugin-replace).


eddywashere:
thanks @nathancahill!

Issue #400, opened 2017-04-22
Docs: It should close all nested modals if the main modal get closed


Skaronator:
This issue has more to do with the docs itself and not with the reactstrap code:

Version: 4.5.0

Reproduce:
1. Open https://reactstrap.github.io/components/modals/
2. Scroll down to 'Nested Modals' and launch the example
3. Open the nested modal
4. Click on "All Done"
5. Launch Example again and it'll open the nested modal as well.

I assume it would be better when the nested modal won't reopen again?

TheSharpieOne:
Right, it should dismiss/close the nested modal before dismissing/closing the outer one. The animation would look better doing that as well.

PR #439, opened 2017-05-18
Docs: Add method to close/dismiss all modals

miguelc1221:
(Docs) Addresses: #400 

I went ahead an added a method to close all the models properly when `All Done` is clicked.

### DIFF
--- a/docs/lib/examples/ModalNested.js
+++ b/docs/lib/examples/ModalNested.js
@@ -13,6 +13,7 @@ class ModalExample extends React.Component {
 
     this.toggle = this.toggle.bind(this);
     this.toggleNested = this.toggleNested.bind(this);
+    this.closeAllModals = this.closeAllModals.bind(this);
   }
 
   toggle() {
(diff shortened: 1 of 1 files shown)


docs/lib/examples/ModalNested.js
```diff
@@ -27,6 +28,12 @@ class ModalExample extends React.Component {
     });
   }
 
+  closeAllModals() {
+    this.setState({ nestedModal: !this.state.nestedModal }, () => {
```
TheSharpieOne: It would be best to just set these explicitly to `false`.
Also, the `setState` callback does not ensure the closing animation has completed or anything like that.

TheSharpieOne:
#434 added `onClose` callback to modals. This can be used to have one modal close after the other.

TheSharpieOne:
I'm going to close this as it is inactive and the request changes have not been made. Feel free to add more commits and reopen when it is ready.

PR #456, opened 2017-06-07
Fix(Dropdown, TetherContent): add touchstart event to TetherContent, …

### LINKED ISSUE #458: Popovers, tooltips, Dropdown doesn't close on iOS devices.
- version `#4.6.2`
- components: Popover, Tooltip, Drodown

How to reproduce:
- Open components with iPhone or iPad (iOS 10.3.1) in demo -> popover
- Open some popover
- Try to close it

Here my pull request: https://github.com/reactstrap/reactstrap/pull/456
I just added touchstart event in TetherContent and Dropdown components. Please check

Arsolio:
Just added touchstart event to fix issue with toggle for iOS devices. 

Closes #458


eddywashere:
Is it ever possible to trigger both events or is it always one or the other? I've seen reference to using `cursor: pointer` as a hack around this issue. That could be a work around for others until I have time to test this.

ozhe:
@eddywashere Hi! I also looked into this issue for sometime and here are some notes you might find useful:

- iOS browser(Safari, WKWebView) will only trigger the `touchstart` event.
- Desktop browsers will only trigger the `click` event.
- For Android browsers will trigger `click` if `touchstart` is not listened. If both are listened, only the `touchstart` will be triggered so there is no double triggering issue.

I haven't tested old Android versions but the statement above is true on 6.x and 7.x.

In conclusion, I think this PR is legit and I will be happy to see it get merged. (Currently I'm using the `cursor: pointer` hack)

PR #462, opened 2017-06-10
fix(DropdownToggle): Fixes #461 (reverting bad PR#402) which ensures

### LINKED ISSUE #461: DropdownToggle has invalid/extra color attribute
The DropdownToggle component leaks the `color` attribute with default value of "secondary" which isn't used by bootstrap for stylizing the DropdownToggle html native element. It appears it was a copy/paste issue from other components that has never been used. This results in invalid HTML5 warnings when validating source.

```
<div data-toggle="dropdown" aria-haspopup="true" color="secondary"
```

- version 4.6.2
- components: `DropdownToggle`

virgofx:
color attribute is NOT leaked. The color attribute was added into props because it originally existed as a defaultProp (but not in the prop types) as result from a copy/paste error. The correct solution is to remove from both as it's not used in stylizing the DropdownToggle in any fashion which results in the attribute leaking as an html prop which causes HTML5 validation to error.

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -6,7 +6,6 @@ import Button from './Button';
 
 const propTypes = {
   caret: PropTypes.bool,
-  color: PropTypes.string,
   children: PropTypes.node,
   className: PropTypes.string,
   cssModule: PropTypes.object,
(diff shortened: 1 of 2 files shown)


virgofx:
Updated PR based on feedback from @apaatsio. This should be good to go.

virgofx:
ping @eddywashere

PR #464, opened 2017-06-10
fix(Navbar): Fixes #463 by ensuring that the default role is now empty

### LINKED ISSUE #463: Navbar component should no longer use 'navigation' as default role
The HTLM5 `nav` element should not use a role of 'navigation' as this is the default and not allowed to set. Additionally, BS4 no longer includes this attribute in any example.

https://www.w3.org/TR/html5/sections.html#the-nav-element

virgofx:
fix(Navbar): Fixes #463 by ensuring that the default role is now empty for nav elements.

### DIFF
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -19,7 +19,6 @@ const propTypes = {
 
 const defaultProps = {
   tag: 'nav',
-  role: 'navigation',
   toggleable: false,
 };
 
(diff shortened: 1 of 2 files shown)


virgofx:
The default is now not set; however, we could adjust this based on the tag. For example if the tag is nav and the role is `navigation` then it's dropped. That could be the other alternative that way ... div elements get the role attribute. 
@eddywashere  Thoughts?

virgofx:
ping @eddywashere

PR #469, opened 2017-06-19
support the "active" property for dropdown items

ened:
As title, add the "active" class to the DropdownItem element.


eddywashere:
I don't see it in the docs exactly, but adding active class does trigger a background change 👍

PR #399, opened 2017-04-20
move to external react-transition-group dependency

ened:
Per https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html:

> react-addons-css-transition-group - Use react-transition-group/CSSTransitionGroup instead. Version 1.1.1 provides a drop-in replacement.

Please review the webpack.base.config.js changes whether it needs changes.

PR #474, opened 2017-06-22
Externalize ReactTransitionGroup (fixes big file size)

### LINKED ISSUE #472: File Size in 4.6.0 has jumped from 95kB to 335.6kB
### Issue description

Looks like the file size has tripled since release 4.6. I will spend some time later this week to see what caused it.

[4.5.0](https://unpkg.com/reactstrap@4.5.0/dist/) - 95.35kB:

image

[4.6.0](https://unpkg.com/reactstrap@4.6.0/dist/) - 335.6kB:

image

balloob:
Reactstrap bundle size jumped from 95kB to 335kB with the release of v4.6.0 (https://github.com/reactstrap/reactstrap/issues/472).

After observing the bundled output using [`source-map-explorer`](https://www.npmjs.com/package/source-map-explorer) I noticed that React, ReactDOM and React Transition Group all three were part of the bundle! [Visualized bundle](https://sourcemap-nrtnjyylaw.now.sh).

This was introduced when upgrading the React Transition Group package in #399. The imports were updated to point at modules that live under the `react-transition-group` package. For example:

```js
import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup';
```

By importing a module under the package, Rollup no longer found a matching entry in the list of imports that should be treated as external. This resulted in Rollup including it in the bundle.

This PR fixes it by setting up the correct external entries for the React Transition Group package and the imports that we use. I have also added back the entries to the global list for React Transition Group that were removed in #399 so that UMD builds will work correctly when used together with the UMD builds of React Transition Group. [Visualized bundle after the fix](https://sourcemap-qzasczgqli.now.sh/).

image

Fixes #472.

This PR also reverts #433. #433 worked around the wrong bundling by pointing at the `lib` folder. This is wrong because the lib folder contains individually transpiled files. By being individually transpiled, each file will contain a copy of the babel helpers that are used, resulting in each component being bigger than it should be.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "reactstrap",
   "version": "4.6.2",
   "description": "React Bootstrap 4 components",
-  "main": "dist/reactstrap.min.js",
+  "main": "lib/index.js",
   "jsnext:main": "dist/reactstrap.es.js",
   "module": "dist/reactstrap.es.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "reactstrap",
   "version": "4.6.2",
   "description": "React Bootstrap 4 components",
-  "main": "lib/index.js",
+  "main": "dist/reactstrap.min.js",
   "jsnext:main": "dist/reactstrap.es.js",
   "module": "dist/reactstrap.es.js",
   "scripts": {
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -19,16 +19,12 @@ const config = {
     'react',
     'react-dom',
     'react-transition-group',
-    'react-transition-group/CSSTransitionGroup',
-    'react-transition-group/TransitionGroup'
   ],
   // Used for the UMD bundles
   globals: {
     react: 'React',
     'react-dom': 'ReactDOM',
     'react-transition-group': 'ReactTransitionGroup',
-    'react-transition-group/CSSTransitionGroup': 'ReactTransitionGroup.CSSTransitionGroup',
-    'react-transition-group/TransitionGroup': 'ReactTransitionGroup.TransitionGroup',
   },
   targets: [
     { dest: 'dist/reactstrap.es.js', format: 'es' },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -19,11 +19,16 @@ const config = {
     'react',
     'react-dom',
     'react-transition-group',
+    'react-transition-group/CSSTransitionGroup',
+    'react-transition-group/TransitionGroup'
   ],
   // Used for the UMD bundles
   globals: {
     react: 'React',
     'react-dom': 'ReactDOM',
+    'react-transition-group': 'ReactTransitionGroup',
+    'react-transition-group/CSSTransitionGroup': 'ReactTransitionGroup.CSSTransitionGroup',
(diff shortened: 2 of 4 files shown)


nathancahill:
:+1:

nathancahill:
We should also add bundle size as a step in the build process, like this: https://github.com/developit/vhtml/blob/master/package.json#L13

virgofx:
@balloob If the references to `react-transition-group` imported via ES6 imports instead of cherrypicking would we need the cherrypicked references ? 

e.g. Change:  `import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup'` to `import { CSSTransitionGroup } from `react-transition-group`.

balloob:
@virgofx good point. I have addressed your comments 👍

balloob:
@nathancahill I will leave that up to a future PR. I also wonder if we should just have it output or actually make it a check on CI?

balloob:
@TheSharpieOne You've commented variations of your comment at three different spots. I will respond to all of them here.

You state that every lib ever is pointing at unbundled files. However, React will ship bundled by  Rollup as a single build for both development and production starting React 16: https://unpkg.com/react@16.0.0-alpha.13/index.js See this talk by one of the Facebook React maintainers at ReactEurope about how the size and load times both went down with 10%: https://youtu.be/djOc1EK07Tk?t=300

Since `lib/index.js` will import every component, there is no cherry picking going on. If you want to cherry-pick without using modules, you have to import from `lib/Card.js` etc, which is still very much possible. Importing `lib/index.js` however will contain everything as [per the source](https://unpkg.com/reactstrap@4.6.0/lib/index.js). This means that importing `lib/index.js` is equivalent to the compiled bundle, except that the Babel helpers will be part of each individually transpiled file. This averages a 500 byte cost per file, which with 70 components results in 39kB extra being shipped to users of Reactstrap. That's a 42% increase over the compiled file size.

The warning was because the 4.6 bundle was broken and actually included the whole of React in the compiled bundle! #433 didn't fix it, it just hid the problem for npm users while still having a broken build for UMD users. This PR will fix the build again and no warning will be issued.

TheSharpieOne:
When with react 16, they generate cjs and umd, right now and with your changes there is only umd with reactstrap. React 16 doesn't reference the umd output in their package.json. 
Maybe you need to create another file and point to that, just don't point to the umd output as that is most certainly wrong. 
Since size is the main concern (not just simply not including dependencies which should not be included) take a look at cjs vs umd file sizes in react 16, at least a 10% different in output size.

balloob:
That can be tweaked by marking lodash.omit and lodash.tonumber as external for a CJS build. [Would save us 10kB.](https://sourcemap-qzasczgqli.now.sh/)

Feel free to submit that as a followup PR once this one has been merged.

This PR saves 240kB for UMD bundles and 29-39kB* for CommonJS bundles. On top of that the loading time will be faster due to the flat bundle that Rollup creates.

*: 39kB for only including the Babel helpers once. Up to 10kB of waste if consumer also depends on lodash.omit.

balloob:
@TheSharpieOne I have reverted the change to main. No reason to delay fixing the ES/UMD bundles over a discussion about the CJS entrypoint.

balloob:
#477 includes this PR and creates a CommonJS build without any dependency included.

PR #477, opened 2017-06-26
Optimize CommonJS and module build

balloob:
This PR will split our Rollup config in two: one for the CJS/ES build and one for UMD. This is a backwards compatible change. It will also address the minimized name issue that people experienced in Storybook.

Bundled output (names link to the actual output):

 - `lib` - babelified source, stays the same
 - [`dist/reactstrap.cjs.js`](https://gist.github.com/balloob/79cb8c9d930dae9ab5a0c4cf8900e823#file-reactstrap-cjs-js) - unminified CommonJS build, pointed at from `package.main`
 - [`dist/reactstrap.es.js`](https://gist.github.com/balloob/79cb8c9d930dae9ab5a0c4cf8900e823#file-reactstrap-es-js) - unminified Module build, pointed at from `package.module`
 - [`dist/reactstrap.min.js`](https://gist.github.com/balloob/79cb8c9d930dae9ab5a0c4cf8900e823#file-reactstrap-min-js) - minified UMD build, to be included by users directly

image

The CJS/ES builds will not include any dependency and will not be minimized. These builds are going to be consumed by other bundlers and thus will be minimized when bundled for production (if the user chooses so). These bundles will be returned when `require('reactstrap')` or `import Reactstrap from 'reactstrap';`.

If you want to cherry pick only specific components:

With CommonJS: `const Card = require('reactstrap/lib/Card');`
With modules: `import { Card } from 'reactstrap';`

The UMD build is meant to be included directly on a webpage and will include the dependencies but not the peer dependencies. I've added a check so that when people add a peer dependency but not add it to the UMD globals object, it will fail the build.

This PR includes #474 as it was needed to generate correct builds.

CC @TheSharpieOne 

### Testing

To update source to run tests against the CommonJS build:

```
sed -i "" "s/\.\.\/\'/\.\.\/\.\.\/dist\/reactstrap.cjs'/" src/__tests__/*.spec.js
```

To update source to run tests against the module build:

```
sed -i "" "s/\.\.\/\'/\.\.\/\.\.\/dist\/reactstrap.es'/" src/__tests__/*.spec.js
```

To update source to run tests against the UMD build:

```
sed -i "" "s/\.\.\/\'/\.\.\/\.\.\/dist\/reactstrap.min'/" src/__tests__/*.spec.js
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,7 +7,7 @@
     "transform-object-rest-spread"
   ],
   "env": {
-    "dist-dir": {
+    "lib-dir": {
       "plugins": ["transform-es2015-modules-commonjs"]
     },
     "webpack": {
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "start": "BABEL_ENV=webpack webpack-dev-server --config ./webpack.dev.config.js --watch",
     "build-docs": "cross-env WEBPACK_BUILD=production webpack --config ./webpack.dev.config.js --progress --colors",
     "build": "rollup -c",
-    "prebuild": "BABEL_ENV=dist-dir babel src --out-dir lib --ignore src/__tests__/",
+    "prebuild": "BABEL_ENV=lib-dir babel src --out-dir lib --ignore src/__tests__/",
     "create-release": "npm test && sh ./scripts/release",
     "publish-release": "npm test && sh ./scripts/publish",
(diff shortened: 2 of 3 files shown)


eddywashere:
@balloob thanks! can you fix conflicts?

balloob:
@eddywashere done.

PR #478, opened 2017-06-27
Remove lodash.omit

balloob:
When I was analyzing [the source map](https://sourcemap-qzasczgqli.now.sh/) for #474 I noticed that the package `lodash.omit` takes up 9.29kB!

[Looking at the source](https://unpkg.com/lodash.omit@4.5.0/index.js), we can see that it pulls in a lot of dependent lodash functions.

I've added our own omit function to `utils.js`, which minifies to 105 bytes, and works with the use cases that we provide to it:

```js
export function omit(obj, omitKeys) {
  const result = {};
  Object.keys(obj).forEach(key => {
    if (omitKeys.indexOf(key) === -1) {
      result[key] = obj[key];
    }
  });
  return result;
}
```

This saves the user ~9kB.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -154,3 +154,16 @@ export function mapToCssModules(className, cssModule) {
   if (!cssModule) return className;
   return className.split(' ').map(c => cssModule[c] || c).join(' ');
 }
+
+/**
+ * Returns a new object with the key/value pairs from `obj` that are not in the array `omitKeys`.
+ */
+export function omit(obj, omitKeys) {
+  const result = {};
+  Object.keys(obj).forEach(key => {
+    if (omitKeys.indexOf(key) === -1) {
+      result[key] = obj[key];
+    }
+  });
+  return result;
+}
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "classnames": "^2.2.3",
     "lodash.isfunction": "^3.0.8",
     "lodash.isobject": "^3.0.2",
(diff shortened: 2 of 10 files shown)


src/utils.js
```diff
@@ -154,3 +154,16 @@ export function mapToCssModules(className, cssModule) {
   if (!cssModule) return className;
   return className.split(' ').map(c => cssModule[c] || c).join(' ');
 }
+
+/**
+ * Returns a new object with the key/value pairs from `obj` that are not in the array `omitKeys`.
+ */
+export function omit(obj, omitKeys) {
+  const result = {};
+  Object.keys(obj).forEach(key => {
```
mkalish: Is it worth having a null/undefined check before this line?

balloob: Right now we have the following 2 ways that omit is used:

```js
omit(this.props, ['some', 'keys', 'to', 'omit'])

omit(this.props, Object.keys(propTypes))
```

That means that both the object and the keys to omit are always passed in.

Right now if you pass in `null` for either parameter, it will blow up because it will execute `Object.keys(null)` or `null.indexOf(key)`.

Would you expect `obj` or `omitKeys` to be able to be `null` ?

mkalish: I was thinking of potential of `obj` to be null, but after a second I agree that it is unnecessary.

eddywashere:
I'd imagine most people already use a ton of lodash so this would be a minor increase in filesize. But for those not using all of it, this will be a big save. 👍

PR #483, opened 2017-06-28
fix(DropdownToggle, NavbarToggler, Tooltip): Pass through cssModule to child components

ajs139:
Trying to use CSS Modules proved problematic for a few of the components. These changes pass through the CSS Module to the child component when appropriate.

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -76,6 +76,7 @@ class DropdownToggle extends React.Component {
     } else if (!tag) {
       Tag = Button;
       props.color = color;
+      props.cssModule = cssModule;
     } else {
       Tag = tag;
     }
--- a/src/NavbarToggler.js
+++ b/src/NavbarToggler.js
@@ -18,8 +18,6 @@ const defaultProps = {
   type: 'button'
 };
 
-const navbarToggleIcon = <span className="navbar-toggler-icon" />;
-
 const NavbarToggler = (props) => {
   const {
     className,
@@ -40,7 +38,7 @@ const NavbarToggler = (props) => {
 
   return (
     <Tag {...attributes} className={classes}>
-      {children || navbarToggleIcon}
+      {children || <span className={mapToCssModules('navbar-toggler-icon', cssModule)} />}
     </Tag>
   );
 };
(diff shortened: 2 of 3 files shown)


src/ButtonGroup.js
```diff
@@ -35,7 +35,7 @@ const ButtonGroup = (props) => {
   ), cssModule);
 
   return (
-    <Tag {...attributes} className={classes} />
+    <Tag {...attributes} cssModule={cssModule} className={classes} />
```
TheSharpieOne: Here the default `Tag` is `div`, in which case the `cssModules` prop would not want to be passed down

TheSharpieOne:
I like what you are trying to do here, and you definitely exposed a pretty large issue with the way `tag` works with `cssModules` but I don't know if this is the best way to solve it.
We do not want cssModules being passed down when the `Tag` is *not* a reactstrap component, but we don't know what Tag is when it is a component. Several places you can pass one reactstrap component to another as a `tag` and it would make sense, and in those cases, the component passed as the tag would not receive `cssModule`.
The larger change would be to change the way css modules integrate and using something like a wrapping provider which exposes the modules in the context and have each of the components use the context to get the modules... but IDK if that is wanted as this project aims to be simple and not use context except for where absolutely necessary.

ajs139:
Thanks for the feedback, I see exactly what you mean. I've removed the changes to ButtonGroup, the other three changes seems less problematic, is that right?

One solution to allow the ButtonGroup to pass down the cssModule to its children could be this:

```
return (
    <Tag {...attributes} className={classes}>
      {React.Children.map(children, child => {
        let updatedChild = child;
        if (
          child.type.propTypes &&
          'cssModule' in child.type.propTypes &&
          !child.props.cssModule
        ) {
          updatedChild = React.cloneElement(child, { cssModule });
        }
        return updatedChild;
      })}
    </Tag>
);
```

However, I agree that this is probably too complex, and would need further discussions to be considered.

ajs139:
@TheSharpieOne - on reflection, I'm not sure changing ButtonGroup was the right plan anyway.

The idea was to be able to give ButtonGroup a CSS Module that would style its child components, but I think this will really be easy when https://github.com/webpack-contrib/css-loader/issues/520 is resolved.

Please let me know if you need anything else from me.

kuzvac:
Hi guys, thank you all for this fix. @TheSharpieOne when we can expect this fix in npm release? :)

PR #499, opened 2017-07-14
add justified attribute for ButtonGroup component

ened:
We should have a sweep one day to add all missing attributes. Perhaps with the BS beta.

### DIFF
--- a/src/ButtonGroup.js
+++ b/src/ButtonGroup.js
@@ -23,6 +23,7 @@ const ButtonGroup = (props) => {
     className,
     cssModule,
     size,
+    justified,
     vertical,
     tag: Tag,
     ...attributes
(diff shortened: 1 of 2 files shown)


ssteinerx:
Test is failing due to missing props validation:

`26:5  error  'justified' is missing in props validation  react/prop-types`

TheSharpieOne:
`.btn-group-justified` was removed for BS v4. The migration guide says so as well as it is not present in any of the docs nor code when searching. 
https://github.com/twbs/bootstrap/blob/6cf1a106d4fcd043f47d3eab4ffef3eaf868ce38/docs/4.0/migration.md

TheSharpieOne:
I am closing this as `.btn-group-justified` doesn't exist in bootstrap 4.

ened:
Ok. I shall look for a alternative solution then. We needed those justified. :)

PR #508, opened 2017-07-24
Fixed event name typo

jhsware:
However, there is already an event fired on `componentWillLeave` with the same name so perhaps one of these should be removed or renamed.


mkalish:
So I believe this is the correct event that should be fired after the slide has left the view, there is a chance that I'm misinterpreting these docs though https://v4-alpha.getbootstrap.com/components/carousel/#events

jhsware:
You are absolutely right! I'll close this

Issue #484, opened 2017-06-29
How to set options globally?


ahalimkara:
### Issue description

- version `#4.8.0`
- components: `tooltip`

I would like to set delay.hide property globally for tooltips to 0 instead of setting in every tooltip component. Is it possible?
`delay: PropTypes.oneOfType([
    PropTypes.shape({ show: PropTypes.number, hide: PropTypes.number }),
    PropTypes.number
  ])`

PS:
Since early versions of Bootstrap, the default delay is always 0. So i don't understand why it is 250ms for reactstrap. I feel like the page works slow when it has delay on hide.

TheSharpieOne:
Perhaps the default should be `0` to match what bootstrap's is. Other than that, you can create a wrapper component which will default values for you.
```
import {Tooltip} from 'reactstrap';
export default function MyTooltip(props) {
  return <Tooltip delay={0} {...props} />
}
```

TheSharpieOne:
@ahalimkara did this answer your question?

ahalimkara:
@TheSharpieOne thank you for your answer, yea it answers my question however i thought maybe core developers will consider to make default values same with bootstrap's defaults. Thanks again and sorry for late reply.

TheSharpieOne:
I believe the default delay was to match the animation duration since the animation is true by default for bootstrap. Since there was no way to control the animation in reactstrap (which is another thing that doesn't line up with bootstrap) it was probably a compromise.
I made a PR to default hide to 0 and add animation and a way to control the animation. This should satisfy this issue

TheSharpieOne:
I'm closing this since the the issue is no longer active and it appears it was answered above. Please comment and it will be opened if you feel the issue has not been resolved / the question has not been answered.

PR #514, opened 2017-07-28
fix(Tooltip): match BS default values

TheSharpieOne:
Default tooltip hide delay to 0 to match bootstrap
Add animation prop defaults to true to control animation
Add animationDuration prop defaults to 250 to allow animation duration to be overriden
Adjust TetherContent to allow classNames to change and rerender

Closes: #484

Breaks Tooltip: Tooltip's default hide delay is now 0 (previously 250) and now has fade animation with a duration of 250. This was done to better match bootstrap's default. If you like the previous defaults, you can set `delay={hide:0}` and `animation={false}`

Demo: (left: current; right: proposed)
tooltip-animation

### DIFF
--- a/docs/lib/Components/TooltipsPage.js
+++ b/docs/lib/Components/TooltipsPage.js
@@ -43,11 +43,13 @@ export default class TooltipsPage extends React.Component {
   // optionally overide tether config http://tether.io/#options
   tetherRef: PropType.function,
   // function which is passed a reference to the instance of tether for manually \`position()\`ing
+  animation: PropTypes.bool, // defaults true; controls fade animation.
+  animationDuration: PropTypes.number, // default 150; controls how long the animation takes before removing the tooltip
   delay: PropTypes.oneOfType([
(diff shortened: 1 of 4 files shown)


ahalimkara:
@TheSharpieOne that's a great PR, just a small question, is there any particular reason to set animation duration to 250ms as bootstrap's default is[ 150ms](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L226)?

TheSharpieOne:
Their docs didn't mention the duration and I didn't dig. I guess is assumed the 250 hide delay was supposed to be the hide animation duration.

Geczy:
Is this going to be merged?

TheSharpieOne:
It is waiting on review, but also these changes will conflict with changes coming from v5 for bootstrap v4 beta support where tether was swapped out for popper.js

TheSharpieOne:
Closing this as it will cause issues with v5 which will be the next release.

PR #538, opened 2017-08-13
docs(Badge) update and fix docs

mtakac:
Remove default badge and update documentation so it's closer to bootstrap v4 beta
docs

### DIFF
--- a/docs/lib/Components/BadgePage.js
+++ b/docs/lib/Components/BadgePage.js
@@ -6,6 +6,9 @@ import Helmet from 'react-helmet';
 import BadgeExample from '../examples/Badge';
 const BadgeExampleSource = require('!!raw!../examples/Badge');
 
+import BadgeButtonExample from '../examples/BadgeButton';
+const BadgeButtonExampleSource = require('!!raw!../examples/BadgeButton');
+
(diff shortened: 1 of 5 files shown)


mtakac:
After Bootstrap 4 beta release default badge is not working anymore. Bootstrap documentation now uses secondary badges in the first example. I've done the same in Reactstrap docs. I also added few more examples from original bootstrap documentation.

One thing I am not sure about are links in badges. See original docs [here](https://getbootstrap.com/docs/4.0/components/badge/#links). Should we create a whole new component or maybe add new prop to badge component to allow badge links or should they be created the same way as in Bootstrap: by adding a `className` to link? Also do we even want to include them in Reactstrap documentation?

TheSharpieOne:
Thanks for this. I'm going handle the link the same as how Button works, if there is an `href` and the `tag` is defaulted, it will become an anchor tag.
Also, as for the default... I am going to just default it to `secondary` for now. I know you had mentioned in #539 that it should not default to anything and be required, but I am split about it and am erring on the side of not making a breaking change. If anything I would probably keep defaulting it to 'secondary' and add a deprecation message.

PR #557, opened 2017-08-29
Update react transition group (WIP)

amikhalev:


src/Fade.js
```diff
@@ -1,128 +1,62 @@
...
   baseClass: 'fade',
-  baseClassIn: 'show',
-  transitionAppearTimeout: 0,
-  transitionEnterTimeout: 0,
-  transitionLeaveTimeout: 0,
-  transitionAppear: true,
-  transitionEnter: true,
-  transitionLeave: true
+  baseClassActive: 'show',
+  timeout: 150,
```
virgofx: It appears bootstrap default is 300ms

https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L688

We should definitely comment this (briefly here and more thoroughly in the example/documentation) as the timeout here also requires updating `$modal-transition` in custom bootstrap builds as this is important for syncing the CSS UI with the timeouts here.


EDIT: I apologize. I referenced the Modal timeouts against the Fade here.  Bootstrap's Fade components are all 150 so this is correct. The modal is separate which uses 300ms/150ms. Could we at least add documentation in both these locations in source and in documentation examples as its real important that these sync otherwise there will be huge jumps where the transition end timeouts prior or the CSS transition is still continuing.

* Modal timeout changes need to update: `$modal-transition`
* Fade component changes (e.g. Alert) need to update: `$transition-fade`

amikhalev:
@TheSharpieOne

I think this is almost done. Only Carousel doesn't work ATM. I do have a few questions

1. I kinda forgot to do conventional commits, so would you like me to go back and edit each commit message? Or I could squash this into 1 or several commits with better messages.

2. Should I add a doc page for the Fade component? Or is this considered an internal component 

3. Any other changes that need to be made to this code?

TheSharpieOne:
@amikhalev Thanks a ton for tackling this.
1. I can (and will) squash when I merge the commit and change the commit message to use a conventional message
2. I would say yes, create a doc page, but if you don't, it probably isn't that big of a deal (like tether is)
3.
    - The transition prop names exposed for `Alert` vs `Modal`  and `CarouselItem`. It would be nice to have them all be consistent.
    - Make all of the transition props overridable (there are a few which are not)
    - Allow timeout to an object or an integer much like Transition allows (kinda goes with consitent prop names b/c Alert has two timeout props)
    - Speaking of the `Fade` component, would it be possible to use `Fade` with `Alert` rather than having it's own within the component?
    - When checking the status of the transition, can you use the variables exposed by `Transition `?`import Transition, { EXITED, ENTERED, ENTERING, EXITING }  from 'react-transition-group/Transition';`
    - What is going on with Carousel? (do you plan on resolving it in thie PR or should this be merged and Carousel be addressed in a later effort?)
    - I'll add more comment inline

src/CarouselItem.js
```diff
@@ -1,84 +1,62 @@
...
 
-    const itemClasses = mapToCssModules(classNames('carousel-item', ...this.state.animation), cssModule);
 
     return (
-      <div className={itemClasses} ref={(slide) => { this.slide = slide; }}>
-        <img className={classes} src={src} alt={altText} />
-        {children}
-      </div>
+      <Transition
+        in={true}
```
TheSharpieOne: Do we want expose Transition props and spread them here? If so, can you also check for the existance of a user provided `onExited` and `onExiting` (via props) and call them in your callbacks (to not clobber the user's callbacks)

virgofx:
Best practice is to NOT cherry pick. We don't do that anywhere. Why was that changed?

https://github.com/reactstrap/reactstrap/pull/557/files#diff-4015e3ba4b5aea6acd9696987286fcfdR4

Use proper ES imports.

src/CarouselItem.js
```diff
@@ -1,93 +1,73 @@
...
-      </div>
+      <Transition
+        in={isIn}
+        onExiting={this.componentWillExit}
+        onExited={this.componentDidExit}
+        timeout={500}
+      >
+        {(status) => {
+          const { direction } = this.context;
+          console.log("altText: ", altText, ", status: ", status, ", direction: ", direction);
```
mkalish: Extra log statement here

mkalish:
@amikhalev 
I'm happy to take a look at updating the carousel code over the weekend if you want to get this merged

amikhalev:
@virgofx Ok, I wasn't aware that individual imports was preferred

amikhalev:
@mkalish @TheSharpieOne I should be able to fix Carousel (and the rest of your issues), I just made some temp commits for now. I'll probably get it done some time this weekend

virgofx:
@amikhalev Thanks. :)

TheSharpieOne:
@amikhalev I don't want to mess with your flow, but I have some large commits ready to be merged into v5 (such as the popper branch). I had to update react-scripts which updated eslint and jest which changed the test API (the jest 19/20 stuff) and eslint rules. Everything is working in the popper branch and test coverage have been improved (back to 99.66%), but I had to touch lots of test files, including ones you have touched. I also had to fix new linter issues which touched some of the code you have touched.
I'm not sure the best course of action, if you want me to merge my stuff in and you rebase / pull and deal with what conflicts may happen (since you will be more familiar with those files as the major changes are in your branch), update tests which may need to use the new jest stuff, as well as fix linter issues which may arise. Or I can wait to merge my code until you are done and I can deal with the conflicts (most likely keeping your changing and just fixing anything that breaks).

amikhalev:
@TheSharpieOne Go ahead and merge the updates, I can deal with any merge conflicts

amikhalev:
@virgofx It doesn't look like there's a way to access the constants defined [here](https://github.com/reactjs/react-transition-group/blob/master/src/Transition.js#L7) without importing `"react-transition-group/Transition"`. Would you say it is OK to do so in this case?

amikhalev:
@TheSharpieOne

I think I fixed most of your issues. `Carousel` works now (:tada:), `Alert`, `Carousel` and `Modal` all expose their transition props, `Alert` uses `Fade`, and everything uses constants from `react-transition-group`

Here are the issues that I think I still need to address

* [ ] There is some kind of issue with `Modal` which prevents it from dismissing properly
* [ ] If you click the control button or keyboard button too quickly with `Carousel`, the transition breaks and it shows multiple slides stacked vertically
* [ ] Lots of documentation is out of sync (mostly for `Carousel`)
* [ ] Lots of tests are broken (might wait to fix this one until new testing stuff is merged)
* [ ] Probably should write more/better tests for `Fade`, `Carousel` (esp. slide prop)
* [ ] I think that since `Fade` is used by both `Alert` and `Modal`, plus it could be used by users of reactstrap for fading arbitrary components, it should probably get it's own documentation page explaining it's props in more detail

amikhalev:
@TheSharpieOne It looks like `Collapse` could be made to use `Transition`. Should I pass it's existing props to `Transition` or just update them to match?

virgofx:
I was thinking the same thing. I believe we could just wrap around Transition and the Collapse component itself should be much smaller code-wise as we should re-use much of the transition logic/props, etc. So the closer we can keep inline with transition would be my vote.

amikhalev:
In that case, what should I name props?
`in` or `isOpen`?
`onOpened` or `onEntered`?
`onClosed` or `onExited`?
Should they be consistent between `Alert`, `Fade`, `CarouselItem`, `Collapse` and `Modal`? Or maybe more than one should be supported for each component. I'm not sure what's best here

TheSharpieOne:
I think we should probably keep `isOpen`, even though `in` is more generic. In the few cases where `isOpen` doesn't make any sense, I'm fine with using `in` there (even if it isn't 100% consistent). For the callbacks, I would prefer the same names as the transition component.

holgersindbaek:
When is this branch getting merged? I'm getting a lot of issues with react-transition-group.

carlschubert:
I've also been watching this thread.  I've been holding off on a few things waiting for react-transition-group to be updated.  It would be nice to know the timeline.

amikhalev:
All of the code is done. The only thing that it needs is updated tests (a lot of which is just removing tests because the code was simplified a lot). I'll probably get it done later today

amikhalev:
Ok, looks like all tests pass now. I'm pretty sure all docs are updated too. I think this is ready for final review

src/Collapse.js
```diff
@@ -1,159 +1,131 @@
...
 
-  getDelay(key) {
-    const { delay } = this.props;
-    if (typeof delay === 'object') {
-      return isNaN(delay[key]) ? DEFAULT_DELAYS[key] : delay[key];
-    }
-    return delay;
+  onExiting(node) {
+    // getting this variable triggers a reflow
+    const _unused = node.offsetHeight; // eslint-disable-line no-unused-vars
```
TheSharpieOne: I wonder if the minifier will remove this or not

src/Collapse.js
```diff
@@ -1,159 +1,131 @@
...
-
-      this.transitionTag = setTimeout(() => {
-        this.setState({
-          collapse: HIDDEN,
-          height: null
-        });
-      }, this.getDelay('hide'));
-    }
-    // else: do nothing.
+    ['onEntering', 'onEntered', 'onExit', 'onExiting', 'onExited'].forEach((name) => {
```
TheSharpieOne: Interesting. Maybe `CarouselItem` could do the same.

TheSharpieOne:
Awesome job @amikhalev. Thanks a ton!

amikhalev:
Thanks! It was my pleasure. reactstrap is a great library, glad to contribute!

holgersindbaek:
Nice. Thanks.

PR #560, opened 2017-08-29
Use cross-env for all npm scripts

amikhalev:
Allows all NPM scripts to be run cross-platform

Not sure if this should be merge to `master` or `v5`, or both?


TheSharpieOne:
I'll pull it into v5 when I get a chance

PR #567, opened 2017-09-09
feat(Navbar): Added dark prop

Lobosque:
If there is a `light` prop to add the `navbar-light` class, then to keep consistency there should also be a `dark` prop to add the `navbar-dark` class.
This PR adds the `dark` prop which will add the aforementioned class when set. With a test and the prop entry in the comments.


TheSharpieOne:
This is already addressed in the v5 branch which accounts for changes in bootstrap 4 beta

PR #571, opened 2017-09-12
fix(popperjs): fixed error in build

theJoeBiz:
Fixed build error: 'placements' is not exported by node_modules/popper.js/dist/esm/popper.js
Renamed `Popper` to `PopperContent` in `Tooltip.js` to match `Popover.js`.

### DIFF
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -1,12 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { placements } from 'popper.js';
+import PopperJS from 'popper.js';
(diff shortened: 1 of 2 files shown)


TheSharpieOne:
Good catch. Can you also update [Popover](https://github.com/reactstrap/reactstrap/blob/v5/src/Popover.js#L4) to be the same?

theJoeBiz:
@TheSharpieOne I fixed both tooltips and popovers in the original commit. I'm running the v5 branch locally now and popovers and tooltips are working great. This should be ready to merge.

TheSharpieOne:
Sorry, I missed that ☹️

PR #490, opened 2017-07-03
[wip] new docs

eddywashere:
exploring gatsby based docs. [demo](https://reactstrap.surge.sh/)

TODO:

- add markdown version of component and utility pages
- fix redirects

PR #572, opened 2017-09-13
WIP: v4-beta badges

theJoeBiz:
I updated the badge docs to use the updated variations from the v4-beta docs.
```diff
- default
primary
+ secondary
success
danger
warning
info
+ light
+ dark
```

**TODO:**
`badge-default` class has been removed in v4-beta. `badge-secondary` seems to be the closest match, should reactstrap use that if no color prop is passed?

### DIFF
--- a/docs/lib/examples/Badge.js
+++ b/docs/lib/examples/Badge.js
@@ -5,12 +5,12 @@ export default class Example extends React.Component {
   render() {
     return (
       <div>
-        <h1>Heading <Badge>New</Badge></h1>
-        <h2>Heading <Badge>New</Badge></h2>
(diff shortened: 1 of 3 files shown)


TheSharpieOne:
There is some discussion around defaulting things such as the color. I am for reasonable/sensible defaults (such as `secondary`). (At the very least, it should not default to `default` anymore.)

TheSharpieOne:
Also, just FYI #538 (which is pointing to the wrong branch) and #490 which may make all of the doc changes pointless 😀

theJoeBiz:
@TheSharpieOne Damn! Alright, I'll close this out.

PR #574, opened 2017-09-16
V5

### LINKED ISSUE #113: Append to container options: modal, popover, tooltip & dropdown.
Currently only the dropdown has an option that appends the menu content to the body. It's enabled via the `tether` prop, and is used to auto position the dropdown (below or above button) when content hits the viewport. Not necessarily as an option to append to body but a way to make sure dropdown content is always visible.

It's probably worth letting people opt-out of the append to body behavior. I think it involves some digging into with tether options and how `TetherContent` children get rendered and positioned.

Todo:
- [ ] Document tether option in dropdown 😅 
- [ ] Get a proof of concept with tether option and rendering to body/container vs inline
- [ ] Standardize prop name, in [bootstrap.js](http://getbootstrap.com/javascript/#popovers-options) it's container

### LINKED ISSUE #337: Tooltip `target` requires string, but Tether takes object?

### LINKED ISSUE #424: Standardize `getTarget`
From https://github.com/reactstrap/reactstrap/pull/331#issuecomment-301344176

> It looks like we need to standardize and abstract the getTarget functionality into utils. Right now popover/tooltip have different ways of getting target. To future proof getTarget as much as possible it should support a css selector (string), a dom reference or a function that returns a dom reference. Once that's in it will be easier to check if target matches click target. going to close this out until that happens.

Components to update

- Tooltip
- Popover

### LINKED ISSUE #438: tether.js warning
I just installed "reactstrap": "^4.5.0", and got the following warning.  What do I need to do to resolve  it?
```
/~/reactstrap-tether/dist/js/tether.js
Critical dependencies:
2:479-486 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/reactstrap-tether/dist/js/tether.js 2:479-486
printWarnings @ webpackHotDevClient.js:198
```

### LINKED ISSUE #465: Popover toggle does not work with target?
So, I'm following the [Reactstrap docs example for Popovers](https://reactstrap.github.io/components/popovers/) and everything displays fine. However, the Popover component toggle does not work if I click on the target when it is open. The behaviour is such that it closes first because it thinks that the user is clicking away from the Popover, but then the `onClick` handler is fired immediately after on the target, so it reopens. What results is an action where the toggle does not work after the Popover is open.

Is there anyway to address this? You can test this out yourself for all examples in the docs I just linked.

### LINKED ISSUE #466: Cannot bundle 'reactstrap' by 'browserify' because of 'reactstrap-tether'
Thanks for providing this awesome package which made my development much easier.

I got an issue related to 'reactstrap-tether'. I've also added a comment in https://github.com/HubSpot/tether/pull/175

It is caused by the tether.js forked in 'reactstrap-tether'.
Basically this tether.js will block browserify from bundling 'reactstrap', there will be an error:
```
Error: Cannot find module './utils' from .....
```

The workaround is ignore 'reactstrap-tether' from the bundling.
```javascript
bundle.ignore('reactstrap-tether');
//or command line
browserify -i reactstrap-tether ....
```
But the result will not have ToolTip or other components which used 'tether'.
It seems the webpack or 'create-react-app' could work with current 'reactstrap' setup without any issue.
I think it would be nice to let 'browserify' finish its job properly.

BTW, directly include './node_modules/reactstrap-tether/src/js/tether' won't work.
The 'import' and 'export' keyword will cause trouble. User must provide preset such as es2015.
Is it better to avoid using 'import/export' and keep the minimal dependency?

#438

### LINKED ISSUE #485: static prop on Input component
While this is related to a TypeScript definition file, I wanted to submit the issue here since a change may need to be made in this repo before the definition is updated (+1 for [co-locating definitions](https://github.com/reactstrap/reactstrap/issues/276) in this repo).

As the [comment](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/reactstrap/lib/Input.d.ts#L41-L42) in the `Input` definition file states, `static` is a keyword in both JS and TS and therefore cannot be used as a property in the `InputProps` interface. As a result, folks need to use the `form-control-static` class in their components. That's obviously not the end of the world, but I think it's safe to assume using a prop is preferable.

I wanted to see if there were any ideas about whether there'd be a way to support this by using a non-reserved word as a prop.

/cc  @alihammad @mfal @danilobjr

### LINKED ISSUE #487: Tracking Beta 1 Changes
read through https://github.com/twbs/bootstrap/issues/21568

- **Grid/Layout**
  - [x] Column offsets are gone ([but will be back again in bootstrap v4 beta 2](https://github.com/twbs/bootstrap/pull/23445))
- **Card**
  - [x] Renamed `.card-block` to `.card-body` https://github.com/twbs/bootstrap/pull/22828
  - [x] `inverse` is no longer a thing, just use `.text-white` (keep for convenience?)
  - [x] Outline (`.card-outeline-{color}`) is now border util `.border-{color}`
  - [x] Color (`.card-{color}`) is now bg util `.bg-{color}`
- **Nav/NavBar**
  - [x] nav now has `.nav-fill`
  - [x] nav `vertical` should accept a string for the responsive size (`.flex-{size}-column`)
  - [x] nav now has `.card-header-tabs`/`.card-header-pills` (add boolean `card` prop?)
  - [x] `.navbar-inverse` -> `.navbar-dark`
  - [x] `.navbar-toggler-left` and `.navbar-toggler-right` have been removed
  - [x] `Navbar` [toggle class](https://github.com/reactstrap/reactstrap/blob/358576442e99b99fc88db1157531bacef43994c3/src/Navbar.js#L25-L33) `.navbar-toggleable{-sm|-md|-lg|-xl}` -> `.navbar-expand{-sm|-md|-lg|-xl}`
- **Popover**
  - [x] Renamed `.popover-title` to `.popover-header` and `.popover-content` to `.popover-body` twbs/bootstrap#22829
  - [x] tether -> poppers.js
- **Tooltip**
  - [x] tether -> poppers.js
- **Forms**
  - [x] form validations (feedback and "server-side" classes)
- **Dropdowns**
  - [x] `.show` is now needed on `.dropdown-menu` when open
  - [x] tether -> poppers.js
- Badge
  - [x] `.badge-default` -> `.badge-secondary`

bootstrap v4 beta docs: https://getbootstrap.com/docs/4.0/getting-started/introduction/

The [v5 branch](https://github.com/reactstrap/reactstrap/tree/v5) will have the latest changes while the list is being completed. Items marked as completed on the list indicate that the item is addressed in the v5 branch. This issue will be closed one v5 has been merged into master.

Feel free to add anything not listed with a general description and link to either docs page or pr and I'll update the list 🙌

### LINKED ISSUE #502: react-transition-group 2.x causes an unknown props error on Modal open
### Issue description

- version `#4.8.0`
- components: `modal`

### Steps to reproduce issue

- Install react-transition-group 2.1.0.
- Open a Modal
```
Warning: Unknown props `onExited`, `appear`, `enter`, `exit` on <div> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
    in div (created by Fade)
    in Fade
    in div (created by TransitionGroup)
    in TransitionGroup
```
Downgrading to 1.2.0 fixes the problem.

### LINKED ISSUE #529: reactstrap modal.js show function alway use 'style.position = "relative"'
### Issue description

The modal form is always out of the parent div tag. 

I want to put this modal popup in div tag classnamed 'wrapper01'.
This 'wrapper01' has css style 'position: relative' for parent view.
 
but the modal is always use 'position: relative', so I can't put this modal popup in my parent 'wrapper01'.

here is my html map.

I think 'show()' function in the modal.js has the static position style 'relative'. 

I want to put more elements like 'position: absolute' , 'top: 100px'. so I use this modal popup in my parent div tag.

Can it be fixed? 

- version: 4.8.0
- components: modal.js

### LINKED ISSUE #532: Modal AutoFocus broken with React 16 because of render unstable changes
Rendering a Modal component with defaults fails in React 16 with:

```
Uncaught TypeError: Cannot read property 'parentNode' of undefined
```

The root cause is because the Modal's autoFocus is attempting to reference the `this._dialog` which is set via callback and does not exist as result of changes made in `unstable_renderSubtreeIntoContainer`.

Reference: https://github.com/facebook/react/issues/10309

- Reactstrap: `4.8.0`
- React: `16.0.0-beta.3`
- Components: `Modal`

### Steps to reproduce issue
1. Specify a new `Modal` component with no `autoFocus` attribute or `autoFocus={true}`

### Demonstration
https://www.webpackbin.com/bins/-Kr2BOnFO9GcE691BLYr

To toggle the error adjust the `autoFocus` property (Setting to `false` removes error)
`<Modal isOpen={true} autoFocus={true}>`

### LINKED ISSUE #535: Button color defaultProp as "secondary" is a bad default in Bootstrap 4-beta
- version `#4.8.0`
- components: `Button`

So bootstrap 4 has finally moved from alpha to beta, and with it they have dramatically changed the default colour palette. Most notably they have changed the `btn-secondary` css class from a white color to a dark grey, which doesn't seem like a good default for a button. Unfortunately, "secondary" is the default prop for the reactstrap Button class. I recommend making the default for the Button class to be just the bootstrap `btn` css class or `btn-light` css class, because otherwise I imagine that it might become necessary to explicitly give a color prop to almost every instance of Button.

https://getbootstrap.com/docs/4.0/components/buttons/

### LINKED ISSUE #539: Color prop of Badge component should be required
### Issue description
The `Badge` component can be rendered as default without specifying any `color`. IMO `color` should be required. There isn't any default `Badge` in current version of Bootstrap and even in Reactstrap, the badge without color will not be visible. I fixed the docs in [this PR](https://github.com/reactstrap/reactstrap/pull/538), but I think we should still make the `color` required.

- Reactstrap version `#4.8.0`
- Bootstrap version `#4.0.0-beta`
- components: `Badge`

### Steps to reproduce issue
1. Render `Badge` without specifying `color` property:
```
<Badge>New</Badge>
```

Discussion about similar issue concerning `Button`: https://github.com/reactstrap/reactstrap/issues/535. Currently rendering `Button` without `color` prop will render secondary button. IMO this is inconsistent with Bootstrap but we should at least ensure that all components act in similar and predictable manner. Therefore if this is the behaviour we are after, `Badge` should work similarly as well and render secondary `Badge` if `color` is not specified.

### LINKED ISSUE #548: [R16] UncontrolledTooltip, Tooltip Error when Active in React 16
When hovering over a Tooltip or UncontrolledTooltip using the default sample code for these components from docs, an error will occur in React16 as follows: (Note this is specific to React 16 and not thrown in React 15):

```
reactstrap.es.js:715 Uncaught TypeError: Cannot read property 'focus' of undefined
    at TetherContent.show (reactstrap.es.js:715)
    at TetherContent.handleProps (reactstrap.es.js:681)
    at TetherContent.componentDidMount (reactstrap.es.js:632)
    at commitLifeCycles (react-dom.js:11925)
    at commitAllLifeCycles (react-dom.js:12707)
    at HTMLUnknownElement.callCallback (react-dom.js:1543)
    at Object.invokeGuardedCallbackDev (react-dom.js:1582)
    at invokeGuardedCallback (react-dom.js:1439)
    at commitAllWork (react-dom.js:12828)
    at workLoop (react-dom.js:13100)

The above error occurred in the <TetherContent> component:
    in TetherContent (created by Tooltip)
    in Tooltip (created by UncontrolledTooltip)
    in UncontrolledTooltip (created by <App Component>)
    in span ...
```

If the tooltip component is switching away from Tether to Popper then whoever does that PR if they'd be so kind to include testing with latest React16.x and fix this at same time.

- Reactstrap: `4.8.0`
- React: `16.0.0-beta.5`
- Components: `Tooltip`, `UncontrolledTooltip`

### LINKED ISSUE #551: v4-beta .show class for navbars conflict with Collapse component
reactstrap version `4.8.0`
### Issue description
I'm using the `Collapse` component for my `.navmenu`, in Bootstrap v4 alpha 6, the styling was good.
Now I've updated to v4-beta and this is the result:

capture

As you can see the color of the `.nav-link` inside the collapse are more white. This is because of the `.show` class that is being used in v4-beta. The `.show` class is also used by reactstrap for the `Collapse` component which now messes with the styling.

You can see the class in the source here: https://github.com/twbs/bootstrap/blob/35f80bb12e4e71fd777ee60ffa43711d8f84b1a6/scss/_navbar.scss#L264

### LINKED ISSUE #563: Popover with fixed target; placement incorrect after resize
### Issue description
The placement of a popover bound to a fixed target becomes incorrect after resizing the viewport/window.

- version `#4.8.0`
- components: `Popover`

### Steps to reproduce issue

- Create a container with a `Button` that has CSS property position fixed and a `Popover` with some dummy content. 
- Resize your window; the popover jumps to the left end of the screen

See this [CodeSandbox](https://codesandbox.io/s/zn4wpo31km)

### LINKED ISSUE #583: <input> reference for Input
### Issue description

I'm attempting to make a hidden file input where, when I click a button, the file input dialog shows up. The codepen link below illustrates the problem pretty well -- when you click the button nothing happens. However, when you change the `<Input>` to a basic `<input>` everything works just like you'd expect.

Is there a way to reference the basic `<input>`? Not a huge issue for me here since the input doesn't need to look fancy anyway, but could be an issue for other applications.

- version `#4.8.0`
- components: `Input`

### Steps to reproduce issue

Check out [this](https://codepen.io/anon/pen/xXOyoE) pen. Clicking the button does nothing at first, but once you change the `Input` to `input` everything works as expected.

TheSharpieOne:
Anyone and everyone please review and give feedback. Lots of changes and descriptive commit messages with deprecations and breaking changes information (which I hope get automatically added to the release notes).

Tests have been updated
Docs have been (mostly) updated.

closes #487, closes #529, closes #539, closes #113, closes #337, closes #424, closes #465, closes #438, closes #466, closes #502, closes #532, closes #535, closes #563, closes #548, closes #551, closes #485, closes #583
(Have to put [keyword in front of every issue number](https://help.github.com/articles/closing-issues-using-keywords/#closing-multiple-issues)... why can't it just be comma seprated)


th3fallen:
🚢  it :)

supergibbs:
Ship it but as an beta; you'll get much more testing when devs can easily try it in their apps.

src/Popover.js
```diff
@@ -1,79 +1,171 @@
...
-import TetherContent from './TetherContent';
-import { getTetherAttachments, mapToCssModules, omit, tetherAttachements } from './utils';
+import PopperJS from 'popper.js';
+import PopperContent from './PopperContent';
+import { getTarget, DOMElement, mapToCssModules, omit } from './utils';
 
 const propTypes = {
-  placement: PropTypes.oneOf(tetherAttachements),
-  target: PropTypes.string.isRequired,
+  placement: PropTypes.oneOf(PopperJS.placements),
```
cauethenorio: @TheSharpieOne looks like the `Popover` component is missing the `container` prop which is in the v5 docs.

SimeonC:
For those who want to test I have a branch with the lib files commited in here: https://github.com/SimeonC/reactstrap/tree/v5.

You can install via `npm install simeonc/reactstrap#v5`. I'll try keep it updated alongside this branch until they release something more official!

src/Popover.js
```diff
@@ -1,79 +1,177 @@
...
     if (!this.props.isOpen) {
       return null;
     }
 
-    let tetherConfig = this.getTetherConfig();
-
+    const attributes = omit(this.props, Object.keys(propTypes));
     const classes = mapToCssModules(classNames(
       'popover-inner',
       this.props.className
```
SimeonC: Shouldn't `this.props.className` be on the popperClasses instead?
Currently I get the following scenario:

```jsx
<Popover
    className="day-picker-popover"
    placement="bottom"
    isOpen={open}
>
```

Outputs:
```html
<div data-reactroot="" class="popover show bs-popover-bottom" style="position: absolute; transform: translate3d(176px, 489px, 0px); top: 0px; left: 0px; will-change: transform;" data-placement="bottom">
    <div class="popover-inner day-picker-popover">
```

I expected "day-picker-popover" to be on the parent element not the inner as I can't use the classname to. override popover styles in this case.

TheSharpieOne: Yeah, it probably makes more sense to have that go apply to the popper and create a new prop to extend the inner popover classes.

fabiopaiva-innovactory:
Navbar toggleable true is not the expected.
I saw in the code a kind of fallback you can use toggleable or expandable 
https://github.com/SimeonC/reactstrap/blob/v5/src/Navbar.js#L57

If I use `toggleable={true}`, it will add the class `expand`. I am not sure but I think on the beta version it means always expanded

TheSharpieOne:
@fabiopaiva-innovactory I see what you mean. This makes it a little tricky to do the fallback nicely. Before, `true` would add `navbar-toggleable` which make it toggleable, now `true` adds `navbar-expand` which is always expanded as you mentioned.
I'm going to change `expandable` to `expand` to better line up with bootstrap's class and because I feel that makes more sense given the fact that it being `true` doesn't make it expand<i>able</i> but just alway expand(ed)
For the fallback when the deprecated `toggleable` prop is `true`... the only thing similar enough is using the expand sm breakpoint. It also looks like to better maintain expected results when using toggleable, the breakpoint should be bumped (which will happen automatically to not require developers to change their code and thus not be a breaking change) such as `md` -> `lg`.

TheSharpieOne:
reactstrap@5.0.0-alpha.0 has been published to npm using the `next` tag. It can be installed using `npm install reactstrap@next`

TheSharpieOne:
reactstrap@5.0.0-alpha.1 has been published to npm using the `next` tag. It can be installed using `npm install reactstrap@next`
5.0.0-alpha.1 includes performance fix for `Dropdown`, `Dropdown` keyboard navigation/control, and `plaintext` prop on `Input` (deprecating the `static` prop)

med1sh:
Hi, I use warning, danger, success, and muted states. I also like the feature of highlighting the label text for where the field is in a display state. Can we get that ability added back? the valid prop only has 3 states.

Thanks

supergibbs:
Which component do you mean? You can always add `bg-<state>` or `text-<state>` to className on any component. FYI, I think `muted` was replaced with `light`

med1sh:
@supergibbs You can see the demonstrated changes here: https://github.com/reactstrap/reactstrap/commit/8b2386f9b4b36474d5418f7dd25e6b3f805953f2#diff-bc5250a5d5097846d70ffc59fd9a3769

supergibbs:
@med1sh I see, that looks more like a Bootstrap decision. Take a look at the history for [_forms.scss](https://github.com/twbs/bootstrap/commits/v4-dev/scss/_forms.scss) on June 6th.

TheSharpieOne:
@supergibbs is correct. Bootstrap made this change so reactstrap followed suite.
They completely changed validation, most of it is automatically/handled by css. The props reactstrap has trigger the "server-side" aspect, essentially forcing the validation to display. 
https://getbootstrap.com/docs/4.0/components/forms/#server-side

Also, `.text-muted` is still used on bootstrap's forms documentation though it is no longer in the color utilities docs. Looks like it might be replaced by `.text-secondary` but I am not sure.

med1sh:
Ok, cool. Not sure how much I like that change. Thanks

TheSharpieOne:
@med1sh Yeah, I am not a big fan of it either, but it make sense with the change to use the CSS validation pseudo class which only have the two states (valid and invalid).

Also, it looks like `.text-muted` is here to stay.

autarch:
FWIW, with both this branch and the `5.0.0-alpha.1` tag, I cannot build the docs locally. I tried running `yarn install` followed by `yarn run build-docs` and I get this:

```
$  yarn run build-docs
yarn run v1.1.0
$ cross-env WEBPACK_BUILD=production webpack --config ./webpack.dev.config.js --progress --colors
clean-webpack-plugin: /home/autarch/projects/reference/reactstrap/build has been removed.
Hash: f2fe324d8c78773def51  
Version: webpack 1.15.0
Child
    Hash: f2fe324d8c78773def51
    Version: webpack 1.15.0
    Time: 578ms
        + 2 hidden modules
    
    ERROR in bundle.js from UglifyJs
    SyntaxError: Unexpected token: string (bootstrap-css) [./docs/lib/app.js:1,7]
Done in 1.31s.
```

This is pretty frustrating since there's no other way to see the docs AFAICT (cue a rant about how every package should include its full docs and npmjs.com should host said docs, but that's getting pretty OT ;)

TheSharpieOne:
@autarch I need @eddywashere to allow netlify permission to the repo, but it has a really awesome feature which automatically builds and publishes things (such as docs) with every PR. (super handy right?)
I am getting that same thing when doing `build-docs` (with npm), but you can use the `start` cmd/script to start the local "dev" instance which just compiles, watches, and serves the docs to play around locally. I'll look into the build-docs not working.

cr101:
Will V5 work with React 16?

autarch:
I've used both v4 and the v5 alpha with React 16.

virgofx:
It's mostly compatible with R16; however, there are a few issues with some of the components. A few issues have already been created.

PR #598, opened 2017-09-29
fix(Collapse,Fade): Ensuring props don't leak to child and adding new production/limited dist builds.

virgofx:
* Improved inline documentation
* Updated Readme with improved CDN and distribution file information

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -49,21 +49,33 @@ libConfig.targets = [
 
   Defining this config will also check that all peer dependencies are set up
   correctly in the globals entry.
+
+  Reactstrap has two versions:
+
+  1) `reactstrap.min.js`
+      This file excludes `react-popper` and `react-transition-group` from
+      the dist build where they need to be manually required if any
+      application uses components that require these features.
+
+  2) `reactstrap.full.min.js`
+      This file includes all dependencies.
+
 */
-const umdConfig = baseConfig();
-umdConfig.external = peerDependencies;
-umdConfig.plugins.push(replace({
+const umdFullConfig = baseConfig();
+umdFullConfig.external = peerDependencies;
+umdFullConfig.plugins.push(replace({
   'process.env.NODE_ENV': JSON.stringify('production'),
 }));
-umdConfig.plugins.push(babili({ comments: false }));
-umdConfig.targets = [
-  { dest: 'dist/reactstrap.min.js', format: 'umd' },
+umdFullConfig.plugins.push(babili({ comments: false }));
+umdFullConfig.targets = [
+  { dest: 'dist/reactstrap.full.min.js', format: 'umd' },
 ];
-umdConfig.globals = {
+umdFullConfig.globals = {
   react: 'React',
   'react-dom': 'ReactDOM',
 };
-const missingGlobals = peerDependencies.filter(dep => !(dep in umdConfig.globals));
+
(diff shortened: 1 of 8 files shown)


src/CarouselItem.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import Transition, { ENTERING, ENTERED, EXITING } from 'react-transition-group/Transition';
```
TheSharpieOne: Looks like you are removing these enums to better support when ReactTransitionGroup is not included, but can you put these in the utils and reference them from there?

rollup.config.js
```diff
@@ -49,29 +49,56 @@ libConfig.targets = [
...
+
 */
-const umdConfig = baseConfig();
-umdConfig.external = peerDependencies;
-umdConfig.plugins.push(replace({
-  'process.env.NODE_ENV': JSON.stringify('production'),
+const umdFullConfig = baseConfig();
+umdFullConfig.external = peerDependencies;
+umdFullConfig.plugins.push(replace({
+    'process.env.NODE_ENV': JSON.stringify('production'),
```
TheSharpieOne: formatting, 2 spaces (this whole file).

scripts/post-build
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
TheSharpieOne: Any way to support windows <10 which do not have bash? Maybe have a node.js script and have the script in the package.json be `node <file>`

src/Collapse.js
```diff
@@ -92,11 +90,25 @@ class Collapse extends Component {
...
+    // This also gives us more flexibility in the future to remove the prop-types
+    // dependency in distribution builds (Similar to how `react-transition-group` does).
+    // Note: Without omitting the `react-transition-group` props, the resulting child
+    // Tag component would inherit the Transition properties as attributes for the HTML
+    // element which results in errors/warnings for non-valid attributes.
+    const transitionProps = pick(otherProps, TransitionPropTypeKeys);
+    const childProps = omit(otherProps, TransitionPropTypeKeys);
+
     return (
       <Transition
```
TheSharpieOne: I would like to have a basic fallback which instantly shows/hides (no animation), but we would probably want to tie in the callbacks to that to be more compatible.

virgofx: At some point in future we can do better fallback whereby we don't even use the Fade elements for non fading elements. A lot more work needs to be done for that.

src/CarouselItem.js
virgofx: Yes, we can move these to utils. The main reason is that since we are cherry picking an export it prevents using the full ES6 capability for the module in production. (This gives reduced file size in the now `reactstrap.full.min.js`; however, at the expense of not being able to use these.

scripts/post-build
virgofx: I tried to emulate the other scripts that were there. Let me investigate a little further. If we do `node file` does it autochoose the file/extension based on OS? Example of that somewhere?

TheSharpieOne: Those scripts don't work for me (on windows), I have to run those steps manually (which is easy since they are mostly git commands) and they only run on the "release". This is in the build, which is probably more common for others to do.

PR #608, opened 2017-10-02
Cross-platform postbuild support, improved Readme documentation, updated dependencies.

virgofx:


scripts/postbuild.js
```diff
@@ -0,0 +1,26 @@
...
+// ReactPopper) are properly set as optional globals. Specifically, do not
+// error if the user has not loaded these dependencies. Degrade gracefully.
+
+fs.readFile(__dirname + '/../dist/reactstrap.min.js', 'utf8', function(err, data) {
+  if (err) {
+    return console.error(err);
+  }
+
+  var result = data.replace(
+    /\,([a-zA-Z]+)\.ReactTransitionGroup\.Transition/,
```
TheSharpieOne: Should this and the one below have the global flag?

virgofx: Nope, just want to replace the top occurrence (it's only included once); however, to be sure we don't brick anything else only the factory load -- once.

Issue #607, opened 2017-10-02
No animation on modal mounting


Rootmafia:
**## Issue description**

- version "reactstrap": "next",
- components: Modal
- react version "^16.0.0",

There is no fade in animation on Modal appearing, have you removed  it?

virgofx:
This is a valid bug we need to investigate further.

devonJS:
@virgofx hello! could i take a stab at this one? been wanting to contribute for a while, would investigate tomorrow after work!

TheSharpieOne:
@devonJS go for it. There is a branch which updates the dev version of react in this project to 16. You'll probably need to branch off of that or something to see the issue since this only happens with react 16.

virgofx:
I believe the issue is we need to force a repaint on the initial load once it applies the first class. It looks like CSSTransition has this built in; however, we opted to use Transition, so it may behoove us to switchover; however, we need to update this everywhere and ensure we use the bootstrap version of the classes (e.g. `show`).

TheSharpieOne:
There is an issue which was recently open with react-transition-group: https://github.com/reactjs/react-transition-group/issues/216. Not sure if it is valid or not, but it seems similar to what we are experiencing and thus we may be able to benefit from the solution to it; regardless of whether it is an issue with react-transition-group or not

devonJS:
@virgofx @TheSharpieOne Great, thanks for the direction on this issue. Will consider what you both mentioned above when approaching this!

devonJS:
hmmm there seems to be a couple of things to consider for React 16:

1.  As mentioned, react-transition-group seems to not work as intended
2. Seems ReactDOM.createPortal would be the preferred way to accomplish what ReactDOM.unstable_renderSubtreeIntoContainer does [react/issues/10309](https://github.com/facebook/react/issues/10309#issuecomment-318431005), documentation of what it does seems this way as well

ReactDOM.createPortal is such a big feature in React 16 that for this to play with React 16 and its future versions it should be used, what do you think? It also seems lik

TheSharpieOne:
#601 has the task of upgrading to portals, but to better maintain compatibility with < react 16 the suggestion is to use `createPortal` and fallback to `unstable_renderSubtreeIntoContainer` if it is not there. material did [something similar](https://github.com/callemall/material-ui/pull/7218/commits/1d3c5c1719cc78144d9a2fbf21d60857e1c11ae4)
[...]

PR #610, opened 2017-10-03
chore(deps): Update react to 16

TheSharpieOne:
Updates the version of react used in the docs as well as in the tests to 16

Some of the tests around transitions are failing. This is a good thing since it has been reported in #607 that animations/transitions are not working in react 16. When #607 is resolved (and the issues for Collapse and Carousal's animation) the tests should be passing.

**I do need help fixing the transition/animation issues in react 16 and thus getting the tests to pass**


jameswomack:
Should probably be React ^15 || ^ ^16 and not only 16

TheSharpieOne:
@jameswomack The peerDependencies which allows reactstrap to work with various versions of react when itself is installed as a dependency has `^0.14.9 || ^15.3.0 || ^16.0.0`. The devDependencies version which is one npm install when developing is just `^16.0.0` as it cannot be a range and if it was it would pick the latest version within the range to install.

jameswomack:
@TheSharpieOne Makes sense! I misread the PR :)

JonatanSalas:
@TheSharpieOne I have seen that you mention that you need to fix an issue related to animation/transition. I would like to help you with that! Just tell me which test are failing!

TheSharpieOne:
I think the issues have been addressed. The Travis build is failing because is uses yarn and the yarn lock tells it to use react 15 with enzyme react 16 adaptor. I don't really understand why yarn is still a thing, npm 5 addressed a lot of the issue yarn tried to solve. Now it just splits the community and causes issues.

JonatanSalas:
@TheSharpieOne there's a great adoption on yarn, yes, it's true that NPM 5 addresses a lots of issues that yarn tried to solve. In my case, I use it for a project because of the **workspaces feature**. It simplifies a lot when you have a mono repo with multiple packages.

JonatanSalas:
@TheSharpieOne anyway, great to see that you've addressed the issue! 😄

PR #615, opened 2017-10-03
docs(NavbarToggler): add info about placement

TheSharpieOne:
Also updated the placement of the docs toggler itself.


virgofx:
👍

Issue #628, opened 2017-10-10
Close Button in ModalHeader has wrong class when cssModules was passed.


inferusvv:
### Issue description

- components: `ModalHeader`
- reactstrap version `#5.0.0-alpha.3`
- import method `umd/csj/es` (if umd, are you using the "full" version (only for v5+)?)
- react version `#16.0.0`
- bootstrap version `#4.0.0-beta`

### What is happening?

I've included ModalHeader component in next way:

`<ModalHeader cssModule={bs} toggle={this.toggle}>Header</ModalHeader>`

1. `cssModules` prop was passed
2. `toggle` prop was passed

In this case close button still have class `close` but it should take it from css module.

PR #629, opened 2017-10-10
fix(ModalHeader): Use css modules in close button

inferusvv:
Issue: https://github.com/reactstrap/reactstrap/issues/628


TheSharpieOne:
Awesome! Thanks for finding and fixing the issue. You rock! 🤘

PR #631, opened 2017-10-11
 #607 animation on modal mounting

devonJS:
@TheSharpieOne ok so I know this change looks a little ridiculous, but I definitely did a decent amount of investigating and debugging on the timing of when the transition classes are applied, when rendering into subtree is occurs after adding a new div to the document.body, other ways to re-draw the transition after rendering the first time, the list goes on.

Oddly enough, this fix seems to work where adding the `fade show` class only when `status === 'entered'`, allows `Transition` to work correctly with the fade in.  I guess react-transition-group's `'exiting'` to `'entering'` event transition doesn't play well with React 16.

Let me know your thoughts / test it on your machine, it seems to work on my environment.  Sorry, I know it might seem kind of weird that it came to this one line change, but it seems to work and I like to try to keep PRs as simple as possible.

### DIFF
--- a/src/Fade.js
+++ b/src/Fade.js
@@ -57,7 +57,7 @@ function Fade(props) {
   return (
     <Transition {...transitionProps}>
       {(status) => {
-        const isActive = status === 'entering' || status === 'entered';
+        const isActive = status === 'entered';
         const classes = mapToCssModules(classNames(
           className,
           baseClass,


devonJS:
Interestingly, I've modified react-transition-group's Code Sandbox here: https://codesandbox.io/s/vj8lz0oxo5 to reenact the non-transitioning code.  They even use React 16 in their example.

Line 16 & 17 are commented out (their working code) and I added line 18 & 19, which simulates the code where modal animation does not work.  Basically adds `'fade'` class (opacity: 0) on `'exited'` event and `'fade show'` class (opacity: 1) on `'entering'`.  The fade transition no longer works.  However, if you change line 18 & 19 to:
`exited: { opacity: 0 },`
`entered: { opacity: 1 },`

The fade in / out effect works again, this is what my PR encompasses.

virgofx:
Thanks for looking into this @devonJS . I'm going to take a look this weekend. I did notice the same thing; however, that fix doesn't make sense from a logic perspective as it's unintuitive. Let me do a little more digging to see what I can find.

devonJS:
@virgofx Sure, please post your findings to this thread, I'd definitely like to learn how you would approach this issue.

BananaBobby:
Hey guys, I'm also stuck on updating react to v16 as a freelance project uses reactstrap. 

@devonJS It looks like your demo adds just entering style. So while tooltip is entering it gets opacity: 1 for animation, but when it get entered state - it's set back to its default value (0) as there is no provided style for that state. So if you add another line to style config it seems to work correctly: https://codesandbox.io/s/2ww83ro0p0

BananaBobby:
Probably the correct behaviour for isActive flag is `status === 'entered' || status === 'exiting'`
The flow is following (if we use @devonJS sandbox example):
- User interacts with button
- `entering` status is applied and that's a class of hidden tooltip prepared for animation
- `entered` status is applied that should tell the component which is the final state of animation. At this point component should be considered as active
- After we close the tooltip exiting and exited statuses are applied in reversed order, so `exiting` status is still relates to the active element.

devonJS:
@BananaBobby good catch

TheSharpieOne:
I know removing `entering` doesn't make that much sense, but it works. We can address is better later if needed.

Issue #601, opened 2017-09-29
Child ref null in componentDidUpdate of parent containing a Modal component


michaeldzjap:
### Issue description

- version `#5.0.0-alpha.2`
- components: `Modal`, `ModalBody`, `ModalHeader`

After updating to React 16 I noticed that a child component of `ModalBody` containing a ref is `null` in the `componentDidUpdate()` lifecycle method. Not sure if this is caused by React v16 or if it is due to reactstrap. It seems wrong to me, but perhaps there is a good reason for this that I am overlooking? When you close the modal again you do see that the ref is defined.

### Steps to reproduce issue

See [this](https://codesandbox.io/s/xo428w6pnz) CodeSandbox. Open the JS console and click the button to open the modal. The console prints `null` (tested in latest Chrome).

TheSharpieOne:
This is related to https://github.com/facebook/react/issues/9808
It looks like we will need to switch to `unstable_createPortal` but provide a fallback to `unstable_renderSubtreeIntoContainer` when `unstable_createPortal` doesn't exist to keep supporting 15.x.

Chima1707:
@TheSharpieOne sure that was the reason. If you dont mind, can i work on this for this hacktoberfest?

virgofx:
@TheSharpieOne Even better ... `ReactDOM.createPortal(child, container)`   which marked as stable. Just need to change requirement to >= R16. Maybe we can do it next version?

TheSharpieOne:
Sure, go for it. I would also look at how material UI did itc(I saw a PR with something similar to what we need)
Also, yeah, I saw the stable name and that it was aliased to the unstable name. For better compatibility I would use the unstable name (support the R16 betas)

Chima1707:
Ok, guys will send a fix likely this weekend.

TheSharpieOne:
@Chima1707 how is this coming along?

Chima1707:
@TheSharpieOne am investigating the problem, will give my update by the end of tomorrow

leetercola:
Curious about status, this is currently breaking unit tests in jest & enzyme with react 16.

TheSharpieOne:
As am I. @Chima1707 any update? If you are not able to complete it, it's not a big deal, just let us know.

Chima1707:
I have not been able to complete it, please can someone take a look at it?

martindavid:
@TheSharpieOne is anyone working on this at the moment? I'd like to give it a try?

[...]

PR #645, opened 2017-10-20
#601 Modal bugs on react ^16

martindavid:
- Add `ReactDOM.unstable_createPortal` to support Fiber version.
- Keep `ReactDOM.unstable_renderSubtreeIntoContainer` as a fallback for < 16 version.

Issue: https://github.com/reactstrap/reactstrap/issues/601


martindavid:
#601

TheSharpieOne:
I thought I had commented on this earlier, but I apparently did not. The portal API is supposed to be used in the render AFAIK. This means that the logic would be similar, but separated. The render would return null for react <16 and would return the unstable_createPortal for react 16+. Similarly and separately, the opposite would happen in the renderIntoSubtree method; unstable_renderSubtreeIntoContainer would be called for react <15 and nothing would happen for react 16+.

kinke:
Thanks @martindavid, I can confirm that this fixes the `autoFocus` issue for Modals with React 16 (e.g., escape key closing the modal etc.) and the fade-out animation as well.

> The portal API is supposed to be used in the render AFAIK.

@TheSharpieOne: The changes you propose would require more logic changes; `renderIntoSubtree()` is called in 3 places (show(), hide(), componentDidUpdate()).

fantua:
@TheSharpieOne when this and #763 can be released? We have some troubles according to this issues.

Issue #660, opened 2017-10-27
<Input type="select"> size attribute don't work


tafkanator:
### Issue description

if using `<Input type="select" size="5">`
then in DOM, size attribute is not applied

- components: `Input `
- reactstrap version `5.0.0-alpha.3`
- react version `16.0.0`
- bootstrap version `4.0.0-beta`

### Code

https://codesandbox.io/s/504m2rw7lp

gergely-nagy:
Can I do it?

TheSharpieOne:
Sure, but the reactstrap size prop goes to the bootstrap size for input size and not the html size.
Also size is only used for multiple select (AFAIK), so maybe we can have multiple be a boolean or a number and when it is a number, pass that number to the raw underlying select.

gergely-nagy:
Thanks for your quick reply, @TheSharpieOne ! The **size** attribute is not only used for multiple select.
Please refer to the link below.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-size
So  can we use the size prop for html attribute not just for bootstrap size?

TheSharpieOne:
Maybe we should rename the one that determines the bootstrap class and just pass size through. Not too sure what to call it though

gergely-nagy:
Can be renamed to the bsSize.

TheSharpieOne:
That works. We can also deprecated it in a way where if the value of size is a string like lg or sm it will warn and then map it to the class. That way it is not a breaking change

pablougas:
Is there currently any way to add html size to a multiple select when using reactstrap?

TheSharpieOne:
A release was just cut with the changes to allow `size` on input pass through to the DOM. `bsSize` is now used to control the input's bootstrap size class

pawel-n:
It's working now. But I got warning: 
`Warning: Failed prop type: Invalid prop 'size' of type 'number' supplied to 'Input', expected 'string'`

components: Input
reactstrap version 5.0.0-alpha.4
bootstrap version 4.0.0-beta.2

TheSharpieOne:
Use a string such as `size="5"` just like you would for html.

pawel-n:
my bad. Thanks.

john-tipper:
This appears to have regressed and is still a problem. If I use a string such as @TheSharpieOne suggests then I get a TS error:

```
TS2769 
Type 'string' is not assignable to type 'number | undefined'
```

[...]

PR #662, opened 2017-10-28
fix(Input): fix size prop (#660)

gergely-nagy:


src/Input.js
```diff
@@ -3,12 +3,13 @@
...
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { mapToCssModules, deprecated } from './utils';
+import { mapToCssModules, deprecated, warnOnce } from './utils';
 
 const propTypes = {
   children: PropTypes.node,
   type: PropTypes.string,
   size: PropTypes.string,
+  bsSize: PropTypes.oneOf(['lg', 'sm']),
```
TheSharpieOne: Don't limit this. If someone wants to extend bootstrap to add something like xl (or anything else) they should be able to just add their class and pass it here.

src/Input.js
```diff
@@ -72,11 +73,16 @@ class Input extends React.Component {
       }
     }
```
TheSharpieOne: I'd do a regex check to see if it is a word/not a number to trigger this logic. It works with the custom sizes better.

src/Input.js
```diff
@@ -72,11 +73,16 @@ class Input extends React.Component {
...
 
+    if (['lg', 'sm'].indexOf(attributes.size) > -1) {
```
TheSharpieOne: You should also delete size from attributes so it doesn't get passed down to the DOM in this case.

src/Input.js
```diff
@@ -43,6 +44,7 @@ class Input extends React.Component {
     } = this.props;
 
     const checkInput = ['radio', 'checkbox'].indexOf(type) > -1;
+    const isNotaNumber = new RegExp('(?!^\\d+$)^.+$');
```
TheSharpieOne: `/\D/g` would work and is more clear. `new RegExp('\\D','g')` if you want to use `new RegExp`

gergely-nagy: Yes , this is much better. Sorry, i forgot this global flag . Thank you for taking your time for the review. I'll fix it right away.

PR #674, opened 2017-11-08
fix(Modal): Do not trigger toggle when transitioning out

aripekkako:
I'm not sure, if the approach here is suitable. Using debounce will likely have some corner cases (such as the toggle callback taking a very long time to decide whether or not to close the modal). 

However, using "isOpen" seems to be the best way to check, if we're transitioning out. If you can suggest a better way, I'm happy to change the PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
   "homepage": "https://github.com/reactstrap/reactstrap#readme",
   "dependencies": {
     "classnames": "^2.2.3",
+    "lodash.debounce": "^4.0.8",
     "lodash.isfunction": "^3.0.8",
     "lodash.isobject": "^3.0.2",
     "lodash.tonumber": "^4.0.3",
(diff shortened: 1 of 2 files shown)


TheSharpieOne:
Closing in favor of https://github.com/reactstrap/reactstrap/commit/2cdf225504a2eb0d96cd4fc9cfec90fc093b6ae0#diff-3ddc036af206541151e86f03cfc0c0ee which also handles when an external action causes `isOpen` to be rapidly changed.
The root of the issue was the order of events which caused the dialog reference to be null
1. modal toggled opened, rendered into subtree, dialog reference set
2. modal toggled closed, transition out starts
3. modal toggled opened, , rendered into subtree, dialog reference set
4. transition finishes, dialog reference **unset**
5. modal toggled closed -- error dialog reference is undefined.

If we check for the existence of the dialog reference, then we can tell if a modal is open (or transitioning)
If it tried to open while the reference already exists (while it's already open), we will not attempt to open another one. We also call toggle to "sync" the state of `isOpen` so that it will toggle back to `false` since the modal is closing/transitioning out.

PR #685, opened 2017-11-14
docs(Modal): fix nested modal closing

gergely-nagy:
Added `onClosed` prop according to #400 .


TheSharpieOne:
You got rid of the ability to close just the nested modal; that option should remain. When the dismiss all button is clicked, it should close the nested modal, then when that modal is closed the first modal should close (via the onClosed of the nested modal)
It's a little tricky, but dismiss all button should toggle `this.state.nestedModal` to `false` and set a flag `this.state.closeAll` to `true`. The `onClose` will be a function which will always run any only trigger `this.toggle` if `this.state.closeAll` is `true`.

Issue #637, opened 2017-10-14
Dropdown in Navbar


supergibbs:
### Issue description

- components: `NavDropdown`
- reactstrap version `#5.0.0-alpha.3`
- bootstrap version `#4.0.0-beta`

### What is happening?
In a navbar, the Dropdown still uses popper to position it

### What should be happening?
It should just use Bootstrap's CSS when inside a Navbar
Bootstrap disables popper in a navbar: https://github.com/twbs/bootstrap/blob/v4-dev/js/src/dropdown.js#L146
Example here: https://getbootstrap.com/docs/4.0/components/navbar/ (especially noticeable when navbar is collapsed)

maaz93:
@supergibbs I'd like to take this up. I think I understand the issue, but I'd like to get it clarified before I begin trying to fix this:

Is it that `Dropdown` is currently agnostic of the fact that whether it is in a `NavDropdown`  or not? And `Dropdown`'s `render` returns a `react-popper` `Manager` always?

Does it mean I need to change [this](https://github.com/reactstrap/reactstrap/blob/master/src/Dropdown.js#L186) to use `Tag` when it's in a `NavDropdown`?

I'm new to this project and I could've understood it completely wrong. Please feel free to correct me and guide me in the right direction.

supergibbs:
Yes, the current Dropdown always uses popper but it should only use it when not in a `Navbar`. 

Note, this is different than just modifying `<NavDropdown />` (or `<Dropdown nav />` once #636 is merged). It needs to know that it's in a Navbar, possibly with a new prop if needed.

maaz93:
Alright. I'll start working on it. @TheSharpieOne, could you please add the in-progress tag for this, since I'm working on it?

maaz93:
@supergibbs I've been looking around the code and seeing how I can fix this. Since `Navbar` is a stateless component, I've to first make it stateful and then use `context` to pass down a `isInNavbar` prop to its children. Then, I'll have to use this `isInNavbar` flag in `Dropdown` as well as its children in order to remove any dependencies on `react-popper`.

Is this the best possible solution or you have other suggestions?

TheSharpieOne:
Probably just best to add a navbar prop to the dropdown and trigger logic off if that.

supergibbs:
Hey @maaz93, how's it going? Just a heads up that https://github.com/reactstrap/reactstrap/pull/636 was merged in which may affect your PR.

[...]

PR #692, opened 2017-11-17
Disable popper when inNavbar

supergibbs:
For https://github.com/reactstrap/reactstrap/issues/637


TheSharpieOne:
Tests are needed for the new/different path the code can take when the prop is `true`. Also, sadly the docs do not have an example of a dropdown in a navbar (maybe one can be added?)

supergibbs:
I agree but I am a little lost on how to check that popper isn't used when the prop is set. I can't figure out how to access the style attribute (to ensure it's null) in the final rendered. Other suggestions?

Also FYI, on my mac, `brew install watchman` allows me to run the tests locally.

supergibbs:
@TheSharpieOne Do you have any suggestions on how to test this change?

TheSharpieOne:
You should be able to get the [type](https://github.com/airbnb/enzyme/blob/master/docs/api/ReactWrapper/type.md) of the component which is being rendered and check that it isn't Popper

supergibbs:
@TheSharpieOne Good idea, I added a test for `inNavbar` and updated from master. Let me know if there is anything else!

Issue #575, opened 2017-09-16
Tracking Beta 2 Changes


TheSharpieOne:
https://github.com/twbs/bootstrap/issues/23278

 - [x] Renamed `.table-inverse`, `.thead-inverse`, and `.thead-default` to `.*-dark` and `.*-light`, matching the color schemes used elsewhere in.
 - [x] Responsive tables now generate classes for each grid breakpoint, meaning we've added `.table-responsive-{sm,md,lg,xl}` to the already present `.table-responsive`. You might need to adjust your usage depending on when you want a table to resize.
 - [x] Remove Col `push-*-*` and `pull-*-*` Add Col `order-*-*`
 - [x] Docs: remove message about form validation (Bootstrap: Fix broken valid/invalid sibling selector for form validation mixin.)

TheSharpieOne:
Bootstrap v4 Beta 2 has shipped

tyjohnnew:
https://blog.getbootstrap.com/2017/10/19/bootstrap-4-beta-2/
Yah I second this.

flipxfx:
I've also noticed the grid is in need of some updating.

Needs:
- Remove Col `push` and `pull`
- Add Col `order-`

Other nice to haves on grid:
- Add Col `align-self-`
- Add Row `justify-content-`
- Add Row `align-items-`
- Add Row `align-content-`

supergibbs:
From http://getbootstrap.com/docs/4.0/migration/#beta-2-changes: 

- Removed the unused `.form-control-label` class. If you did make use of this class, it was duplicate of the `.col-form-label` class that vertically centered a `<label>` with it’s associated input in horizontal form layouts.

https://github.com/reactstrap/reactstrap/blob/master/src/Label.js#L104 needs to be updated

ddantzler:
Support for form-row would be nice.

PR #694, opened 2017-11-17
BS4 beta 2 updates for Table

supergibbs:
https://github.com/reactstrap/reactstrap/issues/575


src/Table.js
```diff
@@ -44,16 +46,17 @@ const Table = (props) => {
...
-    inverse ? 'table-inverse' : false,
+    (dark || inverse) ? 'table-dark' : false,
     hover ? 'table-hover' : false,
     reflow ? 'table-reflow' : false
   ), cssModule);
 
   const table = <Tag {...attributes} className={classes} />;
 
   if (responsive) {
+    const responsiveClassName = responsive === true ? 'table-responsive' : `table-${responsive}-responsive`;
```
TheSharpieOne: looks like it should be `table-responsive-${responsive}`.
Also looks like they killed the wrapper needed for responsive (something that was missing in alpha.6 upgrade)

src/Table.js
```diff
@@ -1,18 +1,19 @@
...
   size: PropTypes.string,
   bordered: PropTypes.bool,
   striped: PropTypes.bool,
-  inverse: PropTypes.bool,
+  inverse: deprecated(PropTypes.bool, 'Please use the prop "dark"'),
+  dark: PropTypes.bool,
   hover: PropTypes.bool,
   reflow: PropTypes.bool,
-  responsive: PropTypes.bool,
+  responsive: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['sm', 'md', 'lg', 'xl'])]),
```
TheSharpieOne: I would just allow a string in case someone wants to extend bootstrap to add additional/custom breakpoints 
```
PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
```

src/Table.js
```diff
@@ -44,16 +46,17 @@ const Table = (props) => {
     size ? 'table-' + size : false,
     bordered ? 'table-bordered' : false,
     striped ? 'table-striped' : false,
-    inverse ? 'table-inverse' : false,
+    (dark || inverse) ? 'table-dark' : false,
     hover ? 'table-hover' : false,
     reflow ? 'table-reflow' : false
```
TheSharpieOne: Just noticing reflow was removed in bootstrap alpha.6, can you kill it in this PR :pray:

src/Table.js
supergibbs: But the tests pass? :P wow, can't believe I missed that!

supergibbs:
@TheSharpieOne I think this one is good to go. Let me know if there is anything else.

PR #696, opened 2017-11-17
Replace pull/push with order

supergibbs:
Removed `pull`/`push` from `columnProps` and added `order` (in `Col` and `Label`). I deprecated the props but they won't work at all now; too complex to convert to `order`.

https://github.com/reactstrap/reactstrap/issues/575#issuecomment-338847760

### DIFF
--- a/src/Col.js
+++ b/src/Col.js
@@ -2,7 +2,7 @@ import isobject from 'lodash.isobject';
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { mapToCssModules } from './utils';
+import { mapToCssModules, deprecated } from './utils';
 
 const colWidths = ['xs', 'sm', 'md', 'lg', 'xl'];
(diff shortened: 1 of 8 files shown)


src/Col.js
```diff
@@ -78,8 +79,7 @@ const Col = (props) => {
 
       colClasses.push(mapToCssModules(classNames({
         [colClass]: columnProp.size || columnProp.size === '',
-        [`push${colSizeInterfix}${columnProp.push}`]: columnProp.push || columnProp.push === 0,
-        [`pull${colSizeInterfix}${columnProp.pull}`]: columnProp.pull || columnProp.pull === 0,
+        [`order${colSizeInterfix}${columnProp.order}`]: columnProp.order || columnProp.order === 0,
```
TheSharpieOne: I don't think `0` should trigger a class to be added (`order-*-0`). previously, push and pull could be "reset" for a breakpoint (where `0` came into play), but it doesn't look like order has that ability.
This is just a note as I am going to leave it in as the end user should never use `0` his/her self. Also, bootstrap may add `0` in the future to allow a reset

Issue #709, opened 2017-11-24
Allow passing modifiers to react-popper's Popper component


philipstanislaus:
# Use case

We have some Popovers targets in `overflow: auto` parent elements. By default, popper.js' preventOverflow modifier restricts the popper to `boundariesElement: scrollParent`. This leads to unwanted placement of the popper. To fix it, I would like to set the `boundariesElement` to `viewport`, which is not possible on a Popover right now. 

I am happy to help with a PR if the general approach of passing modifiers to react-popper's Popper component is approved by the maintainers.

TheSharpieOne:
Yeah, it should accept modifiers and spread them on top of the modifiers we end up creating based on some of the other props. It should be known that passing certain modifiers will override the ones created be certain props.
Yes, a PR would be awesome. 
PopperContent needs to accept and use the modifiers. Tooltip and Popover need to explicitly pass modifiers to PopperContent (as it currently gets spread on top another element). Don't forget tests and documentation :wink:

philipstanislaus:
Awesome, I created a PR: #710 Looking forward to some feedback :)

ebuchholz:
Would also love to have this. ❤️❤️

PR #710, opened 2017-11-24
feat(Popover/Tooltip): add possibility to pass modifiers to Popper.js

philipstanislaus:
Implemented popper modifiers on Popover and Tooltip components, as discussed in #709. Please review :)


philipstanislaus:
@TheSharpieOne could you release a new version that includes this PR? It has been 6 weeks. Thanks 👍

PR #724, opened 2017-11-30
Table Inverse doesn't work

TylerShaw:
It is supposed to be table-dark in Bootstrap 4


TylerShaw:
Nevermind, someone has a PR for this already. Sorry about that.

PR #729, opened 2017-12-05
Exposing a global cssModule setter

jlaramie:


src/utils.js
```diff
@@ -44,7 +44,13 @@ export function conditionallyUpdateScrollbar() {
   }
 }
 
-export function mapToCssModules(className, cssModule) {
+let globalCssModule = {};
```
TheSharpieOne: Don't default this to an empty object, leave it undefined so that when nothing is provided `mapToCssModules` will do nothing.

jlaramie: Made the change

PR #758, opened 2018-01-01
chore(docs): upgrade to bootstrap 4.0.0-beta.3

supergibbs:
Also changes `col-form-legend` to `col-form-label` in examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Then open [http://localhost:3000/](http://localhost:3000/) to see your app. The 
 Install reactstrap and Bootstrap from NPM. Reactstrap does not include Bootstrap CSS so this needs to be installed as well:
 
 ```
-npm install --save bootstrap@4.0.0-beta.2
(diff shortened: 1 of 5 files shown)


kraenhansen:
Unfortunately I don't think it's this easy - beta.3 contains [breaking changes](https://blog.getbootstrap.com/2017/12/28/bootstrap-4-beta-3/#breaking-changes) - ex. the `InputGroupButton` doesn't work as expected with CSS from `bootstrap 4.0.0-beta.3`.

Also - "beta.2" is mentioned [a couple of places](https://github.com/reactstrap/reactstrap/search?utf8=%E2%9C%93&q=beta.2&type=) that probably needs updating too.

supergibbs:
You are right, there is more to update than just the `package.json` file but those will be separate PRs. I've already done #757 and am working on #759; hopefully will have ready later today. This PR should be merged in when others are too.

PR #760, opened 2018-01-02
Update docs to install bootstrap using --save-exact

kraenhansen:
As the newly released `bootstrap@4.0.0-beta.3` is semantically compatible with `bootstrap@4.0.0-beta.2` users needs to install using `--save-exact`, to avoid the breaking changes from beta.3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Then open [http://localhost:3000/](http://localhost:3000/) to see your app. The 
 Install reactstrap and Bootstrap from NPM. Reactstrap does not include Bootstrap CSS so this needs to be installed as well:
 
 ```
-npm install --save bootstrap@4.0.0-beta.2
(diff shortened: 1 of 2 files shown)


TheSharpieOne:
AFAIK if you have the exact version specified and don't have the caret (`^`) it will automatically save-exact.

supergibbs:
I see @kraenhansen's point. With `--save`, `version: "^4.0.0-beta.2"` is added and without a lock file, `npm install` would install `4.0.0-beta.3`. Using `--save-exact` will set the `bootstrap` dependency to `version: "4.0.0-beta.2"`.

That being said, since the docs have been updated to `beta.3` this might not matter and once bootstrap stops releasing breaking changes, I don't think `reactstrap` should encourage locking to a specific version.

PR #771, opened 2018-01-07
Move to PureComponent & fix stateless irregularities

Cretezy:
This PR has 3 goals:

* Move all Components to PureComponents, to make them not update when not necessary.
    * This required some fixes in the `Modal.spec.js`, which assumed updates with no prop change.
* Make all stateful components use React.PureComponent.
    * This replaces `import React, { PureComponent } from "react"` to simply use `React.PureComponent`.
* Fix irregularities in stateless ("function") components.
    * The majority used the arrow function declaration, but some used regular function statements. This PR moves them all the arrow function.


TheSharpieOne:
> React.PureComponent’s shouldComponentUpdate() only shallowly compares the objects. If these contain complex data structures, it may produce false-negatives for deeper differences. Only extend PureComponent when you expect to have simple props and state, or use forceUpdate() when you know deep data structures have changed. Or, consider using immutable objects to facilitate fast comparisons of nested data.
Furthermore, React.PureComponent’s shouldComponentUpdate() skips prop updates for the whole component subtree. Make sure all the children components are also “pure”.

Since it doesn't check context changes and also forces users to ensure all of the children components are also "pure", I don't believe this to be a good fit for this library. This library needs to be as flexible as possible and I believe this will make it more difficult to use.

abecks:
I am seeing a lot of unnecessary re-rendering when using the Reactstrap components (via `why-did-you-update`). Since you didn't want to adopt `PureComponent`, do you have any plans to address this problem via `shouldComponentUpdate`?

You may want to reconsider using `PureComponent`. While it does not do a deep comparison of props or state, it DOES do a reference comparison, on each prop and state which I've found to be sufficient. PureComponent really just does some basic `shouldComponentUpdate` logic for you.

PR #782, opened 2018-01-16
window.getComputedStyle can return null in Firefox

### LINKED ISSUE #781: window.getComputedStyle can return null in Firefox
### Issue description

- components: `Modal`
- reactstrap version: latest

### What is happening?

There is a bug in firefox which makes getComputedStyle return `null` in an iframe set to `display:none`.
see: https://bugzilla.mozilla.org/show_bug.cgi?id=548397

Pull request is coming

fantua:
Fixes #781. 

Can it be reviewed and released? Currently we can't init our app in firefox.


fantua:
cc @TheSharpieOne

Issue #761, opened 2018-01-02
Issue with Modal on nested modals unmount.


mmartyniuk:
### Issue description

- components: `Modal`
- reactstrap version `5.0.0-alpha.4`
- react version `16.2.0`
- bootstrap version `4.0.0-beta.2`

### What is happening?

modal-open class is removing from <body> when I have a modal which is always opened and nested modals in it which are mounting and un-mounting when it is needed.

### What should be happening?
modal-open class for parent modal should not be removed till i not close it.
### Steps to reproduce issue
1) Go to https://codepen.io/anon/pen/WdOgPd
2) find body element 
3) click on the buttons and notice that 'modal-open' class disappears while main modal window is opened.

TheSharpieOne:
Looks like an easy fix if anyone wants to put up an PR. Just need to check that the modal is open before removing the class (probably want to also wrap the other things (like the scrollbar stuff) in the same check)

https://github.com/reactstrap/reactstrap/blob/master/src/Modal.js#L159-L160

mmartyniuk:
Please note that opening nested modal adds one more 'modal-open' class to body.

PR #796, opened 2018-01-30
feat(Modal): added support for React Portal

fantua:
- Added `react-portal` package (uses React v16 and its official API for creating portals; has a fallback for React v15).
- Updated `enzyme-adapter-react-16` to latest version for `Portals` support.

1. Currently `handleBackdropClick` is fires for nested modals and close all opened. Need to investigate more on this. 
2. Most test is failing by: 
```js
expect(wrapper.childAt(0).children().length).toBe(0);
```
After using `ReactDOM.createPortal` it's not more 0. Can I remove this checks?

Need some reviews/comments to current changes.

Refers to #601, #761.

### DIFF
--- a/src/__tests__/Modal.spec.js
+++ b/src/__tests__/Modal.spec.js
@@ -25,7 +25,7 @@ describe('Modal', () => {
     jest.clearAllTimers();
   });
 
-  it('should render with the class "modal-dialog"', () => {
+  it('should render modal portal into DOM', () => {
     isOpen = true;
     const wrapper = mount(
       <Modal isOpen={isOpen} toggle={toggle}>
@@ -35,6 +35,18 @@ describe('Modal', () => {
 
     jest.runTimersToTime(300);
     expect(wrapper.childAt(0).children().length).not.toBe(0);
+    wrapper.unmount();
+  });
+
+  it('should render with the class "modal-dialog"', () => {
+    isOpen = true;
+    const wrapper = mount(
+      <Modal isOpen={isOpen} toggle={toggle}>
+        Yo!
+      </Modal>
+    );
+
+    jest.runTimersToTime(300);
     expect(document.getElementsByClassName('modal-dialog').length).toBe(1);
     wrapper.unmount();
   });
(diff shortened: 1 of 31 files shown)


fantua:
cc @TheSharpieOne @supergibbs

supergibbs:
I haven't had a chance to mess around with portals yet but would it be easier to just wrap a `Modal` in another component in your app and tell it to render to a portal? Since reactstrap is still supports react 15 not sure if it should have react 16 stuff built in.

fantua:
> I haven't had a chance to mess around with portals yet but would it be easier to just wrap a Modal in another component in your app and tell it to render to a portal?

It's doesn't make sense because currently `Modal` uses `unstable_renderSubtreeIntoContainer`. `renderSubtreeIntoContainer` isn't stable and have some bugs (ex. #601), my changes allow you to:
- use `Portal` for 16 (which is stable)
- fix component logic and structure (currently it's looks a little bit weird)
- leaves fallback for 15

seberik:
Calling destroy in componenDidUpdate breaks the modal close transition.

TheSharpieOne:
Exit animations are not working
Closing a nested model closes all of the modals. (not even clicking the backdrop)
For the tests, you can change them however you need to; just ensure they have good/complete coverage and are actually testing (not just running through the code to get coverage) and pass.

fantua:
Added new changes:
- fixed Modal exit animation,
- fixed nested Modals,
- updated react-transition-group to 2.2.1

@seberik @TheSharpieOne, guys take a look at this, looks like now all should be good 😃

seberik:
The exit animation now works but when double-clicking the backdrop or escape-key it will close and re-open the modal.

fantua:
Fixed backdrop  double-clicking and tests.

fantua:
Merged with master. Waiting for review.

cc @TheSharpieOne

src/__tests__/Modal.spec.js
```diff
@@ -76,7 +76,7 @@ describe('Modal', () => {
     );
 
     jest.runTimersToTime(300);
-    expect(wrapper.childAt(0).children().length).toBe(0);
+    expect(wrapper.childAt(0).children().length).not.toBe(0);
```
TheSharpieOne: You can remove all of the checks like this. The previous test was ensuring nothing was inserted into the DOM, but now using react-portal, we will always be inserting react-portal, so these checks are more/less pointless (unless you want to check that react-portal get inserted, but I would also only do that in one test, not all of them.)

fantua: Removed those unneeded checks and created new test for this.

PR #802, opened 2018-02-01
fixed wrong usage of state toggle

arteniioleg:
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

### DIFF
--- a/docs/lib/examples/NavbarToggler.js
+++ b/docs/lib/examples/NavbarToggler.js
@@ -12,7 +12,7 @@ export default class Example extends React.Component {
   }
 
   toggleNavbar() {
-    this.setState((prevState, props) => ({
+    this.setState(prevState => ({
       collapsed: !prevState.collapsed
     }));
   }


virgofx:
No need for the second parameter.

arteniioleg:
Done

supergibbs:
Interesting, I didn't know that. The uncontrolled components should prob updated as well. Want to add those to your commit?

TheSharpieOne:
I don't think this is the wanted solution. I understand the issue the react docs are referring to, but it really doesn't apply here.
I believe that the value should be idpontent given the current value in the state. If you think about it, when the value within the state is false and you click it, you would expect it to be true on the next render no matter how many times it runs. It would be similar to ignoring the trigger while the action executes. Some thing do something similar in the "animating"/"transitioning" state.

vanyadymousky:
@TheSharpieOne I guess you have to dig bit more about how setState queues work.

TheSharpieOne:
If we queue up animation, sure, but we don't. If it's visually open and I click to close it, it should close no matter how many times it's clicked before the async batch state reconciliation happens.
With this fix, if it's clicked again before this reconciliation (and subsequent render) happens, it will base the state value on the previous value which was never rendered.

The current way:
User click (calls toggle) 
State value: true
setState to false
User click (calls toggle) 
State value: true
setState to false
Batch update happens, state value is false

Proposed
User click (calls toggle) 
Functional setState queued 
State value: true
User click (calls toggle) 
Functional setState queued
Batch update happens,
Functional setState sets value to false
Functional setState sets value to true
Final state value is true
(user's actions appear to do nothing and a github issue is opened)

Yes, the user clicked it twice and it should probably toggle twice, but sadly a lot of people think everything is a doubleclick. This is just a sample as well, so users can implement this however they want in their projects.

arteniioleg:
> but sadly a lot of people think everything is a doubleclick

So leaving the current behavior is only to not confuse the double-clickers.

What about the case when in tests I simulate a behavior then check if the toggle is in the right state?There things will happen fast and I assume the batch update will happen.
It will be nice to have an actual test code example that proves this (but at the moment I am lazy).

TheSharpieOne:
If everyone agrees that the result of function set state is more desirable (which is what it sounds like) and not just pointing out react's docs explaining how to use it with a basic counter being the reason (because I can point out docs for inline styles and make the same case for never using className (most of this library is just applying classNames)), I'll make the change. It's not that big of a deal especially since it's only an example.
(I'm pretty sure other toggle methods (such as for the uncontrolled stuff) also work this way. Those are actual code that is used and not just examples. They probably should have the same desired affect)

arteniioleg:
Ok.

> https://github.com/reactstrap/reactstrap/pull/802#issuecomment-362924804 Interesting, I didn't know that.

At least now you are aware of this and maybe it will be helpful in future.

PR #805, opened 2018-02-01
test(UncontrolledNavDropdown): add missing tests

gergely-nagy:
Added some tests for UncontrolledNavDropdown


Sunnysh123:
Will it cover all the scenarios @gergely-nagy

TheSharpieOne:
After looking at what UncontrolledNavDropdown was doing I had to make some changes which will invalidate this PR.
UncontrolledNavDropdown is deprecated and was returning another deprecated component while recommending users to use a different component altogether.
UncontrolledNavDropdown will no longer render NavDropdown, instead it will render UncontrolledDropdown with the nav prop.

gergely-nagy:
Okay, thank you for review! I will close this PR.

Issue #785, opened 2018-01-18
Adding "dropright"/"dropleft" prop to Dropdown component


DhawalRank:
### Issue description

- components: `Dropdown`
- reactstrap version: `4.8.0`
- import method 'reactstrap/DropDown';
- react version `16.2.0`
- bootstrap version `4`

### What is happening?

Nesting a dropdown into another dropdown

### What should be happening?

There should be a prop option "drop" right as provided by bootstrap as a class named "dropright" to open the dropdown menu on the right of the dropdown button

TheSharpieOne:
The `DropdownMenu` has a prop `right` which accepts a boolean to indicate whether or not the menu should open to the right. 
See the Alignment section of https://reactstrap.github.io/components/dropdowns/

flipxfx:
Should we create a new issue to get reactstrap caught up with the official release?

TheSharpieOne:
We can just use individual issues and add them to the same milestone or project to group them together.

supergibbs:
Instead of adding more props I deprecated `dropup` and replaced with `dropDirection`. You can set `dropDirection` to  "down" (default), "up", "left" or "right". Is still backwards compatible with `dropup` IF you don't use `dropDirection`.  See #813

PR #813, opened 2018-02-07
feat(Dropdown): Add support for dropleft and dropright (#785)

supergibbs:
Add support for dropleft and dropright (#785)

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -10,8 +10,8 @@ import { mapToCssModules, omit, keyCodes, deprecated } from './utils';
 
 const propTypes = {
   disabled: PropTypes.bool,
-  dropup: deprecated(PropTypes.bool, 'Please use the prop "dropDirection"'),
-  dropDirection: PropTypes.oneOf(['up', 'down', 'left', 'right']),
+  dropup: deprecated(PropTypes.bool, 'Please use the prop "direction"'),
+  direction: PropTypes.oneOf(['up', 'down', 'left', 'right']),
   group: PropTypes.bool,
   isOpen: PropTypes.bool,
   nav: PropTypes.bool,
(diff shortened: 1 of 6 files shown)


src/Dropdown.js
```diff
@@ -6,11 +6,12 @@ import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
 import { Manager } from 'react-popper';
 import classNames from 'classnames';
-import { mapToCssModules, omit, keyCodes } from './utils';
+import { mapToCssModules, omit, keyCodes, deprecated } from './utils';
 
 const propTypes = {
   disabled: PropTypes.bool,
-  dropup: PropTypes.bool,
+  dropup: deprecated(PropTypes.bool, 'Please use the prop "direction"'),
```
TheSharpieOne: I would also indicate the using the value "up": `'Please use the prop "direction" with the value "up".'`

src/Dropdown.js
```diff
@@ -192,7 +195,9 @@ class Dropdown extends React.Component {
         [`btn-group-${size}`]: !!size,
         dropdown: !group && !addonType,
         show: isOpen,
-        dropup: dropup,
+        dropup: direction === 'up',
```
TheSharpieOne:
```
direction !== 'down' && `drop${direction}`
```

src/DropdownMenu.js
```diff
@@ -41,7 +41,23 @@ const DropdownMenu = (props, context) => {
 
   if (context.isOpen && !context.inNavbar) {
     Tag = Popper;
-    const position1 = context.dropup ? 'top' : 'bottom';
+
+    let position1;
+    switch (context.direction) {
```
TheSharpieOne: You can use a hashmap instead and it will be more performant and less code.
```
// this defined outside of the component:
const directionPositionMap = {
  up: 'top',
  left: 'left',
  right: 'right',
  down: 'bottom',
};
// this right here:
const position1 = directionPositionMap[context.direction] || 'bottom';
```
You don't really need the `|| 'bottom'` as the contextTypes should ensure it is in the hashmap.... but just in case?

Issue #814, opened 2018-02-07
Popover with 'Hidden Arrow'


MaximKudriavtsev:
### Issue description

- components: `Popover`
- reactstrap version `5.0.0-beta`
- react version `16.2.0`
- bootstrap version `v4.0 latest`

### What is happening?

I want to hide the arrow from the Popover component. For this, I use the Popover's `modifiers` property, but it does not work. The arrow is still visible.
What am I doing wrong. Please, provide me with a code sample that implements a Popover with 'Hidden Arrow'.

### What should be happening?

![image](https://user-images.githubusercontent.com/15093892/35901851-df1b8da2-0be9-11e8-9c36-090aae80d928.png)

### Steps to reproduce issue

I reproduce my case in this [example](https://stackblitz.com/edit/react-hznpij?file=index.js).

supergibbs:
I think that `enabled` property refers to the modifier of the arrow, not the arrow itself. Wouldn't be hard to add a prop to remove the arrow though. Needs to be added to `PopperContent` then passed through from any components that use it.

TheSharpieOne:
@supergibbs that's what I was thinking because we just hardcode the arrow to always be there. The prop would just conditionally not add it:
https://github.com/reactstrap/reactstrap/blob/2b71fd633ece693089c35ad32d44df84d5776cb3/src/PopperContent.js#L163-L166

gergely-nagy:
so.. @TheSharpieOne @supergibbs , We should add a prop something like "hasArrow" ? This prop is true by default. When prop has false value, must the arrow not to be displayed. I'm right? I can add a PR for this issue.

supergibbs:
Ya, go for it. I think the prop name should just be `arrow` to match the naming style of reactstrap. Thoughts @TheSharpieOne? Needs to be added to `PopperContent` and `Popover` and `Tooltip`

TheSharpieOne:
reactstrap has things like `isOpen` instead of just `open`. `hasArrow` fits that.... but I really dislike props that default to `true` making the user have to do `hasArrow={false}` which seems way verbose. Maybe `noArrow` or `hideArrow` which defaults to `false` so that the user only has to type the prop name which sets the value to `true`

TheSharpieOne:
Fixed via 43efe8b

PR #815, opened 2018-02-07
feat(Popover/Tooltip): add ability to hide arrow (#814)

gergely-nagy:
Added `hideArrow` prop to Popover, Tooltip and PopperContent according to #814  .

### DIFF
--- a/src/PopperContent.js
+++ b/src/PopperContent.js
@@ -10,6 +10,7 @@ const propTypes = {
   className: PropTypes.string,
   placement: PropTypes.string,
   placementPrefix: PropTypes.string,
+  hideArrow: PropTypes.bool,
   tag: PropTypes.string,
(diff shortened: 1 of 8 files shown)


src/PopperContent.js
```diff
@@ -162,7 +165,7 @@ class PopperContent extends React.Component {
     return (
       <ReactPopper modifiers={extendedModifiers} {...attrs} component={tag} className={popperClassName}>
         {children}
-        <Arrow className={arrowClassName} />
+        {hideArrow ? null : <Arrow className={arrowClassName} />}
```
supergibbs: This syntax is a little nicer `{hideArrow && <Arrow className={arrowClassName} />}`. Looks good otherwise!

gergely-nagy: Okay,  that was a dilemma for me too. I will fix it.

PR #817, opened 2018-02-08
fix(Modal): map body class name before removing

### LINKED ISSUE #665: Closing modal doesn't unlock the main body
### Issue description

- components: `Modal`
- reactstrap version `#5.0.0-alpha3`
- import method `npm + import`
- react version `#16.0.0`
- bootstrap version `#4.0.0-beta`

### What is happening?

The modal background isn't clearing to allow access to the main body after closing it.

### What should be happening?

Not that...

### Steps to reproduce issue

1. Open a modal.
2. Close it.

### Error message in console

```
None
```

### Code

https://codepen.io/anon/pen/YEXNMv
 
There might be something else elsewhere in my code that causes this as well, since that link doesn't actually have the same issue, but I get no errors so I'm not sure how to quickly narrow it down.

TheSharpieOne:
Closes #665, Closes #763

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -154,7 +154,7 @@ class Modal extends React.Component {
       document.body.removeChild(this._element);
       this._element = null;
     }
-    
+
     if (!this.props.isOpen) {
       const modalOpenClassName = mapToCssModules('modal-open', this.props.cssModule);
       // Use regex to prevent matching `modal-open` as part of a different class, e.g. `my-modal-opened`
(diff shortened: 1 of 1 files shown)


src/Modal.js
```diff
@@ -155,10 +155,13 @@ class Modal extends React.Component {
       this._element = null;
     }
 
-    // Use regex to prevent matching `modal-open` as part of a different class, e.g. `my-modal-opened`
-    const classes = document.body.className.replace(/(^| )modal-open( |$)/, ' ');
-    document.body.className = mapToCssModules(classNames(classes).trim(), this.props.cssModule);
-    setScrollbarWidth(this.originalBodyPadding);
+    if (!this.props.isOpen) {
```
fantua: Not a good check. In case that Modal never wasn't open before, it's still will removing class from `body`.

TheSharpieOne: Good catch. I ended up just resorting to keeping track of whether or not the instance added or removed the class. This way it will only ever add it if it isn't there and only remove it if it was added.

PR #833, opened 2018-02-13
docs: Update InputGroupAddon example and doc

saeedalipoor:


docs/lib/examples/InputGroupAddon.js
```diff
@@ -5,19 +5,27 @@ const Example = (props) => {
   return (
     <div>
       <InputGroup>
-        <InputGroupAddon addonType="prepend">To the Left!</InputGroupAddon>
+        <InputGroupAddon addonType="prepend">
+          <InputGroupText>To the Left!</InputGroupText>
```
fantua: It's can be used without `InputGroupText` and must have example in docs.

https://github.com/reactstrap/reactstrap/blob/24951541a110451c167ff53e6ebb367972717bc5/src/InputGroupAddon.js#L34-L41

Issue #831, opened 2018-02-13
update react-popper version


swarthy:
### Issue description
Warning due yarn install
```bash
warning reactstrap > react-popper@0.7.5: please use 0.7.4 or 8.0.0 and above
```

### What should be happening?
no warning?

### Steps to reproduce issue
yarn install --dev reactstrap

supergibbs:
Sounds like this is a `react-popper` issue with that version. https://github.com/souporserious/react-popper/issues/81

@TheSharpieOne reactstrap should probably pin to 0.7.4 for now and look into migrating to 0.8.x

virgofx:
We should migrate to 0.8.x as it fixes some other core issues with SSR. Using the current beta Reactstrap with dependencies is currently broken with react-popper until upgrade (Need to use external version @ 0.8.x).

PR #835, opened 2018-02-13
chore(react-popper): upgrade to ^0.8.2 (#831)

supergibbs:
Just a simple package update. No documented breaking changes from `react-popper` and all tests still pass! For #831


hkgsherlock:
the `^0.8.2` looks like being rewritten by someone else. I'm running projects that requires the `react-popper@^0.8.2` since it fixes the `export { ... }` not transformed by babel issue.

PR #840, opened 2018-02-15
Support valid feedback messages.

seberik:
Bootstrap supports form-validation success feedback messages: https://getbootstrap.com/docs/4.0/components/forms/

I added the possibility to provide a valid prop to the render method to add the `.feedback-valid` class instead of `.feedback-invalid`. It might be nicer if we would instead explicitly specify `invalid` or `valid={false}` but that would break the current behaviour. Any thoughts?

### DIFF
--- a/src/FormFeedback.js
+++ b/src/FormFeedback.js
@@ -27,8 +27,7 @@ const FormFeedback = (props) => {
 
   const classes = mapToCssModules(classNames(
     className,
-    !valid && 'invalid-feedback',
-    valid && 'valid-feedback'
+    valid ? 'valid-feedback' : 'invalid-feedback'
   ), cssModule);
 
   return (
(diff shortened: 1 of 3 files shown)


src/FormFeedback.js
```diff
@@ -8,23 +8,27 @@ const propTypes = {
...
     cssModule,
+    valid,
     tag: Tag,
     ...attributes
   } = props;
 
   const classes = mapToCssModules(classNames(
     className,
-    'invalid-feedback'
+    !valid && 'invalid-feedback',
```
TheSharpieOne: make this one line:
```
valid ? 'valid-feedback' : 'invalid-feedback'
```
or even something like
`${valid ? '' : 'in'}valid-feedback`

iKonrad:
Hi, when would this get updated in npm?

PR #842, opened 2018-02-19
Set col class only if no other classes are specified

### LINKED ISSUE #750: Suppress `col` className in `Col` component
### Issue description

- components: `Col`
- reactstrap version `#5.0.0-alpha.4`
- bootstrap version `#4.0.0-beta.2`

### What is happening?

Sometimes it's not desirable to have a column element having the class `col`. For example, one might want to reproduce the [Stacked to horizontal](https://getbootstrap.com/docs/4.0/layout/grid/#stacked-to-horizontal) example from Bootstrap's documentation.

In my specific case, I wan my columns to have only the class `col-md-4`. If I use `<Col md='4'>My content</Col>` the resulting HTML element will have the class `col col-md-4`. By looking into `Col`'s source code, I found out that I can get rid of the `col` class by setting `xs={false} md='4'`, but I believe that's not the correct way of doing it (right?).

I think it would be nice to have an "intended" way of suppressing the `col` class name.

pravdomil:
fixes #750

### DIFF
--- a/src/__tests__/Col.spec.js
+++ b/src/__tests__/Col.spec.js
@@ -35,18 +35,21 @@ describe('Col', () => {
     expect(wrapper.hasClass('col-wtf-1')).toBe(true);
     expect(wrapper.hasClass('order-wtf-2')).toBe(true);
     expect(wrapper.hasClass('offset-wtf-4')).toBe(true);
+    expect(wrapper.hasClass('col')).toBe(true);
   });
 
   it('should pass col size specific classes as Strings', () => {
     const wrapper = shallow(<Col sm="6" />);
 
     expect(wrapper.hasClass('col-sm-6')).toBe(true);
+    expect(wrapper.hasClass('col')).toBe(true);
   });
 
   it('should pass col size specific classes as Numbers', () => {
     const wrapper = shallow(<Col sm={6} />);
 
     expect(wrapper.hasClass('col-sm-6')).toBe(true);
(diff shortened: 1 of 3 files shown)


src/__tests__/Col.spec.js
```diff
@@ -35,21 +35,18 @@ describe('Col', () => {
     expect(wrapper.hasClass('col-wtf-1')).toBe(true);
     expect(wrapper.hasClass('order-wtf-2')).toBe(true);
     expect(wrapper.hasClass('offset-wtf-4')).toBe(true);
-    expect(wrapper.hasClass('col')).toBe(true);
```
TheSharpieOne: I would explicitly check for it to not have the class (change `true` to `false`) as with the test as it is now, `col` could be added back and nothing would break.

pravdomil: reverted and changed to `true`

pravdomil:
> Can you add a test for the case when no columns are specified, ensuring the class col is added.
Also, ensure that the col class is not added in the other tests.

https://github.com/pravdomil/reactstrap/blob/962695c77335d33f5567b0416a3cdc14885643f8/src/__tests__/Col.spec.js#L6

pravdomil:
@TheSharpieOne when you are planning to release next version?

TheSharpieOne:
released 5.0.0-beta.2 via `@next` tag

PR #887, opened 2018-03-06
fix(build): inform rollup of react-popper exports

TheSharpieOne:
This brings react-popper back to ^0.8.2, but adds some rollup config to allow the build to deal with how react-popper is defining the exports.

CC @supergibbs

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "lodash.isobject": "^3.0.2",
     "lodash.tonumber": "^4.0.3",
     "prop-types": "^15.5.8",
-    "react-popper": "0.8.1",
+    "react-popper": "^0.8.2",
     "react-portal": "^4.1.2",
     "react-transition-group": "^2.2.1"
   },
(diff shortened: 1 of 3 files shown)


virgofx:
@TheSharpieOne  can you zip up the /dist contents and attach those  and I'll test the external mechanism as outlined in that other Issue to make sure this works.

TheSharpieOne:
@virgofx https://gist.github.com/TheSharpieOne/c329933e65ad9e4821c1784a686ea40c

virgofx:
:+1:  will test later tonight (hopefully if time permits)

TheSharpieOne:
@virgofx so what is the fix for `TypeError: Cannot use 'in' operator to search for 'default' in undefined`? Do we just add another edit to the postbuild script to change
`var ReactDOM__default = 'default' in ReactDOM ? ReactDOM['default'] : ReactDOM;` to `var ReactDOM__default = ReactDOM && 'default' in ReactDOM ? ReactDOM['default'] : ReactDOM;`

virgofx:
Let me look at it tomorrow. We could do that ... I'd like to know why it changed. Not sure if it's a rollup change , or because of the way we're using one of our dependencies.

TheSharpieOne:
This is not longer needed.

PR #926, opened 2018-03-26
docs(Readme): Update Readme with latest stable Reactstrap versions.

virgofx:
Intent is to ensure boilerplate example is more up-to-date which is useful if people use exact external versions in keeping latest versions hot in cache.


virgofx:
Thanks!

Issue #871, opened 2018-02-27
Input elements shouldn't have children


jhsware:
https://github.com/reactstrap/reactstrap/blob/0648cc2c24a5dd55f93892df9c6c9d144338f670/src/__tests__/Input.spec.js#L60

TheSharpieOne:
They should when the "input" is a `<select>` or  `<textarea>`, which the `Input` can render given `type="select"` or `type="textarea"` respectfully.

jhsware:
The default input type is `text` so the test should be something like this: `<Input type="select">Yo!</Input>`

Are you sure `type="textarea"` is allowed children? The React-clone Inferno.js complains if textarea has children.

TheSharpieOne:
Sounds like a defect for inferno. `<textarea>` is not a void element, it allows content within it. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

Yes, the test itself probably could use `type="textarea"`, but we don't check the type to determine if it can have children (we probably should, it would be simple since we use the type to determine the `tag` and if the `tag` is input, we should not allow children).

jhsware:
React doesn't follow HTML-specs in that regard: https://reactjs.org/docs/forms.html

```
<textarea value={this.state.value} onChange={this.handleChange} />
```

Only `<select>` should have children from what I can tell.

TheSharpieOne:
Yeah, looks like they are removing the ability for `<textarea>`s to have children in react 17: https://github.com/facebook/react/pull/11639

jhsware:
I have made a port of Reactstrap to Inferno.js so I get stuck in the middle at times... :D

PR #927, opened 2018-03-26
 chore(Input): remove support for children content in <textarea/>

gergely-nagy:
according to #871

### DIFF
--- a/src/Input.js
+++ b/src/Input.js
@@ -93,8 +93,9 @@ class Input extends React.Component {
       attributes.type = type;
     }
 
-    if (type === 'textarea') {
-      attributes.children = null;
+    if (attributes.children && !(plaintext || staticInput || type === 'select' || typeof tag !== 'string' || tag === 'select')) {
+      warnOnce(`Input with a type of "${type}" cannot have children. Please use "value"/"defaultValue" instead.`);
+      delete attributes.children;
     }
 
     return (
--- a/src/Input.js
+++ b/src/Input.js
@@ -93,6 +93,10 @@ class Input extends React.Component {
       attributes.type = type;
     }
 
+    if (type === 'textarea') {
+      attributes.children = null;
+    }
+
     return (
       <Tag {...attributes} ref={innerRef} className={classes} />
     );
--- a/src/Input.js
+++ b/src/Input.js
@@ -93,7 +93,7 @@ class Input extends React.Component {
       attributes.type = type;
     }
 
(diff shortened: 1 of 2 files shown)


src/__tests__/Input.spec.js
```diff
@@ -63,6 +63,12 @@ describe('Input', () => {
     expect(wrapper.text()).toBe('Yo!');
   });
 
+  it('should render without children when type is "textarea"', () => {
+    const wrapper = shallow(<Input type="textarea">Yo!</Input>);
+
+    expect(wrapper.text()).toBe('');
```
virgofx: Shouldn't the test be checking to ensure that 'Yo!' is included inside the textarea. This test doesn't cover that.

src/Input.js
```diff
@@ -93,6 +93,10 @@ class Input extends React.Component {
       attributes.type = type;
     }
 
+    if (type === 'textarea') {
```
TheSharpieOne: I think we may want to do something a little more restrictive. It looks like the only input which allows children is `<select>` (and `plaintext`/`staticInput`). Maybe we should do something like
```js
if (!(plaintext || staticInput || type === 'select' || typeof tag !== 'string')) {
  attributes.children = null;
}
```
After though, I also added `typeof tag !== 'string'` as if the user is passing another component to render (via the `tag` prop), we should just let them do what they want as they may render a `<select>` or something else which allows children. (similar use-case as it being in the `if` condition above)

src/__tests__/Input.spec.js
TheSharpieOne: Currently, that is how it will work (and react will spit out a warning), but in react 17 it will not work. React wants people to use `value`/`defaultValue` even for `<textarea>`, making `<textarea />` a void element.
This becomes a breaking change for reactstrap now, ahead of the breaking change react 17 will force us to make. It simplifies things a little it makes `<textarea>` and `<input>` basically the same.... which IMHO, HTML should have done `<input type="textarea" />` (I wonder why they didn't... I'll have to Google)

src/Input.js
TheSharpieOne: I think it should be `typeof tag !== 'string'` since we negate the entire condition.
For instance, if `typeof tag` is "function", then `typeof tag !== 'string'` will be `true`, which would get negated to become `false` which skip the code block and keep children.

It can be rewritten to make it more clear:
```js
if (!plaintext && !staticInput && type !== 'select' && typeof tag === 'string') {
  attributes.children = null;
}
```

PR #929, opened 2018-03-27
pass Popper modifiers to DropdownMenu

borm:


src/DropdownMenu.js
```diff
@@ -53,7 +55,10 @@ const DropdownMenu = (props, context) => {
     const position2 = right ? 'end' : 'start';
     attrs.placement = `${position1}-${position2}`;
     attrs.component = tag;
-    attrs.modifiers = !flip ? noFlipModifier : undefined;
+    attrs.modifiers = !flip ? {
+      ...noFlipModifier,
+      ...modifiers,
```
virgofx: If we want no-flip ... should we reverse the order here? That way if one specifies `flip={false} modifiers={{flip:true}}` flip will win out on the modifiers?  Rare case, just trying to think which should win?

virgofx:
Looks like one of the tests broke as well. Will need an additional test as well to test different modifier situations ... and handle the expected case for: `flip={false} modifiers={{flip:{enabled:true}}}`

borm:
```flip={false} modifiers={{flip:{enabled:true}}}```

reverse the order should fix this case

File: src/DropdownMenu.js [hunk 2]
```diff
@@ -9,13 +9,15 @@ const propTypes = {
...
   flip: PropTypes.bool,
+  modifiers: PropTypes.object,
   className: PropTypes.string,
   cssModule: PropTypes.object,
 };
 
 const defaultProps = {
   tag: 'div',
   flip: true,
+  modifiers: {},
```
TheSharpieOne: Don't need a default, undefined is fine, even with the spread operator.

src/__tests__/DropdownMenu.spec.js
```diff
@@ -157,7 +157,7 @@ describe('DropdownMenu', () => {
       }
     );
 
-    expect(wrapper.find(Popper).prop('modifiers')).toBe(undefined);
+    expect(wrapper.find(Popper).prop('modifiers')).toEqual({});
```
TheSharpieOne: Without the default prop setting it to and object, this should be able to be reverted

borm:
i don't know why, but i can't start project on local env, something wrong with `CopyWebpackPlugin` i think, or maybe anything else
```
borms-iMac:reactstrap borm$ node -v
v4.6.2
borms-iMac:reactstrap borm$ npm -v
2.15.11

```
image

TheSharpieOne:
localhost vs 127.0.0.1? IIRC webpack dev server checks the host header and will not serve content if it doesn't match the settings.

borm:
localhost also not worked

docs/lib/Components/DropdownsPage.js
```diff
@@ -291,6 +293,86 @@ DropdownItem.propTypes = {
...
   </DropdownMenu>
+</Dropdown>`}
+          </PrismCode>
+        </pre>
+
+        <h3>Modifiers</h3>
+        <div className="docs-example">
+          <Row>
+            <Col>
+              <Dropdown>
```
TheSharpieOne: You need `isOpen` and `toggle` props, or you can use `UncontrolledDropdown` here (and just keep `Dropdown` with some dummy `isOpen` and `toggle` props in the raw example code which is displayed)

borm: omg, okay)

PR #937, opened 2018-04-02
turn zindex reactive

lslv1243:
i was trying to set zindex on state update relative to the amount of modals opened and i noticed that it was not updating in the dom.


src/Modal.js
```diff
@@ -120,6 +120,10 @@ class Modal extends React.Component {
     if (this.props.autoFocus && this.state.isOpen && !prevState.isOpen) {
       this.setFocus();
     }
+
+    if (prevProps.zIndex !== this.props.zIndex) {
```
TheSharpieOne: Probably also want to check to see if `this._element` exists. Also, it would be nice to add a test to ensure this functionality isn't lost in future changes.

lslv1243: did what you asked

Issue #942, opened 2018-04-05
Docs - Search 


gergely-nagy:
We should support search feature in the documentation. [React ](https://reactjs.org/) and [Bootstrap ](https://getbootstrap.com/docs/4.0/getting-started/introduction/) docs is using [Algolia ](https://community.algolia.com/docsearch/) for search. If the the owner of the site request a key from algolia, I can work on this.

TheSharpieOne:
I sent a request
> We'll review your request shortly, after making sure your website is eligible to DocSearch.
This is a manual process and can take some time.
We'll then email you the code you need to integrate on your website

TheSharpieOne:
> If you add id attribute to your "lvlX" elements, we could get the anchor and thus scroll down to the right position thanks to the URL.
>
> Your .nav-link "Utilities" element redirect to a "404 Page Not Found" page contrary to "Components", it will not be crawled. It will not be scraped either since the .docSearch-content class is missing.
>
> Please PR your config once you have made these changes, I will help you fine tuning your DocSearch implementation.
>
> In the meanwhile, congratulations, your search is now ready! 
>
> I've successfully configured the underlying crawler and it will now run every 24h. 
>
> You're now a few steps away from having it working on your website: 
>
> - Copy the following CSS/JS snippets and add them to your page 
```html
<!-- at the end of the HEAD --> 
<link rel="stylesheet", href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" /> 

<!-- at the end of the BODY --> 
<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script> 
<script type="text/javascript">  docsearch({ 
 apiKey: '2999e8cb716a50f4bd09a3addb18bf4f', 
 indexName: 'reactstrap', 
 inputSelector: '### REPLACE ME ####', 
 debug: false // Set debug to true if you want to inspect the dropdown 
}); 
</script>
```
>
> - Add a search input in your page if you don't have any yet. Then update the inputSelector value in JS snippet to a CSS selector that targets your search input field. 
>
> - Optionally customize the look and feel by following the DocSearch documentation (https://community.algolia.com/docsearch/documentation/) 
> - You can also check your configuration in our github repo (https://github.com/algolia/docsearch-configs/blob/master/configs/reactstrap.json). 
>
> Please open a pull request if want to leverage your configuration! 
>
[...]

PR #950, opened 2018-04-12
feat(docs): Add search for documentation

gergely-nagy:
according to #942

### DIFF
--- a/docs/lib/app.js
+++ b/docs/lib/app.js
@@ -65,6 +65,7 @@ export default (locals, callback) => {
           <link rel=icon href=/assets/favicon.ico>
           <link rel="stylesheet" href="/assets/style.css"/>
           <link rel="stylesheet" href="/assets/docs.css"/>
+          <link rel="stylesheet", href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
         </head>
         <body>
           <div id="app">${body}</div>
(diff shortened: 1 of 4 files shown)


docs/lib/app.js
```diff
@@ -65,6 +65,7 @@ export default (locals, callback) => {
           <link rel=icon href=/assets/favicon.ico>
           <link rel="stylesheet" href="/assets/style.css"/>
           <link rel="stylesheet" href="/assets/docs.css"/>
+          <link rel="stylesheet", href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
```
virgofx: Please update end point to use CDNJS 

https://cdnjs.com/libraries/docsearch.js

docs/lib/app.js
```diff
@@ -77,6 +78,15 @@ export default (locals, callback) => {
           </script>
           <script src="/assets/prism.js" data-manual></script>
           <script src="/bundle.js"></script>
+          <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
```
virgofx: Please update endpoint to use CDNJS

Please update end point to use CDNJS 

https://cdnjs.com/libraries/docsearch.js

docs/lib/app.js
TheSharpieOne: The one used is what Algolia (the search provider) provided. It has a feature which cdnjs does not, it can use semver to provide the latest version.
It has some pluses and minuses, such as since the version and this the content can change, we cannot use SRI.

virgofx: Both use Cloudflare as a backend and I would argue that CDNJS has best global performance and features (aside from ~latest semver) which has been deprecated by most major CDNs (google, CDNJS, etc) because it's bad practice and I realize developer just pointed link to JSDELIVR ... I'd still say it's okay using CDNJS.

Either way more of a recommendation to be consistent with our CDN offerings for main reactstrap.

TheSharpieOne: It's about cache hits and misses. Using what they recommend gives the best chance of a cache hit.
In the end, it really doesn't matter. I just want to get this feature added to the docs.

PR #969, opened 2018-04-20
add reference to Data Grid built on top of Reactstrap

kvet:


TheSharpieOne:
The only place we do this is in the README "In the wild" section. If you would like to add it there, that would be permitted.

Issue #960, opened 2018-04-17
reactstrap with react-router-bootstrap


FutureKode:
I'm upgrading a header component from react-bootstrap to reactstrap, that's all fine, only thing is that the dropdown toggles are no longer active when a dropdown menu item is active. The previous code was like this:

```
<NavDropdown eventKey={3} title="Topology" id="topology-dropdown">
                <LinkContainer to="/topology-search">
                  <MenuItem>Search</MenuItem>
                </LinkContainer>
                <LinkContainer to="/model-query">
                  <MenuItem>Model</MenuItem>
                </LinkContainer>
              </NavDropdown>
```

The equivalent in `reactstrap`:

```
<UncontrolledDropdown nav inNavbar>
                <DropdownToggle nav caret>
                  Topology
                </DropdownToggle>
                <DropdownMenu>
                  <DropdownItem
                    tag={Link}
                    to="/topology-search"
                    activeClassName="active"
                  >
                    Search
                  </DropdownItem>
                  <DropdownItem
                    tag={Link}
                    to="/model-query"
                    activeClassName="active"
                  >
                    Model
                  </DropdownItem>
                </DropdownMenu>
              </UncontrolledDropdown>
```

Can I still use react-router-bootstrap with reactstrap?

TheSharpieOne:
> Can I still use react-router-bootstrap with reactstrap?

No, react-router-bootstrap solves an issue for react-bootstrap which reactstrap solved with the `tag` prop.

As for the issue of "the dropdown toggles are no longer active when a dropdown menu item is active", it looks like that is a feature built into react-bootstrap which reactstrap is lacking, but bootstrap itself doesn't even have that feature (which is why reactstrap didn't implement it).

It can definitely be added via a prop toggle (like `inheritActive` or something) which would look to see if any of the children are active and add the active class. That way it would be optionally used and default to how bootstrap is.

FutureKode:
thanks for this great library and helpful answer @TheSharpieOne 

I would like to contribute and take this on. I have forked the repo and will create a new branch according to the `contributing` docs in the readme. Should it be a patch version or a minor version?

TheSharpieOne:
[...]

PR #977, opened 2018-04-24
feat(Dropdown): setActiveFromChild prop

FutureKode:
See issue: #960


TheSharpieOne:
I'm not 100% sure this will work with the use-case in #960. Using react-router's `Link` with it's `activeClassName` means that the `active` prop will not get set, only the "active" class if and when it is active and even that is only evaluated afterwards.
Looking at how react-bootstrap created this feature, they are just matching routes directly in their code to determine if the route is active rather than using anything react-router would give them.
So while this would work, it would not be able to leverage react-router to determine if the link is active, leaving it up to the user to add that logic in the `active` prop.

FutureKode:
Yes you are right :( it doesn't work when using react-router `Link`

TheSharpieOne:
Still a nice feature. The kinks for the Link integration can be worked out later. For now, people can manage the active state of the dropdown via the items's active props rather than having additional logic in their code.

apaatsio:
There's zero documentation for `setActiveFromChild`. It would be nice to always update the documentation when new features are added.

apaatsio:
I don't think the property name is enough documentation in this case. A simple usage example would be clarifying enough, though.

FutureKode:
@apaatsio good idea. I have created a [PR](https://github.com/reactstrap/reactstrap/pull/1014), let me know if it is okay :)

PR #992, opened 2018-05-01
ship untranspiled source

pravdomil:
see http://2ality.com/2017/06/pkg-esnext.html#package-authors

this PR enables ship untranspiled source of reactstrap installed from github via `npm i reactstrap/reactstrap` so that you can transpile to target you want

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "README.md",
     "CHANGELOG.md",
     "lib",
-    "dist"
+    "dist",
+    "src"
   ],
   "keywords": [
     "reactstrap",
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "jsdelivr": "dist/reactstrap.min.js",
   "unpkg": "dist/reactstrap.min.js",
   "cdn": "dist/reactstrap.min.js",
+  "esnext": "src",
   "scripts": {
     "report-coverage": "coveralls < ./coverage/lcov.info",
     "test": "cross-env BABEL_ENV=test react-scripts test --env=jsdom",


TheSharpieOne:
You need to also include the src directory in the files which get published to npm.

pravdomil:
yea, true, do not know how

TheSharpieOne:
add `"src"` to the package.json `"files"` array.

Also, I was reading the comment for that article you linked, the `esnext` and "raw" source has some issues as it assumes the raw source doesn't require special transforms or plugins to be parsed or compiled. I'm fine with making this change, but I don't know if that package.json `esnext` property is going to take off and as soon as there is an issue, I'll probably drop it (but probably keep shipping the raw; just removing the reference)

pravdomil:
src folder added

> the esnext and "raw" source has some issues as it assumes the raw source doesn't require special transforms or plugins to be parsed or compiled

they mean if you the package depends on some non-standard babel transform eg. something in stage-0 or custom babel plugin, then it will fail to compile since a developer himself handle babel configuration, I have check what `reactstrap` uses and it's only ES6 transform and [transform-object-rest-spread](https://github.com/tc39/proposal-object-rest-spread) which is now in stage-4 so that should be fine. Everything which is in stage-4 is considered as `esnext`. BTW: react is the case.

> I'm fine with making this change, but I don't know if that package.json esnext property is going to take off and as soon as there is an issue, I'll probably drop it (but probably keep shipping the raw; just removing the reference)

agree, this is more like a proof of concept, I'm curious how it will work

PR #994, opened 2018-05-02
Fix warning about invalid arguments to oneOfType

jasco:
Fixes console warning:

    Invalid argument supplied to oneOfType, expected an instance of array


TheSharpieOne:
Thanks!

Issue #334, opened 2017-02-17
UncontrolledAccordion? 


AlJohri:
Would be great to have a version of UncontrolledCollapse or perhaps more specifically an UncontrolledAccordion. Especially one that does something similar to bootstrap's `data-parent`:

>To add accordion-like group management to a collapsible control, add the data attribute data-parent="#selector". Refer to the demo to see this in action.

https://v4-alpha.getbootstrap.com/components/collapse/#accordion-example

Right now I'll have to handle the mutually exclusive state logic myself.

threehams:
`recompose` is extremely useful for this, once you wrap your head around the API.

For instance, for an uncontrolled tooltip:
```
const StatefulTooltip = compose(
  withState("isOpen", "setOpen", false),
  withHandlers({
    toggle: ({ isOpen, setOpen }) => () => setOpen(!isOpen)
  }),
  mapProps(({ setOpen, ...rest }) => rest),
)(Tooltip);
```

(I know Reactstrap provides an uncontrolled tooltip, but this pattern can apply to any controlled component).

TheSharpieOne:
Accordion is essentially multiple Collapse components working together to ensure only one is open at a time. Not 100% sure how to make it uncontrolled in a similar way as the other uncontrolled component even with recompose in the mix.
An Accordion component may be needed to act as the controller on the children collapse.
BS4 uses data-parent to indicate that a collapse is apart of the accordion. This helps nested collapse components within the accordion not be apart of the accordion. Interesting things happen: https://codepen.io/TheSharpieOne/pen/qXdqbb so we cannot simply say any Collapse under an Accordion is part of the Accordion.

foodgy:
I'm just curious why there is no UncontrolledCollapse?

TheSharpieOne:
@foodgy The trigger event and element for the collapse is not directly tied to the collapse itself. Unlike the other Uncontrolled components, collapse can be triggered by different events from different elements. For instance, is the trigger element a button or a card header? Is the collapseable content adjacent, above or below the trigger? All of the variation make it more difficult to provide a single component to do it all easily, so the logic is left up to the developer.

We could have something like an UncontrolledCollapseCard, which looks like a Card where the CardHeader is the trigger. Multiple of these tied together with a little more logic could be an UncontrolledAccordion.

sscaff1:
You can create an UncontrolledCollapse with a HOF. No other dependencies necessary.

```js
import React, { Fragment, Component } from 'react';
[...]

PR #1009, opened 2018-05-07
UncontrolledCollapse with tests and doc updates closes 334

sscaff1:
This should close out #334. Let me know if there is anything I missed.


src/utils.js
```diff
@@ -129,23 +131,24 @@ export function getTarget(target) {
...
-
 /* eslint key-spacing: ["error", { afterColon: true, align: "value" }] */
 // These are all setup to match what is in the bootstrap _variables.scss
 // https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss
 export const TransitionTimeouts = {
   Fade:     150, // $transition-fade
   Collapse: 350, // $transition-collapse
   Modal:    300, // $modal-transition
-  Carousel: 600, // $carousel-transition
+  Carousel: 600 // $carousel-transition
```
virgofx: Project uses trailing commas. Curious as to why these were removed?  x 4 below.

sscaff1: Hmm interesting. My personal prettier setting are "prettier.trailingComma": "es5", so not sure why they were removed. I'll go back and do it though.

sscaff1: The problem is 'comma-dangle': 0, in the eslintrc file. So prettier did this automatically for me. Do you want me to change the eslint rule?

sscaff1: So eslint is autoformating for me because the rule 'comma-dangle': 0 is in the eslintrc file. Do you want me to change it. I change it to 2 and it puts the dangling comma. I could also turn off prettier and just manually put the dangling comma. Let me know. @virgofx

virgofx: Interesting ... I don't know the answer. Not sure on the history of the eslint line ... versus how the rest of the project was setup. I believe initially it was setup with comma-dangle. Will need more input from eddy or @TheSharpieOne

TheSharpieOne: The [eslint comma dangle rule is off](https://github.com/reactstrap/reactstrap/blob/b68826d81e1d24f2c26f4a725a3652d519a43dd5/.eslintrc.js#L21), meaning it shouldn't affect the project one way or the other. We should enforce them being there as it leads to cleaner code diffs.
Yes, can you please add prettier. Also, if you can, clean up the eslint to allow comma dangle, enforce single quotes (double quotes for jsx props) and remove anything that is no longer needed or conflicts with prettier.

package.json
```diff
@@ -86,7 +81,8 @@
     "edgji (https://github.com/edgji)",
     "nlrowe (https://github.com/nlrowe)",
     "npm-to-cdn-bot (by Forbes Lindesay) (https://github.com/npmcdn-to-unpkg-bot)",
-    "polmauri (https://github.com/polmauri)"
+    "polmauri (https://github.com/polmauri)",
+    "Steven Scaffidi (https://github.com/sscaff1)"
```
TheSharpieOne: These were originally generated (AFAIK). @eddywashere do you recall what you used to generate this list?

src/utils.js
sscaff1: It's very strange, I'm not sure why prettier takes the comma away because I thought the above as well. I'll clean up this PR by tonight (probably just after work) and then introduce another PR with prettier-eslint integrated. Thanks for the feedback @TheSharpieOne

src/utils.js
```diff
@@ -197,11 +200,32 @@ export const PopperPlacements = [
...
+  'left-start'
 ];
 
 export const canUseDOM = !!(
   typeof window !== 'undefined' &&
   window.document &&
   window.document.createElement
 );
+
+export function findDOMElements(el) {
```
TheSharpieOne: This is similar to `getTarget` above with the differences being this only accepts a string to pass to `querySelectorAll`, it returns a list of elements, and it more SSR save.
Can this be refactored to allow for different types of arguments (such as what `getTarget` allows: function, string, raw DOM reference). Then can `getTarget` be refactored to use this can just grab the first element in the list.

src/utils.js
```diff
@@ -197,11 +200,32 @@ export const PopperPlacements = [
...
+  return canUseDOM && window.document.querySelectorAll(el);
+}
+
+export const defaultToggleEvents = ['touchstart', 'click'];
+
+export function addMultipleEventListeners(els, handler, events) {
+  events.forEach((event) => {
```
TheSharpieOne: Can you ensure events is an array before using it, it would be nice to make this works for single events which may be applied to multiple targets.

src/utils.js
```diff
@@ -197,11 +200,32 @@ export const PopperPlacements = [
...
+    els.forEach((el) => {
```
TheSharpieOne: Can you ensure e`els` is an array before using it, it would be nice to make this works for multiple events which may be applied to a single targets.

sscaff1:
@TheSharpieOne I tried to address all of your comments from your review. I squashed the two commits just for the sake of keeping the git history clean (I didn't really find anything on this in terms of the contributing guidelines). A couple of issues could probably come out of this:

1. Update tooltip and other components to use addMultipleEventListeners (or a similar pattern).
2. Integrate prettier-eslint as mentioned above. I mean really this is your choice but you thumbs up it so I thought I would mention it. 😄 
3. I was also thinking you could probably drop the dependency on lodash as well. There are only a couple of instances here and there of it which could easily be removed.

Anyway, let me know what you think. I'm happy to work on one or more of the above mentioned issues.

Issue #1012, opened 2018-05-09
Tooltip Accessibility


TheSharpieOne:
Based on https://www.w3.org/TR/wai-aria-practices/#tooltip it should

- never receives focus. Focus stays on the textbox.
- be shown via keyboard focus or by the onmouse event.
- be hidden by removing focus from the text box or by moving the mouse off the textbox.
- be hidden by pressing the Escape key (bootstrap currently doesn't do this)

Also,
- The element that serves as the tooltip container has role tooltip.
- The element that triggers the tooltip references the tooltip element with aria-describedby. (not sure if this is realistic or possible, we probably depend on the developer using the component to add this, but it would be nice if it happened automatically while keeping the current API for tooltips which allows them to be disjoint from their trigger)

TheSharpieOne:
Not 100% sure if `Popover` is more like a tooltip of a dialog/modal of sorts. We definitely don't want mouse events other than click opening/closing it...

PR #1025, opened 2018-05-17
feat(Tooltip): add tooltip accessibility #1012

RockinRonE:
This adds shows a tooltip on focus and hides it when out of focus. Also adds the ability to close tooltip using the escape key. Adds `role='tooltip'` to the tooltip.


src/Tooltip.js
```diff
@@ -156,6 +168,11 @@ class Tooltip extends React.Component {
   addTargetEvents() {
     this._target.addEventListener('mouseover', this.onMouseOverTooltip, true);
     this._target.addEventListener('mouseout', this.onMouseLeaveTooltip, true);
+    this._target.addEventListener('keydown', this.onEscKeyDown, true);
+    this._target.addEventListener('focusin', this.handleFocus, true);
```
TheSharpieOne: Because this calls the same method as `focusout` which just blindly toggles the tooltip, it can get into a bad state where `focusin` hides the toggle and `focusout` shows it. We probably want to explicitly show on `focusin` and hide on `focusout`
tooltip-focus
 
Also, this is more evident when using `esc` to close the tooltip and then tabbing away.
tooltip-esc

RockinRonE:
@TheSharpieOne Thanks for the explanation!

RockinRonE:
@TheSharpieOne Thanks! This is my first code contribution to open source. I encourage you to critique my PRs as I want to improve. Mind tagging me when you change the handles so I can see what you did?

TheSharpieOne:
@RockinRonE 98dfa034969

PR #1035, opened 2018-05-25
Added src for Typescirpt support

Domino987:


src/Media.js
```diff
@@ -31,8 +32,9 @@ const Media = (props) => {
     middle,
     object,
     right,
+    src,
     tag,
-    top,
```
TheSharpieOne: The comma is needed.

PR #1036, opened 2018-05-25
add src for Media in typescript

Domino987:


src/Media.js
```diff
@@ -31,6 +32,7 @@ const Media = (props) => {
     middle,
     object,
     right,
+    src,
```
TheSharpieOne: we keep `src` in the `attributes` since we access it from there below and so that it automatically gets added to the rendered element since we spread `attributes`

TheSharpieOne:
no need to close the PR when you make a change, just push up the new commit(s)

Domino987:
I see , good to know.

src/Media.js
Domino987: So should I delete the top or bottom one xD

Domino987: So hier exactly do you need me to add src to the Media component, so that I can use it in typescript?

TheSharpieOne: none of these files control typescript, that is all handled in a separate repository by other people.

Domino987: Oh that's right , sorry for that

Issue #1038, opened 2018-05-27
Input Element of type="select" leaks the type attribute


virgofx:
The **Input** element leaks the `type` attribute when type Select. By default the `type` attribute is included because it's utilized for the various HTML `input` types (e.g. `file`, `input`, `text`, etc.) so it's useful; however, this component also uses the type attribute to render alternate tags (e.g. for `type="select"`) and having the type attribute leak breaksthe HTML5 spec as it does not allow for the `type` attribute for the `<select />` element. [w3c reference](http://w3c.github.io/html/single-page.html#the-select-element)

![image](https://user-images.githubusercontent.com/739719/40581941-13a9db38-611b-11e8-9be8-f4776c7c389e.png)

**Input**
```jsx
return (
    <Input type="select">
        <option value="1">Test</option>
    </Input>
);
```

**Output**
```html
<select name="test" type="select" class="form-control">
  <option value="1">Test</option>
</select>
```

- components: `Input`
- reactstrap version `6.0.1`
- import method `umd`
- react version `6.4.0`
- bootstrap version `4.1.1`

---

The issue is because of the following code. `type` is already plucked from attributes and we add it back only for type input (makes sense) .... but were doing a second check on the type of tag which will be undefined for most defaults and thus will get evaluated in the block because `typeof undefined` = `'undefined'` `!==` `'string'`

**Input.js**
```jsx
    if (Tag === 'input' || typeof tag !== 'string') {
      attributes.type = type;
    }
```

Suggesting changing to:
```jsx
    if (Tag === 'input' || typeof tag !== undefined) {
      attributes.type = type;
    }
```

Feedback wanted ...

gergely-nagy:
Nice catch. The `<textarea />` tag has the same issue.

hermawan22:
Hi, may i work on it?
It will be first contribution for reactstrap.

What i think to fix it : 
` if (Tag === 'input' || tag) {
    attributes.type = type;
}`

I think we don't need to check if typeof tag is string or not, coz i think tag will only read string as tag element and function to return new element.

TheSharpieOne:
We should probably check that `tag` is a function. The main reason is to account for someone using the static input (which normal renders a `<p>`) and wanted to change the DOM element to something else. In that case without checking the type of tag, it would add type which is unwanted.

hermawan22:
HI @TheSharpieOne thank u very much for suggestion and explanation.

[...]

PR #1041, opened 2018-05-28
fix(Input):  type attribute should not render for select and textarea

hermawan22:
## What i change : 
I change code on Input component in `src/Input.js`
And added test case for it.


hermawan22:
This PR to fix this issue #1038

Issue #1034, opened 2018-05-25
UncontrolledCollapse issue with tapping


locpd:
### Issue description

- components: `UncontrolledCollapse`
- reactstrap version `master`
- import method `es6` 
- react version `#16.2.0`
- bootstrap version `#4.0.0`

### What is happening?
In Chrome browser on mobile (or in Device Mode in Chrome on desktop), when tapping on the Button to trigger the toggle, the UncontrolledCollapse component quickly opens and closes. Looks like the onClick event handler has been bound twice.

### What should be happening?
The UncontrolledCollapse component should just open or close after tapping toggle Button.

### Steps to reproduce issue

1. Open https://reactstrap.github.io/components/collapse/ in Chrome on mobile or use "Toggle device toolbar" in Chrome on desktop.
2. In the "Uncontrolled Collapse" section, tap the "Toggle" Button.

PR #1043, opened 2018-05-28
fix(UncontrolledCollapse): collapsing on mobile (#1034)

illiteratewriter:
fixed bug where uncontrolled collapse was not collapsing in mobile


illiteratewriter:
The issue was both `touch` handler and `click` handler was being called, causing it to toggle twice, and make it initial state itself. This now prevents `click` on mobile.

PR #1056, opened 2018-06-04
fix(Util): Modify to use CSS escaped string when 'taget' is ID

skyoo2003:
When trying to pass a specific DOM id attribute to the target property of the 'Tooltip' component, an exception is thrown when the DOM id is an invalid selector so I modified lines for making 'target' to be CSS escaped.

---

When I run my test cases that uses the invalid selector, I get the following logs.

```
SyntaxError: '[0].etcdName-Tooltip' is not a valid selector

      at emit (node_modules/nwsapi/src/nwsapi.js:552:17)
      at _querySelectorAll (node_modules/nwsapi/src/nwsapi.js:1371:9)
      at Object._querySelector [as first] (node_modules/nwsapi/src/nwsapi.js:1310:14)
      at DocumentImpl.<anonymous> (node_modules/jsdom/lib/jsdom/living/nodes/ParentNode-impl.js:65:42)
      at DocumentImpl.querySelector (node_modules/jsdom/lib/jsdom/utils.js:123:45)
      at Document.querySelector (node_modules/jsdom/lib/jsdom/living/generated/Document.js:689:45)
      at getTarget (node_modules/reactstrap/src/utils.js:127:30)
      at Tooltip.componentDidMount (node_modules/reactstrap/src/Tooltip.js:67:20)
      at commitLifeCycles (node_modules/react-dom/cjs/react-dom.development.js:14248:22)
      at commitAllLifeCycles (node_modules/react-dom/cjs/react-dom.development.js:15342:7)
```

```
The target 'config.masterHosts-Tooltip' could not be identified in the dom, tip: check spelling

      at findDOMElements (node_modules/reactstrap/src/utils.js:210:13)
      at getTarget (node_modules/reactstrap/src/utils.js:224:15)
      at Tooltip.componentDidMount (node_modules/reactstrap/src/Tooltip.js:72:20)
      at commitLifeCycles (node_modules/react-dom/cjs/react-dom.development.js:14248:22)
      at commitAllLifeCycles (node_modules/react-dom/cjs/react-dom.development.js:15342:7)
      at HTMLUnknownElement.callCallback (node_modules/react-dom/cjs/react-dom.development.js:100:14)
...
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "lodash.isfunction": "^3.0.9",
     "lodash.isobject": "^3.0.2",
     "lodash.tonumber": "^4.0.3",
+    "css.escape": "^1.5.1",
     "prop-types": "^15.5.8",
     "react-popper": "^0.10.4",
     "react-transition-group": "^2.3.1"
--- a/src/__tests__/utils.spec.js
+++ b/src/__tests__/utils.spec.js
@@ -136,6 +136,16 @@ describe('Utils', () => {
       document.querySelectorAll.mockRestore();
     });
 
+    it('should query the document for the css escaped id target if the target is a string and could not be found normally', () => {
+      const element = document.createElement('div');
+      element.setAttribute('id', '[0].thing');
+      document.body.appendChild(element);
+      jest.spyOn(document, 'querySelectorAll');
+      expect(Utils.getTarget('[0].thing')).toEqual(element);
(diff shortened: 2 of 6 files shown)


virgofx:
This adds an unnecessary dependency for a very small edge case. The existing error is sufficient in my opinion. I don't think should get merged.

TheSharpieOne:
This should be done in userland, as in the user of this library should include whatever libraries they need to create the correct string to pass to reactstrap.

Issue #1062, opened 2018-06-06
Form validation with tooltips


Abisar:
### Issue description

- components: `Form` `FormFeedback`
- reactstrap version `#5.0.0`
- bootstrap version `#4.1.1`

### What is happening?

I can't find examples of the tooltips for form validation that are documented here: [Tooltips](https://getbootstrap.com/docs/4.1/components/forms/#tooltips)
I see it's not implemented [here](https://github.com/reactstrap/reactstrap/blob/master/src/FormFeedback.js)

virgofx:
Currently not implemented.

PR #1074, opened 2018-06-15
Form feedback tooltip

juanmaguitar:
#1062

### DIFF
--- a/docs/lib/examples/FormFeedback.js
+++ b/docs/lib/examples/FormFeedback.js
@@ -20,12 +20,12 @@ export default class Example extends React.Component {
         <FormGroup>
           <Label for="examplePassword">Invalid input</Label>
           <Input invalid />
-          <FormFeedback validMode={'tooltip'}>Oh noes! that name is already taken</FormFeedback>
+          <FormFeedback tooltip>Oh noes! that name is already taken</FormFeedback>
         </FormGroup>
         <FormGroup>
           <Label for="examplePassword">Invalid input</Label>
           <Input invalid />
-          <FormFeedback valid validMode={'tooltip'}>Sweet! that name is available</FormFeedback>
+          <FormFeedback valid tooltip>Sweet! that name is available</FormFeedback>
         </FormGroup>
       </Form>
     );
--- a/docs/lib/examples/FormFeedback.js
+++ b/docs/lib/examples/FormFeedback.js
@@ -20,8 +20,12 @@ export default class Example extends React.Component {
         <FormGroup>
           <Label for="examplePassword">Invalid input</Label>
           <Input invalid />
-          <FormFeedback>Oh noes! that name is already taken</FormFeedback>
-          <FormText>Example help text that remains unchanged.</FormText>
+          <FormFeedback validMode={'tooltip'}>Oh noes! that name is already taken</FormFeedback>
+        </FormGroup>
+        <FormGroup>
+          <Label for="examplePassword">Invalid input</Label>
+          <Input invalid />
+          <FormFeedback valid validMode={'tooltip'}>Sweet! that name is available</FormFeedback>
         </FormGroup>
       </Form>
     );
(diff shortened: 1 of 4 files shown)


src/FormFeedback.js
```diff
@@ -8,26 +8,30 @@ const propTypes = {
   tag: PropTypes.string,
   className: PropTypes.string,
   cssModule: PropTypes.object,
-  valid: PropTypes.bool
+  valid: PropTypes.bool,
+  validMode: PropTypes.oneOf(['feedback', 'tooltip'])
```
TheSharpieOne: I would prefer just a prop named `tooltip` so the usable would be:

TheSharpieOne: I would prefer a prop named `tooltip` so the usage would be:
```jsx
<FormFeedback>Oh noes! that name is already taken</FormFeedback>
vs
<FormFeedback tooltip>Oh noes! that name is already taken</FormFeedback>
```

juanmaguitar: I was doing that at the same time you were commenting it ;)

src/FormFeedback.js
```diff
@@ -21,13 +22,17 @@ const FormFeedback = (props) => {
...
     tag: Tag,
     ...attributes
   } = props;
 
+  const validMode = tooltip ? 'tooltip' : 'feedback'
+
   const classes = mapToCssModules(classNames(
     className,
-    valid ? 'valid-feedback' : 'invalid-feedback'
+    'd-block',
```
TheSharpieOne: this class needs to be removed. It's causing the feedbacks to display when the shouldn't.

docs/lib/examples/FormFeedback.js
```diff
@@ -20,8 +20,12 @@ export default class Example extends React.Component {
         <FormGroup>
           <Label for="examplePassword">Invalid input</Label>
           <Input invalid />
-          <FormFeedback>Oh noes! that name is already taken</FormFeedback>
-          <FormText>Example help text that remains unchanged.</FormText>
+          <FormFeedback tooltip>Oh noes! that name is already taken</FormFeedback>
```
TheSharpieOne: Can you update the examples so that there is what was previously there (an input with no validation, an input with valid feedback, and an input with invalid feedback) to add an input with valid tooltip feedback, and an input with invalid tooltip feedback.

right now the one example is not working because the field is marked as invalid and the tooltip is set to be valid (this mis-match breaks the display css within bootstrap). Also I don't want to replace the current feedback examples, but instead extend the example to include these new variants

src/FormFeedback.js
juanmaguitar: Oks, i was trying this fast approach to solve an issue with the class `valid-tooltip` that has `display: none`
In the bootstrap examples, the `display:block` is applied to this elements through some parent classes like `was-validated` and some other structures → https://imgur.com/a/Ds5z50D

I don't know how to replicate this behaviour here.
Any suggestions?

juanmaguitar: Something like this could be a good solution for now?

```
const validMode = tooltip ? 'tooltip' : 'feedback';
const classes = mapToCssModules(
    classNames(
      className,
      (validMode === 'tooltip') && 'd-block',
      valid ? `valid-${validMode}` : `invalid-${validMode}`
    ),
    cssModule
  );
```

juanmaguitar: Forget it!! I wasn't get it how this is working. I think it's ok now

juanmaguitar:
@TheSharpieOne Is there anything else to be done here?

src/FormGroup.js
```diff
@@ -32,6 +32,7 @@ const FormGroup = (props) => {
 
   const classes = mapToCssModules(classNames(
     className,
+    'position-relative',
```
iamandrewluca: wouldn't be better to give user control over `position-relative` ?  
If he adds tooltips he can add that class himself!?!

Issue #824, opened 2018-02-09
Add ability to disable fade on <Alert />


charlex:
I see this [PR](https://github.com/reactstrap/reactstrap/pull/339) was merged to allow `<Modal fade={false} />`, but I'm wondering if the `fade={false}` prop could also be used on the `<Alert />` component.

### Issue description

- components: `<Alert />, <Fade />`
- reactstrap version `5.0.0-beta`
- import method `umd` (I don't think I'm using "full")
- react version `16.2.0`
- bootstrap version `4.0.0`

### What is happening?

Attempting to use `fade={false}` on `<Alert />`, much like you can use it on the modal: `<Modal fade={false />`.

### What should be happening?

When using `<Alert fade={false} />` there should be no transition.

TheSharpieOne:
I like this idea and I would even take it a step further. Maybe instead of toggling a single transition, we can have the ability to pick the animation and/or disable it.
The prop can be `animation` (or `transition`, but we already have `transition`...)
`animation="fade"`, (or `transition={Fade}` is we determine to just pass in the transition component) `transition="collapse"`, or `transition={false}`. We can still default to `Fade` but this allows the transition to be disabled as well as changed.

zanisis:
but in transition always give me Warning: Failed prop type: Invalid prop `transition` of type `boolean` supplied to `Alert`, expected `object`. but i already give transition object.

```
<Alert color="dagger" isOpen={this.state._error} transition={{in: true, timeout: 150}}>
  Something Text
</Alert>
```

TheSharpieOne:
@zanisis it's currently not possible and need to be addressed. My previous comment was a suggestion to further enhance the user's control over the transitions.

gergely-nagy:
@TheSharpieOne I'd like to jump in on this! :rocket:

TheSharpieOne:
@gergely-nagy go for it 👍

googamanga:
Just make timeout 0...

mdanilakis:
```
<Alert color="info" transition={{ baseClass: '', timeout: 0 }}>
  Hello world
</Alert>
```

electriccode:
See if this would work #1078

charlex:
Thanks everyone for your work!

stevula:
Since the prop type for transition is now a React element, this can now be accomplished by specifying `<Alert transition={null}>...</Alert>`.

PR #1078, opened 2018-06-16
#824 Add ability to disable fade on <Alert />

electriccode:


src/UncontrolledAlert.js
```diff
@@ -1,12 +1,22 @@
...
+  fade: true
+};
+
+class UncontrolledAlert extends Component {
   constructor(props) {
     super(props);
 
     this.state = { isOpen: true };
     this.toggle = this.toggle.bind(this);
+    this.fade = this.props.fade;
```
TheSharpieOne: I don't think any changes are needed to this file, but at the very least, this change does nothing.

electriccode:
@TheSharpieOne done

TheSharpieOne:
Yeah, we usually just don't define the props for the uncontrolled components but when I was cleaning it up I thought it would be interesting if we exposed the original props through. I probably should have only done it for prop types and not default props but I didn't give it much thought.

electriccode:
I can add back `prop types` from `Alert` component, but this is not observed at other places so lets stick with this for now.

Issue #1004, opened 2018-05-04
Tooltip not closing on mobile


ntucakovic:
### Issue description
Tooltip can't close on mobile devices when opened on hover event.

- components: `Tooltip` (probably Popover as well, haven't tried)
- reactstrap version `#5.0.0` also on `#6.0.1`
- import method `es`
- react version `#16.3.1`
- bootstrap version `#4.0.0`

### What is happening?

Compare the behavior between the two on mobile (bugs on both Android and iOS).
Reactstrap: Tooltip is opened on hover event and you can't close it.
Bootstrap: Tooltip is opened on click only and can be closed when clicked away.

https://reactstrap.github.io/components/tooltips/
https://getbootstrap.com/docs/4.1/components/tooltips/

### What should be happening?

Tooltip should close when clicking away, anywhere on the page.

ntucakovic:
This is also happening on latest Reactstrap version, v6.0.1. Updated issue to reflect that.

Kofaka:
Have the same thing.

olefrank:
Anyone found a solution to this?

nathanbacon:
Just submitted a PR #1083 that fixes this.

ntucakovic:
Any update on when this will be released? I got an email that it's merged :)

mounlion:
Thanks you!

PR #1083, opened 2018-06-19
#1004 fixed tooltip not disappearing on mobile

nathanbacon:
I tested this on iOS and the proper behavior is observed. I hope clearing the show timeout is on the right track. (#1004 )

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -157,7 +157,7 @@ class Tooltip extends React.Component {
       if (!this.props.isOpen) {
         this.toggle();
       }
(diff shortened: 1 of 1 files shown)


ntucakovic:
Just tested on Android's latest Chrome and on Mac OS' Chrome device emulator. Works as expected.

TheSharpieOne:
This breaks the "Tooltip Disable Autohide" feature which allows users to select the text within the tooltip. Now, when clicking to select the text, it will close the tooltip (on desktop).

Issue #1079, opened 2018-06-16
Feature: Close Tooltip when clicking outside of it (useful when high hide delay set)


juanmaguitar:
### Issue description

- components: `Tooltip`
- reactstrap version `6.1.0`

### What is happening?

If the Tooltip is set with some delay on hide, clicking outside doesn't close it
https://codesandbox.io/s/j3lp8k5np9

### What should be happening?

Clicking outside the tooltip should close it

TheSharpieOne:
This is not how bootstrap has implemented it. In order to mimic bootstrap as much as possible, we cannot have it close when clicked outside.

bootstrap: https://codepen.io/TheSharpieOne/pen/pKpzBj

If you feel this is an issue, please raise it bootstrap. Once bootstrap changing it's implementation, reactstrap can do the same.

juanmaguitar:
Opened the issue on bootstrap → https://github.com/twbs/bootstrap/issues/26719

PR #1085, opened 2018-06-21
Tooltip delay can't be circumvented by clicking away from trigger 

nathanbacon:
This unimplements #1079 by checking if there is a hide timer already active. This mimics the behavior of Bootstrap, see [#1079 (comment)](https://github.com/reactstrap/reactstrap/pull/1083#pullrequestreview-130130462)


bpas247:
Is this issue still relevant? If so, then we should update the branch and resolve the merge conflicts.

Issue #1082, opened 2018-06-18
Use React 16's forwardRef for most/all components


seansfkelley:
reactstrap is a perfect use-case for React 16's new [`forwardRef` API](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#forwardref-api). Since most components in this library have predictable and well-defined DOM they produce, it would be very useful to have them forward their refs along to the underlying elements so consumers like myself don't have to wrap access in `ReactDOM.findDOMNode` when we want to e.g. implement a scroll-related feature.

juanmaguitar:
Totally Agree!! We're using some `reactstrap` components in our project and in order to expand the current behavior it would be really handy (and much more consistent) being able to select the DOM elements via `ref` selections in the same components

For example, doing this in the `Tooltip` component 

```
class Tooltip extends React.Component {
  constructor(props) {
    super(props);

    this.refTooltipInner = React.createRef();
    ...
  }

  render() {
    ...
     return (
      <PopperContent
        ...
      >
        <div
          ref={this.refTooltipInner}
          ...
        />
      </PopperContent>
    );
  }
}
```

Would allow to have access to this DOM element easily _from the outside_

TheSharpieOne:
I'm all for this, but it will be a breaking change since it will change the require version of react needed as a peerDep.

illiteratewriter:
If adding `refs` is all that is needed, we can do that like how we did [https://github.com/reactstrap/reactstrap/issues/1054](url), so that we don't have to change react version, right? So wouldn't it be better to do it like that so we don't have to bumb the version up?

seansfkelley:
True. You could put `innerRef` on everything, and use this issue as a tracking issue for the eventual deprecation and replacement with `forwardRef` as noted in #1054.

seansfkelley:
@TheSharpieOne when would you accept a PR for this breaking change? I keep coming up against this cause I want to attach refs to things that often turn out to be stateless components. I'd be open to making a PR to apply the change as well. We might be able to keep `innerRef` around as a deprecated alias to preserve API-compatibility even as the minimum React version is raised.

It's a shame that 16.3 was introduced not 6 days after #916 dropped support for versions below 16. :(

TheSharpieOne:
there is a v7 branch, make PRs with breaking changes to that branch, eventually it will be merged and released.

seansfkelley:
[...]

PR #1087, opened 2018-06-23
feat(Modal): add ref to Modal (#1082)

illiteratewriter:


docs/lib/Components/ModalsPage.js
```diff
@@ -93,10 +93,11 @@ export default class ModalsPage extends React.Component {
   // timeout is 150ms by default to match bootstrap
   // see [Fade](/components/fade/) for more details
   backdropTransition: PropTypes.shape(Fade.propTypes),
-  // modalTransition - controls modal transition 
+  // modalTransition - controls modal transition
   // timeout is 300ms by default to match bootstrap
   // see [Fade](/components/fade/) for more details
   modalTransition: PropTypes.shape(Fade.propTypes),
+  innerRef: PropTypes.object,
```
TheSharpieOne: I missed this, but technically, this can still be a string or a function as well as the new createRef object.

src/Modal.js
```diff
@@ -48,6 +48,7 @@ const propTypes = {
   ]),
   backdropTransition: FadePropTypes,
   modalTransition: FadePropTypes,
+  innerRef: PropTypes.object,
```
TheSharpieOne: I missed this, but technically, this can still be a string or a function as well as the new createRef object.

Issue #1089, opened 2018-06-25
add innerRef to Tooltip


juanmaguitar:
### Issue description

- components: `Tooltip`
- reactstrap version `6.1.0`

### What is happening?

There's no way to access (from the outside) to the DOM element that contains the tooltip

### What should be happening?

There should be a way to access (from the outside) the DOM element that contains the tooltip. 

### Code

This can be easily achieved by adding a `ref={this.props.innerRef}` to the `div` that wraps the tooltip

This issue may be related to https://github.com/reactstrap/reactstrap/issues/1082

PR #1090, opened 2018-06-25
Tooltip inner ref

juanmaguitar:
Adding a innerRef prop to the `div` wrapping the `Tooltip` so it can accesed _from the outside_
This PR solves the issue https://github.com/reactstrap/reactstrap/issues/1089

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -34,6 +34,7 @@ const propTypes = {
     PropTypes.string,
     PropTypes.number
   ]),
+  innerRef: PropTypes.object,
 };
 
 const DEFAULT_DELAYS = {
@@ -225,6 +226,7 @@ class Tooltip extends React.Component {
       >
         <div
           {...attributes}
+          ref={this.props.innerRef}
           className={classes}
           role="tooltip"
           aria-hidden={this.props.isOpen}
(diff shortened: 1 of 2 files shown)


src/Tooltip.js
```diff
@@ -34,6 +34,7 @@ const propTypes = {
     PropTypes.string,
     PropTypes.number
   ]),
+  innerRef: PropTypes.object,
```
TheSharpieOne: Technically this can still be a string or a function as well as the new createRef object.

Also need to document the prop (at the very least add it to the props on the docs for this component)

juanmaguitar: You're right. I've changed the propType definition and added the proper explanation in the docs

src/Tooltip.js
```diff
@@ -34,7 +34,10 @@ const propTypes = {
     PropTypes.string,
     PropTypes.number
   ]),
-  innerRef: PropTypes.object,
+  innerRef: PropTypes.oneOfType([
+    PropTypes.func, 
+    PropTypes.string
```
TheSharpieOne: `PropTypes.object` as well, I was merely stating that it can be these 2 in addition to the one previously specified.

PR #1095, opened 2018-06-26
Added element as a prop type for the label prop

iKonrad:
Hi, for the custom input field, when I pass a react element as a label it throws a validation error from PropTypes.

With this solution I can't pass a `I agree to <a>Terms and conditions</a>` element because it's rendering as a text.

I think Custom Input should accept either text or element as labels.

### DIFF
--- a/src/CustomInput.js
+++ b/src/CustomInput.js
@@ -7,7 +7,7 @@ const propTypes = {
   className: PropTypes.string,
   id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
   type: PropTypes.string.isRequired,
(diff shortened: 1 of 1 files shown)


src/CustomInput.js
```diff
@@ -7,7 +7,7 @@ const propTypes = {
   className: PropTypes.string,
   id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
   type: PropTypes.string.isRequired,
-  label: PropTypes.string,
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
TheSharpieOne: probably just want `PropTypes.node` which is defined as:Anything that can be rendered: numbers, strings, elements or an array (or fragment) containing these types.

Issue #1058, opened 2018-06-04
Opening and closing a modal causes fixed content to jump


banderso-n:
### Issue description

- components: `Modal`
- reactstrap version `6.0.1`
- import method `umd/csj/es` 
- react version `16.2.0`
- bootstrap version `4.1.1`

### What is happening?

Opening and closing a modal causes fixed content to jump as the body's scrollbar disappears and reappears.

### What should be happening?

Fixed content should appear to stay in place the way unfixed content does.

### Steps to reproduce issue

1. Have position:fixed content on the page (e.g. with a `.fixed-top` class).
1a. It's more noticeable when there's content that's aligned to the right side of the fixed content.
2. Have enough content on the page so that a scrollbar appears.
3. Open and close a modal.
4. Observe right-aligned content appears to jump as the body's scrollbar disappears and reappears.

### Error message in console

N/A

### Code

https://codepen.io/banderso-n/pen/OENMNO

PR #1099, opened 2018-06-26
fix(Modal): Opening and closing a modal causes fixed content to jump (#1058)

banderso-n:
Update setScrollbarWidth to additionally apply padding to full-width position-fixed content.

https://github.com/reactstrap/reactstrap/issues/1058

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -16,7 +16,12 @@ export function getScrollbarWidth() {
 }
 
 export function setScrollbarWidth(padding) {
-  document.body.style.paddingRight = padding > 0 ? `${padding}px` : null;
+  const paddingRight = padding > 0 ? `${padding}px` : null;
+  const fullWidthFixedContent = document.querySelectorAll('.fixed-top, .fixed-bottom');
+  const elementsAffectedByScrollbarWidth = [document.body, ...fullWidthFixedContent];
+  elementsAffectedByScrollbarWidth.forEach((element) => {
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -16,7 +16,12 @@ export function getScrollbarWidth() {
 }
 
 export function setScrollbarWidth(padding) {
-  document.body.style.paddingRight = padding > 0 ? `${padding}px` : null;
+  const paddingRight = padding > 0 ? `${padding}px` : null;
+  const fullWidthFixedContent = document.querySelectorAll('.fixed-top, .fixed-bottom');
+  const elementsAffectedByScrollbarWidth = [ document.body, ...fullWidthFixedContent ];
+  elementsAffectedByScrollbarWidth.forEach(element => {
+    element.style.paddingRight = paddingRight;
```
TheSharpieOne: When these component re-render, it will end up removing the padding which is added here.

banderso-n: Good catch. Back to the drawing board.

banderso-n:
As @TheSharpieOne pointed out, when the components re-render, the padding will get blown away. I don't think this approach will work so I'll close the PR.

src/utils.js
TheSharpieOne: The only why I can think this would work would be using context so that a provider can inform the consumers what the padding currently is as well as re-render them when it changes.

PR #1114, opened 2018-07-07
feat(UncontrolledCollapse): isOpen prop (#1110)

### LINKED ISSUE #1110: Set default or initial state of UncontrolledCollapse to isOpen
How can I make the UncontrolledCollapse to be opened initially.?

devshawn:
Adds isOpen prop to allow UncontrolledCollapse to be open by default if desired. Closes #1110

### DIFF
--- a/docs/lib/Components/CollapsePage.js
+++ b/docs/lib/Components/CollapsePage.js
@@ -6,6 +6,7 @@ import SectionTitle from '../UI/SectionTitle';
 
 import CollapseExample from '../examples/Collapse';
 import UncontrolledCollapseExample from '../examples/CollapseUncontrolled';
+import UncontrolledCollapseIsOpenExample from '../examples/CollapseUncontrolledIsOpen';
 
 import CollapseEventsExample from '../examples/CollapseEvents';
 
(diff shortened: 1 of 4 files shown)


devshawn:
Sure, I get that -- sorry if I don't completely understand... I thought the point of an Uncontrolled component was to not manage state/value programmatically from your own component?

Are you saying to rename it to something like defaultValue? I'm not sure how it removes the ability to take manual control, it just sets the default state and the rest is toggled as before through the toggler. I'm just a bit confused on your response, sorry! :)

TheSharpieOne:
You are pulling `isOpen` out of the (`rest`) props which are passed so it cannot override the `isOpen={this.state.isOpen}` defined: `<Collapse isOpen={this.state.isOpen} {...rest} />`

Previously, if the user defined `isOpen` it would override, with this change it cannot. While an edge-case, it's useful when manual control has to be taken based on other actions the user has performed but not have to control it entirely.

devshawn:
Sorry for the long delay, I got way too busy and forgot :) - what would you like the prop changed to? defaultValue? initialValue? something else?

TheSharpieOne:
I think this will be addressed in https://github.com/reactstrap/reactstrap/pull/1222/files#diff-838b2b775bc166813a8e6a5744718a24

Issue #1022, opened 2018-05-16
It should be possible to change the trigger for UncontrolledTooltip


DiabolusExMachina:
### Issue description

- components: `UncontrolledTooltip`
- reactstrap version `6.0.1`
- react version `16.3.2`
- bootstrap version `^4.1.1`

### What is happening?

It is not possible to change the trigger of the UncontrolledTooltip. It is always **hover** 

### What should be happening?
It should be possible to change the trigger for UncontrolledTooltip like this

```
<UncontrolledTooltip trigger="focus" placement="right" target="id">
           Some Text
</UncontrolledTooltip>
```
Popper.js Supports this see here: https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/tooltip-documentation.md

TheSharpieOne:
I have been wanting to better align tooltip and popover (as well as their uncontrolled counterparts). It seems like they do similar things in similar ways except the trigger is different... but the trigger, as this issue mentions, should be able to be changed. Technically, you could have a popover work on mouse events and a tooltip work on click events. The only real different between the two would be the default events it uses and the classNames which align with the bootstrap styles for each.
If anyone wants to take a stab at resolving this, please keep the above in mind and also address this similar concern with the popover component (maybe by creating a base component with all of the logic and having other components which implements the defaults for tooltip and popover.)

illiteratewriter:
@TheSharpieOne  `react-popper` has an events prop. Wouldn't passing a prop to `react-popper` solve this?

TheSharpieOne:
react-popper has [eventsEnabled](https://github.com/FezVrasta/react-popper#eventsenabled) which defaults to `true`. It controls [eventsEnabled from Popper.js](https://popper.js.org/popper-documentation.html#Popper.Defaults.eventsEnabled) which just determines if it should listen for resize and scroll events to update.

illiteratewriter:
I'll give this a shot.

illiteratewriter:
Will do for popover too...

illiteratewriter:
@TheSharpieOne I want to build this like you said (**by creating a base component with all of the logic and having other components which implements the defaults for tooltip and popover.**), but I have never written tests, so when I write tests, should I write them for the base component or for the Tooltip component?

PR #1128, opened 2018-07-12
feat(Tooltip) : add trigger for tooltip (#1022)

illiteratewriter:
trigger added for tooltip

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -40,6 +40,7 @@ const propTypes = {
     PropTypes.string,
     PropTypes.object
   ]),
+  trigger: PropTypes.string,
 };
 
 const DEFAULT_DELAYS = {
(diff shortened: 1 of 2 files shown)


src/Tooltip.js
```diff
@@ -173,16 +174,34 @@ class Tooltip extends React.Component {
...
-    this._target.addEventListener('focusin', this.show, true);
-    this._target.addEventListener('focusout', this.hide, true);
-
-
-    ['click', 'touchstart'].forEach(event =>
-      document.addEventListener(event, this.handleDocumentClick, true)
-    );
+    if (this.props.trigger) {
+      let triggers = this.props.trigger.split(' ');
+      if (!triggers.includes('manual')) {
```
TheSharpieOne: `Array.prototype.includes` is not supported by IE. Rather than requiring the users to add a polyfill, can you use `Array.prototype.indexOf` can check that it is `-1`. Similar for the usages below (but checking that it is `-1`)

src/Tooltip.js
```diff
@@ -176,17 +176,17 @@ class Tooltip extends React.Component {
   addTargetEvents() {
     if (this.props.trigger) {
       let triggers = this.props.trigger.split(' ');
-      if (!triggers.includes('manual')) {
-        if (triggers.includes('click')) {
+      if (triggers.indexOf('manual') === -1) {
+        if (triggers.indexOf('click') >= -1) {
```
TheSharpieOne: You probably want ` !== -1` or `> -1`

PR #1136, opened 2018-07-20
feat(DropdownMenu): centering dropdown menus

vepanimas:
A user should have an ability to render center aligned dropdowns. The Popper.JS library supports this behavior, so we can easily implement it in the reactstrap.


TheSharpieOne:
Note: bootstrap does **not** have this option http://getbootstrap.com/docs/4.1/components/dropdowns/

vepanimas:
You're right. I'm trying to fix it here https://github.com/twbs/bootstrap/pull/26938. Please, don't close it yet.

PR #1145, opened 2018-07-26
feat(Dropdown): use portal in dropdowns (#1098)

### LINKED ISSUE #1098: Dropdown doesn't work properly with Portals
### Issue description

- components: `Dropdown`, `DropdownMenu`
- reactstrap version `#6.1.0`
- import method `es` 
- react version `#16.1.1`
- bootstrap version `#4.1.1`

### What is happening?

I'm using both `reactstrap` and `react-virtualized` Table. Table has an `overflow: hidden` on its rows, so to show Dropdown properly I need to use Portals. Despite warnings about incorrect children for `Manager`, dropdown is shown perfectly well. But if I add a `react-router` Link (or any other customized `<a>` component with `onClick` property) here, everything breaks. 

I'm not sure what the exact reason of this, but if I use Dropdown without a Portal, `onClick` handlers works, and if I use Portal, it is ignored completely.

### What should be happening?

I expect Dropdown working for Portals as well as it works without it

### Steps to reproduce issue

1. Create a simple component that uses `DropdownItem` with `tag={Link}`, where `Link` is a simple component that adds click handler with `e.preventDefault()`.
2. Click on dropdown item.
3. Page will start to reload despite it is prevented.

### Error message in console
No error message, page just starts to reload, because `e.preventDefault()` for `<a>` doesn't work

### Code
[Example](https://stackblitz.com/edit/react-suzwrk?file=DropdownWithPortal.js) (red dropdown does have portal, blue one does not).

xbojch:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

Two tests are failing because the DropdownMenu is now wrapped in a Portal. Would love to get input on how to work around either by:
- changing failing tests
- making the use of portal optional and add additional tests

Closes #1098

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -42,11 +42,14 @@ const childContextTypes = {
   isOpen: PropTypes.bool.isRequired,
   direction: PropTypes.oneOf(['up', 'down', 'left', 'right']).isRequired,
   inNavbar: PropTypes.bool.isRequired,
+  menuContainer: PropTypes.oneOfType(Element).isRequired
 };
 
 class Dropdown extends React.Component {
   constructor(props) {
     super(props);
+
+    this._menuContainer = document.createElement('div');
 
     this.addEvents = this.addEvents.bind(this);
     this.handleDocumentClick = this.handleDocumentClick.bind(this);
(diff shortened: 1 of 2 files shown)


virgofx:
As mentioned in other comments, I'm weary of these changes. I don't think Portals should be used for dropdown and their are probably better ways to solve the underlying issue.

TheSharpieOne:
We use react-popper for the dropdowns I don't think they are intented to be used with portals like this.

xbojch:
Created [an example](https://stackblitz.com/edit/react-ttkj7j) to show the issue I was overcoming with a Portal. Perhaps making it optional is the way to go? If there is a better way by using popper directly I'm all open for it.

Issue #1118, opened 2018-07-10
UncontrolledTooltips do not position correctly if the wrapper has an overflow: auto


Muntaner:
### Issue description

- components: `UncontrolledTooltip`
- reactstrap version `^6.1.0`
- react version `^16.4.0`
- bootstrap version `^4.1.1`

### What is happening?

The title says it all: UncontrolledTooltips do not position correctly if the wrapper has an overflow: auto.
After a lot of searching, [I landed on this popper.js issue](https://github.com/FezVrasta/popper.js/issues/648) which is exactly the problem I am reporting for Reactstrap.
### What should be happening?

UncontrolledTooltip should be placed correctly on the target DOM element.

### Steps to reproduce issue

1. Create a div and wrap it in another div with the CSS (or the React style) overflow: auto.
2. Create an UncontrolledTooltip and set as target this div.

### Error message in console
No error in console, just UX.

### Code
Check the following section.

### A solution I found
By looking at the issue [I pointed at previously](https://github.com/FezVrasta/popper.js/issues/648), seems like the problem is the following: 
> It's the correct behavior, the preventOverflow modifier uses by default, as boundariesElement, the scrollParent, you can change it to viewport or window or anything else supported.

So, I resolved the issue by adding the `modifiers` prop to the UncontrolledTooltip:

`
<UncontrolledTooltip modifiers={{preventOverflow: {boundariesElement: 'window'}}} autohide={false} delay={{ show: 200, hide: 5 }} placement='left' target={`${id}_${j}`}>
...
</UncontrolledTooltip>
`

TheSharpieOne:
Looks like [bootstrap exposes an easy way to set the boundary](https://github.com/twbs/bootstrap/blob/abb48686267db3dadaa390e62538d23216c046d3/js/src/tooltip.js#L298-L300) (also [documented](http://getbootstrap.com/docs/4.1/components/tooltips/#options). We should probably expose a `boundary` prop which sets the `preventOverflow.boundariesElement` in the same way.

pachuka:
@TheSharpieOne - I think the boundariesElement can also apply to the Dropdown/UncontrolledDropdown, should a new issue be created for that?

tarik-besic:
Nice solution with the modifiers prop, works great now!

PR #1149, opened 2018-07-31
Pass boundariesElement prop from UncrontrolledTooltip -> Popper (#1118)

monteith:
Pass boundariesElement prop from UncrontrolledTooltip -> Tooltip -> Popper modifiers

- [x] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

Addresses issue #1118

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -19,6 +19,7 @@ const propTypes = {
   isOpen: PropTypes.bool,
   disabled: PropTypes.bool,
   hideArrow: PropTypes.bool,
+  boundariesElement: PropTypes.string,
   className: PropTypes.string,
   innerClassName: PropTypes.string,
   arrowClassName: PropTypes.string,
@@ -247,6 +248,7 @@ class Tooltip extends React.Component {
         target={this.props.target}
         isOpen={this.props.isOpen}
         hideArrow={this.props.hideArrow}
+        boundariesElement={this.props.boundariesElement}
         placement={this.props.placement}
         placementPrefix={this.props.placementPrefix}
         arrowClassName={this.props.arrowClassName}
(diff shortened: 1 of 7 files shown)


src/Tooltip.js
```diff
@@ -247,6 +248,7 @@ class Tooltip extends React.Component {
         target={this.props.target}
         isOpen={this.props.isOpen}
         hideArrow={this.props.hideArrow}
+        boundariesElement={this.props.boundariesElement}
```
TheSharpieOne: Can you also make a similar change for the `Popover` component?

src/UncontrolledTooltip.js
```diff
@@ -17,3 +22,7 @@ export default class UncontrolledTooltip extends Component {
     return <Tooltip isOpen={this.state.isOpen} toggle={this.toggle} {...this.props} />;
   }
 }
+
+UncontrolledTooltip.propTypes = propTypes;
```
TheSharpieOne: This is not really needed. The props are not explicitly used or accessed by this component, they are just passed down. No need to specify them here but if you want, you can do `UncontrolledTooltip.propTypes = Tooltip.propTypes` since they will always be the same.

src/Tooltip.js
monteith: 👍 done

monteith: 👍 done

NathanCH:
This is a great fix! How are releases typically handled in Reactstrap? Would you expect this to go out in 6.3.2?

TheSharpieOne:
it will go in `6.4.0` since it is a feature and that indicates a minor version bump. There is a process documented in the README for how to initiate a release else I am going to wait for some more fixes to put into the release.

seansfkelley:
Ah, ok, that makes sense, especially if Bootstrap does similar. Thanks for the explanation!

Issue #1156, opened 2018-08-02
Dropdown Select according keypress


Lug16:
Hi all,
I would like to know if there is any way to select an item in the Dropdown component according to a key pressed. Sorry if this is not the way to ask a question here.

TheSharpieOne:
👋 Hi @Lug16! Do mean like a hotkey (pressing `3` would select the 3rd item) or typeahead (pressing `H` would select the item starting with "H")?

Either way is not supported as bootstrap itself doesn't have that feature. The only dropdown keyboard support is arrow keys (shifting which item is active) and enter/space (selecting the active item)

I wouldn't mind adding a feature like this as this sounds like it can help with accessibility.

Lug16:
Hey @TheSharpieOne thanks for your response. Yeah, I meant (pressing `H` would select the item starting with "H")
It would be awesome to have something like that, I'm not a very experienced JS developer, but if you need my help, I will gladly help you.

Thanks

Lug16:
Guys, I know that the thread is closed and the PR adding the functionality is merged, but just downloaded the latest version and I was not able to make it work. Here is a sandbox with the test that I did. https://codesandbox.io/s/qqq4q74o26 . Am I missing something?

TheSharpieOne:
it hasn't been released yet. I'll try to cut a release some time today.

bugovicsb:
Can this feature be switched off somehow? I would like to implement my own search functionality on keydown.

PR #1160, opened 2018-08-04
Select first element matching pressed key in dropdown (#1156)

monteith:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [] My change requires a change to the documentation.

#1156

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -132,33 +132,28 @@ class Dropdown extends React.Component {
     const disabledClass = mapToCssModules('disabled', this.props.cssModule);
 
     const items = container.querySelectorAll(`.${menuClass} .${itemClass}:not(.${disabledClass})`);
-    const char = String.fromCharCode(e.which).toLowerCase();
-    const itemsOfChar = char !== '' &&
-      Array.from(items).filter(item =>
-        !!item.textContent && item.textContent.substr(0, 1).toLowerCase() === char);
-
     if (!items.length) return;
 
     let index = -1;
 
-    if (itemsOfChar.length) {
-      index = Array.from(items).indexOf(itemsOfChar[0]);
-    } else {
-      for (let i = 0; i < items.length; i += 1) {
-        if (items[i] === e.target) {
-          index = i;
-          break;
-        }
-      }
+    const charPressed = String.fromCharCode(e.which).toLowerCase();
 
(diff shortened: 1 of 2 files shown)


src/Dropdown.js
```diff
@@ -132,23 +132,32 @@ class Dropdown extends React.Component {
     const disabledClass = mapToCssModules('disabled', this.props.cssModule);
 
     const items = container.querySelectorAll(`.${menuClass} .${itemClass}:not(.${disabledClass})`);
+    const char = String.fromCharCode(e.which).toLowerCase();
+    const itemsOfChar = char !== '' &&
```
TheSharpieOne: `Array.from` is is not supported in IE and also does not get transpiled. It would require a polyfill.

src/Dropdown.js
```diff
@@ -132,23 +132,32 @@ class Dropdown extends React.Component {
...
+      Array.from(items).filter(item =>
```
TheSharpieOne: charAt or direct index access would have better performance

src/Dropdown.js
```diff
@@ -132,23 +132,32 @@ class Dropdown extends React.Component {
...
+        !!item.textContent && item.textContent.substr(0, 1).toLowerCase() === char);
 
     if (!items.length) return;
 
     let index = -1;
-    for (let i = 0; i < items.length; i += 1) {
-      if (items[i] === e.target) {
-        index = i;
-        break;
+
```
TheSharpieOne: If `char` is an empty string `itemsOfChar` will be `false` and this will throw an error.

src/__tests__/Dropdown.spec.js
```diff
@@ -452,7 +452,7 @@ describe('Dropdown', () => {
 
       expect(Dropdown.prototype.toggle.mock.calls.length).toBe(0);
 
-      wrapper.find('#first').hostNodes().simulate('keydown', { which: 52 });
+      wrapper.find('#dropdown').hostNodes().simulate('keydown', { which: 52 });
```
monteith: This test doesn't work when we trigger `keydown` events directly on the dropdown items, but I'm unable to actually replicate that scenario from my browser. It appears to me like this test is a more accurate representation of a user interaction. And it makes the tests pass. Of course, I could be wrong! 😄

Issue #310, opened 2017-01-27
Modal should trap focus


TheSharpieOne:
- version `v4.0.1`
- components: `Modal`

### Steps to reproduce issue

Open modal and hit tab. Focus should stay in the modal. You should not be able to focus on the element behind the modal/backdrop.

https://reactstrap.github.io/components/modals/
http://v4-alpha.getbootstrap.com/components/modal/

toddgeist:
Safari on mac OS Sierra shows the problem.
With Chrome, you can't focus on anything behind the modal.  But it tabbing does take you off to the Window elements. So it isn't as bad.

chrismay:
> With Chrome, you can't focus on anything behind the modal. 

Chrome (65 on Linux, at least) _does_ allow you to focus on elements behind the modal, but you have to tab through the URL bar first. 

On the reactstrap demo page linked above, clicking "launch modal" and then pressing tab-tab-tab-tab-tab-esc-tab will put the focus onto the first clickable element on the page, the "Reactstrap" link behind the modal. (The "esc" dismisses the URL bar autocomplete). Subsequent tabs will navigate through the component menu on the right-hand-side of the page.

PR #1161, opened 2018-08-04
Trap focus in modal (#310)

monteith:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [] My change requires a change to the documentation.

I was unable to implement a test with enzyme. Focus on elements doesn't seem to advance with tabs in a mounted component. I could use some help understanding how to accomplish this test.

#310

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -9,6 +9,7 @@ import {
   setScrollbarWidth,
   mapToCssModules,
   omit,
+  focusableElements,
   TransitionTimeouts
 } from './utils';
 
@@ -86,6 +87,7 @@ class Modal extends React.Component {
     this.handleBackdropMouseDown = this.handleBackdropMouseDown.bind(this);
     this.handleBackdropMouseUp = this.handleBackdropMouseUp.bind(this);
     this.handleEscape = this.handleEscape.bind(this);
+    this.handleTab = this.handleTab.bind(this);
     this.onOpened = this.onOpened.bind(this);
     this.onClosed = this.onClosed.bind(this);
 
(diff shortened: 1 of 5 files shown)


src/Modal.js
```diff
@@ -182,6 +184,31 @@ class Modal extends React.Component {
     }
   }
 
+  handleTab(e) {
+    if (e.which !== 9) return;
+
+    const focusableChildren = this._element.querySelectorAll(focusableElements.join(', '));
+    const currentFocus = document.activeElement;
```
TheSharpieOne: `document.activeElement` can have issues in IE, specially within iframes.
I just ran into it the other day, https://github.com/Availity/react-block-ui/issues/18

src/utils.js
```diff
@@ -260,3 +260,16 @@ export function addMultipleEventListeners(_els, handler, _events) {
...
+  'a[href]',
+  'area[href]',
+  'input:not([disabled])',
+  'select:not([disabled])',
+  'textarea:not([disabled])',
+  'button:not([disabled])',
+  'object',
+  'embed',
+  '*[tabindex]:not(.modal)',
+  '*[contenteditable]',
```
TheSharpieOne: Probably also want `'audio[controls]'`, `'video[controls]'`, and to account for `contenteditable="false"` with `'[contenteditable]:not([contenteditable="false"])'`

Also, do you need the `*` in the selectors?

src/utils.js
```diff
@@ -260,3 +260,16 @@ export function addMultipleEventListeners(_els, handler, _events) {
     });
   };
 }
+
+export const focusableElements = [
+  'a[href]',
+  'area[href]',
+  'input:not([disabled])',
```
TheSharpieOne: Need to also account for `type="hidden"`

src/Modal.js
monteith: Wrapped this in a try/catch, defaulting to the first focusable child. Tested on IE10, IE11 on win7 and MSEDGE on win10. Let me know if I missed a version.

src/utils.js
monteith: Maybe not. Personal preference. Removed 👍

src/utils.js
monteith: 👍

PR #1171, opened 2018-08-12
feat(Button): onClick of inner button returns value of props.onClick.

fmichea:
This change makes Button implementation return the value of props.onClick in its own onClick function.

Rationale for the change: in my own use case, I have some onClick function call fetch methods. These onClick methods return the promise returned by fetch. For normal application use, this does not do anything. In unit tests however, this allows the unit-test to trigger the click and await on the handling of the response before doing assertions (using fetch-mock...). After switching from react-bootstrap to reactstrap, those tests do not work anymore since this onClick implementation does not pass the promise along.

This change is backwards compatible if you assume nobody relies on onClick returning undefined. Please let me know if I can improve this change in any way.

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 


fmichea:
Side note: I got have a temporary work around, I will do a proper clone to redo this commit to fix the tests and re-submit soon.

PR #1181, opened 2018-08-18
feat(Tooltip,Popover): base component for Tooltip and Popover (#1022)

illiteratewriter:
Added a base component which can be shared between both Tooltip and 
Popover. Also add custom trigger for Popover component

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 


src/TooltipPopoverWrapper.js
```diff
@@ -0,0 +1,286 @@
...
+    clearTimeout(this._showTimeout);
+    this._showTimeout = undefined;
+  }
+
+  clearHideTimeout() {
+    clearTimeout(this._hideTimeout);
+    this._hideTimeout = undefined;
+  }
+
+  handleDocumentClick(e) {
```
TheSharpieOne: I think the logic between tooltip and popover is really the same but it was just implemented differently in the original components. It looks like they do the opposite checks to perform the opposite action.

There is some extra stuff on the second condition more specific to `this._popover`, but it is just checking that the content is not within the popover. A similar check is probably needed for tooltip or made into another prop to control clicks within the "add element".

The goal is for this to be as generic as possible.

illiteratewriter: This is done by passing a `ref` as `prop` down to the `div` inside, right? Since `Tooltip` is providing `innerRef` as a `prop`, how can I do the same for `Tooltip` (since the function can change from the parent created by the user)?

TheSharpieOne:
superseded by #1222

PR #1190, opened 2018-08-28
fix(Modal): use static openCount to become resilient to classList modification

kraenhansen:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

For the last two bullets - I didn't add a test, but modified the existing tests.

Calls to document.body.classList removes duplicate class names.
This commit makes the Modal more resilient to these modifications by keeping a static counter on open Modals and adds "modal-open" only when the
initializing the first modal and removes "modal-open" only when destroying the last modal.

This fixes reactstrap#1189

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -249,16 +249,13 @@ class Modal extends React.Component {
     conditionallyUpdateScrollbar();
 
     document.body.appendChild(this._element);
-    if (!this.bodyClassAdded) {
-      if (Modal.openCount === 0) {
-        document.body.className = classNames(
-          document.body.className,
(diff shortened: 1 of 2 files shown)


virgofx:
Thanks for the PR

kraenhansen:
Thanks for the simplification @TheSharpieOne.

Another approach that I considered was to add a field on the `document.body` so the component class could be used across multiple documents without getting confused, but that felt a bit too hacked.

Anything else that needs to happen before a merge and released?

kraenhansen:
Any chance this could get released in the near future? I have an app in production that suffers from this bug and I would love to fix it by simply upgrading 👍

Issue #1198, opened 2018-09-03
Full support for React 16.3 ref objects as `target` property


josh-degraw:
According to #947, there should be support for using the React 16.3 `createRef` api as the `target` element in components such as `Tooltip`. If this is the case, the type definitions should be updated to reflect this. 

I am currently trying to use this api to use tooltips and I'm using Typescript, and if this is implemented, the type definitions do not reflect this. I'd be happy to submit a PR to fix this if necessary.

TheSharpieOne:
I believe your usage may not be right. `Tooltip`'s `target` accepts a DOMElement which can be obtained React 16.3 "ref objects" via `myref.current`. You should be passing in the current reference from the object and not the object itself.
```jsx
this.myRef = React.createRef();
//...
<div ref={this.myRef} />
// this would not return the reference
<Tooltip target={this.myRef}>
// This is how you get the reference
<Tooltip target={this.myRef.current}>
```

Just as before the ref object people used functions to obtain the ref, those functions could not be used as `target`, but the ref obtained from those functions could be used: 
```jsx
getRef = ref => {
  this.myRef = ref;
}
//...
<div ref={this.getRef} />
// this would be the "old way" equivalent of what you are trying to do
<Tooltip target={this.getRef}> {/* This never worked */}
// this is how the "old way" worked
<Tooltip target={this.myRef}> {/* This works */}
```

Note: the `target` _must_ be provided, so you need to ensure you have the ref before you attempt to render `Tooltip`

josh-degraw:
@virgofx Cool I'll start in on it.

@TheSharpieOne I feel like this adds unnecessary boilerplate to the user of the library. It's not a hard thing to add a check in the rendering logic to allow for passing in the full Ref object. In fact I would argue that it's a better practice because it keeps it more declarative, and the intent is very clear.

This way, I can write 
``` tsx
class MyComponent extends React.Component {
  private target = React.createRef<HTMLAnchorElement>();

  private container = React.createRef<HTMLDivElement>();

  public render() {
      return (
          <div ref={this.containerObj}>
              <p>Somewhere in here is a <a href="#" ref={this.refObj}>tooltip</a>.</p>
              <UncontrolledTooltip container={this.container} target={this.target}>
                  Hello world!
              </UncontrolledTooltip>
          </div>
      );
  }
}
```

[...]

PR #1200, opened 2018-09-04
feat(Popover/Tooltip): Implented usage of react 16.3 RefObject as target

josh-degraw:
- [x] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

R.E. issue #1198

This pull request adds support for using React 16.3 `createRef` object as the `target` and `container` props for `Tooltip` and `Popover` components. I have updated the type definitions in [my fork](https://github.com/josh-degraw/DefinitelyTyped/tree/reactstrap-josh-degraw/types/reactstrap) so that those can be updated as well.

I added tests to cover my changes, but testing is definitely _not_ my forte, so if I did something wrong there or I'm missing something, please let me know and I'll fix / update it ASAP.

### DIFF
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -2,20 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import PopperContent from './PopperContent';
-import { getTarget, DOMElement, mapToCssModules, omit, PopperPlacements } from './utils';
+import { getTarget, mapToCssModules, omit, PopperPlacements, targetPropType } from './utils';
 
 const propTypes = {
   placement: PropTypes.oneOf(PopperPlacements),
-  target: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.func,
-    DOMElement,
-  ]).isRequired,
-  container: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.func,
-    DOMElement,
-  ]),
+  target: targetPropType.isRequired,
+  container: targetPropType,
   boundariesElement: PropTypes.string,
   isOpen: PropTypes.bool,
   disabled: PropTypes.bool,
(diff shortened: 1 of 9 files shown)


TheSharpieOne:
I like where this is going, but it points out another issue that probably needs to be solved as well. Currently we only get the target in `componentDidMount`. Chances are `current` will still be `null` when that runs and even if it updates within the ref the target will never update.

We probably need to grab the target every render. This would help with other cases as well. When the target prop changes, it should trigger a rerender and it should target the new target.

We will not be able to determine when `current` changed via an update check as the ref object will never change and never trigger an update by itself even if the `current` property chances within the ref.

virgofx:
Yeah .. i wasn't sure this was going to work originally; however, I like the concept of being able to supply the refs directly -- wasn't sure if react-popper supported it? Either way, we should just make sure we cover all cases in a thorough test suite which would ensure we don't break any existing apps moving forward.

josh-degraw:
@TheSharpieOne I wouldn't worry too much about it being null by`componentDidMount`.

Straight from [the docs](https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element):

> React will assign the current property with the DOM element when the component mounts, and assign it back to null when it unmounts. ref updates happen before `componentDidMount` or `componentDidUpdate` lifecycle hooks.

Also, when using the `ref` object I don't think we'll need to worry that much about knowing if `current` changed. React seems to be pretty smart about it. I've made a really simple [tester app](https://github.com/josh-degraw/reactstrap-ref-tester) using a build of my changes just to show that the functionality works. It switches where the `ref` points to after 5 seconds and the tooltip rerenders just fine.

ashleydw:
Is there a PR for the type definitions? The latest types are missing the `TargetElement` from your fork.

https://github.com/josh-degraw/DefinitelyTyped/commit/f1980369c66ab2a199c35a2d8adcfdbd8bfe98c6
https://github.com/DefinitelyTyped/DefinitelyTyped/commit/03041459989ac55f3031e9bef0492482a5575f91

Issue #1182, opened 2018-08-18
Feature Request - Support Close Icon Utility


dillonredding:
Bootstrap has the [close icon utility](https://getbootstrap.com/docs/4.1/utilities/close-icon), which is just a specialized button.

It would be nice to have that supported through, perhaps, the `Button` component.

dillonredding:
I feel like option 1 is more inline with Bootstrap's design and the approach that reactstrap takes.

If you allow for option 2, I would think that `CloseIcon` would handle the `span`/`aria-hidden` for you, so you could just do

    <CloseIcon>&times;</CloseIcon>

Perhaps `&times;` is the default, allowing you to use `<CloseIcon/>` in most cases. And maybe all that essentially translates to option 1.

median-man:
I agree that option 1 is more consistent. I've created #1206 to add the `close` prop to `<Button />`.

PR #1206, opened 2018-09-08
feat(Button): add close icon support (reactstrap/reactstrap#1182)

median-man:
This PR addresses #1182.

Add support for close icon utility from Bootstrap 4.
This commit adds an optional boolean property, 'close', to the
Button component. Applying the close prop will render the Close
icon utility from Bootstrap 4.

Update documentation on the Button component to demonstrate usage
of the close prop.

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

This is my first time contributing. I welcome any constructive feedback you have to offer if you are reviewing this  PR.


TheSharpieOne:
Sorry for the delay, but what do you think about defaulting `children` to `<span aria-hidden>&times;</span>` when `close` is `true`? That way people would be able to do `<Button close />` to get what is probably the most common usages which includes `×`. Providing children would of course show the children instead.

median-man:
I think that is a great idea! I'll update my pr shortly.
> Sorry for the delay, but what do you think about defaulting `children` to `<span aria-hidden>&times;</span>` when `close` is `true`? That way people would be able to do `<Button close />` to get what is probably the most common usages which includes `×`. Providing children would of course show the children instead.

median-man:
@TheSharpieOne: I do have a question about version control. How should I handle this change? Should I close the pr. Rebase the commit for the feature and include the default child behavior in that commit before creating a new pr? Or, should I just add the default behavior as a fix on a separate commit?

median-man:
Pending checks, I've added the default children to the commit.

Issue #1216, opened 2018-09-10
Expected subtree parent to be a mounted class component


Mustafazubair:
- components: `Fade`
- reactstrap version `#6.0.0`

**ISSUE  ?**
![image](https://user-images.githubusercontent.com/43081391/45328957-f9d00b00-b512-11e8-8f27-0f850a9beab7.png)

**CODE**
![image](https://user-images.githubusercontent.com/43081391/45328972-14a27f80-b513-11e8-9be0-4fb657f90d76.png)

Happening also for collapse component.

Issue #1201, opened 2018-09-04
PopperContent doesn't pass context to child components


josh-degraw:
### Issue description
When using the react 16.3 context api with the Popover component, since the implementation of `PopperContent` uses `unstable_renderSubtreeIntoContainer`, the context cannot be passed down accurately. However, if `ReactDOM.createPortal` was used instead, the context would be passed down automatically to all descendant components.

- components: `Popover`: `PopperContent`
- reactstrap version `6.4.0`
- import method `es` 
- react version `16.4.2`
- bootstrap version `4.1.1`

### What is happening?
Context is not passed to child components when using a `Popover`.

### What should be happening?

The `PopperContent` and all subchildren of a `Context.Provider` should be passed the context value.

### Steps to reproduce issue

View [this demo repository](https://github.com/josh-degraw/portal-demo) and run the code to see an example.

### Code

  
[Direct link to demo file](https://github.com/josh-degraw/portal-demo/blob/master/src/PortalTester.js).

PR #1254, opened 2018-10-08
fix(PopperContent): Use create portal instead of unstable_renderSubtr…

Imundy:
…eeIntoContainer

This update *should* have no visible effects.

- [x] Bug fix  
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

I am getting an obscure error occasionally with `unstable_renderSubtreeIntoContainer` on react versions newer than 15.4.1 (possibly older versions too) where react *thinks* the container is not mounted. It is the same error as https://github.com/reactstrap/reactstrap/issues/1216
I unfortunately am having trouble creating a minimally reproducible example 😕

### DIFF
--- a/src/__tests__/Tooltip.spec.js
+++ b/src/__tests__/Tooltip.spec.js
@@ -83,7 +83,7 @@ describe('Tooltip', () => {
 
     const tooltips = document.getElementsByClassName('tooltip');
 
-    expect(wrapper.find('.tooltip.show').hostNodes().length).toBe(0);
+    expect(wrapper.find('.tooltip.show').hostNodes().length).toBe(1);
     expect(tooltips.length).toBe(1);
     expect(tooltips[0].textContent).toBe('Tooltip Content');
     wrapper.detach();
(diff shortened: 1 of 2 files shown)


src/__tests__/Tooltip.spec.js
```diff
@@ -83,7 +83,7 @@ describe('Tooltip', () => {
 
     const tooltips = document.getElementsByClassName('tooltip');
 
-    expect(wrapper.find('.tooltip.show').hostNodes().length).toBe(0);
+    expect(wrapper.find('.tooltip.show').hostNodes().length).toBe(1);
```
Imundy: I *think* this is actually the desired behavior. before the PopperContent was creating a div manually and then mounting into that div. Now it's wrapping the content in the div while it creates the portal which means we do have a host node, but the dom ends up being the same as before.

kyriiherman:
@Imundy actually, there is an open issue (but I'm not sure) https://github.com/reactstrap/reactstrap/issues/1201

EugeneChu:
Tested the version of `reactstrap` pointed to this commit in our project. We still have problems with `PopperContent` and passing elements with `Context` as it's children

eugenechu:
⬆️ is so much needed, @TheSharpieOne we would so much appreciate a minor release with this thing!
Thanks in advance

verbold:
+1 on a minor release. This would allow reactstrap to be compatible with the [v6.0.0](https://github.com/reduxjs/react-redux/releases) react-redux release.

Issue #1257, opened 2018-10-10
UncontrolledCollapse error on IE11. Object doesn't support property or method 'forEach'


themetis:
### Issue description
- components: `UncontrolledCollapse`
- reactstrap version `6.5.0`
- import method `es`
- react version `16.5.2`
- bootstrap version `4.1.3`

### What is happening?

I've got an error when it is rendered on IE11.
The error is  "Object doesn't support property or method 'forEach'"

### Code

https://stackblitz.com/edit/reactstrap-uncontrolled-collapse-llryd4

gergely-nagy:
Merged via #1262 ! Will be out in next release.

z1m1n:
@gergely-nagy when will be next release?

juanmiguelbesada:
Already fixed and merged. You should close this.

PR #1258, opened 2018-10-11
fix(issue #1257)- IE forEach error

jeff-nz:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

This fix will shim the object array data preventing it to error out on IE and hopefully other browsers.

Fixes issue 1257
https://github.com/reactstrap/reactstrap/issues/1257

In addition I have also found a link tag with a comma that also causes to error out on IE

### DIFF
--- a/docs/lib/app.js
+++ b/docs/lib/app.js
@@ -65,7 +65,7 @@ export default (locals, callback) => {
           <link rel=icon href=/assets/favicon.ico>
           <link rel="stylesheet" href="/assets/style.css"/>
           <link rel="stylesheet" href="/assets/docs.css"/>
-          <link rel="stylesheet", href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
+          <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
         </head>
         <body>
           <div id="app">${body}</div>
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
   "homepage": "https://github.com/reactstrap/reactstrap#readme",
   "dependencies": {
     "classnames": "^2.2.3",
+    "lodash.foreach": "^4.5.0",
     "lodash.isfunction": "^3.0.9",
     "lodash.isobject": "^3.0.2",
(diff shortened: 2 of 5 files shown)


jeff-nz:
Feel free to merge the fix whenever you guys ready.

TheSharpieOne:
IE11 supports both `Object.keys` and `Array.prototype.forEach`: https://caniuse.com/#feat=es5

I believe the issue is not with the usage of those functions specifically, but that in a particular case, forEach is used on something that is not an array. It looks like `els` can be an array _or_ a `NodeList`, and a quick glance at https://developer.mozilla.org/en-US/docs/Web/API/NodeList indicates that [IE doesn't support `forEach` on `NodeList`](https://developer.mozilla.org/en-US/docs/Web/API/NodeList#Browser_compatibility).

It even go into detail with an example:

> Although NodeList is not an Array, it is possible to iterate over it with forEach(). It can also be converted to a real Array using Array.from().
> 
> However, some older browsers have not implemented NodeList.forEach() nor Array.from(). This can be circumvented by using Array.prototype.forEach() [...]

```js
var list = document.querySelectorAll('input[type=checkbox]');
Array.prototype.forEach.call(list, function (checkbox) {
  checkbox.checked = true;
});
```

Probably need to turn https://github.com/reactstrap/reactstrap/blob/0dc2acda96913db56418a880b9ebd807a9c05076/src/utils.js#L275 and https://github.com/reactstrap/reactstrap/blob/0dc2acda96913db56418a880b9ebd807a9c05076/src/utils.js#L281 into

```js
Array.prototype.forEach.call(els, (el) => {
```

jeff-nz:
i see thanks. I'll do that now.

jeff-nz:
Yes, Array.prototype.forEach.call() works but the data we are passing is still causing it to break on IE 11.

Unlike using loadash's foreach 
- it checks and identifies the type of data > rebuilds if necessary > loops through
- using this will also allow the system to support older IE browsers that are commonly used by the banks.

Please let me know what you think. I only see benefits using the lodash foreach but I can go in further cleaning the data before running Array.prototype.forEach.call().

jeff-nz:
Does reactstrap have a list of browser it supports?

PR #1270, opened 2018-10-18
fix(FormGroup): remove `.position-relative`

### LINKED ISSUE #1269: Remove `.position-relative` from FormGroup
PR #1074 introduced the option to have form feedback as a tooltip. That's great!  
It also added the class `.position-relative` to `FormGroup`.  
As bootstrap states

> Be sure to have a parent with position: relative on it for tooltip positioning.
> http://getbootstrap.com/docs/4.1/components/forms/#tooltips

User should be responsible of adding that class

iamandrewluca:
PR #1074 introduced the option to have form feedback as a tooltip. That's great!  
It also added the class `.position-relative` to `FormGroup`.  
As bootstrap states

> Be sure to have a parent with position: relative on it for tooltip positioning.
> http://getbootstrap.com/docs/4.1/components/forms/#tooltips

User should be responsible of adding that class

Resolves #1269

### DIFF
--- a/src/FormGroup.js
+++ b/src/FormGroup.js
@@ -32,7 +32,6 @@ const FormGroup = (props) => {
 
   const classes = mapToCssModules(classNames(
     className,
-    'position-relative',
     row ? 'row' : false,
     check ? 'form-check' : 'form-group',
     check && inline ? 'form-check-inline' : false,


TheSharpieOne:
I'm not too sure how semver would apply. I think introducing it was a breaking change and fixing it reverts that breaking change which is allowed without a major bump.

PR #1271, opened 2018-10-19
fix(Modal): only show backdrop when prop is true

### LINKED ISSUE #1267: Modal ignores backdrop={false} when fade={false}
- components: `Modal`
- reactstrap version `#6.5.0`
- react version `#16.5.2`
- bootstrap version `#4.1.3`

### What is happening?

Modal ignores the `backdrop={false}` prop when `fade={false}`, and renders backdrop.
The backdrop is removed when `fade={true}`

### What should be happening?

The backdrop should never be present when `backdrop={false}`
We use this combo of props as we always disable fade, but sometimes disable backdrop.

### Code

https://stackblitz.com/edit/reactstrap-nepff6?file=Example.js

(Stackblitz is giving trouble intermittently) 

```
import React, { Component } from 'react';
import { Modal } from 'reactstrap';

class Example extends Component {
  render() {
    return (
      <div>
        <Modal isOpen backdrop={false} fade={false}>Hello World!</Modal>
      </div>
    )
  }
}

export default Example;
```

bestguy:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

Change introduced in v6.4.0 causes Modal backdrop to incorrectly render when `backdrop={false}`, if `fade={false}`.  This change will only render backdrop if `backdrop` is true.

Fixes #1267

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -348,14 +348,14 @@ class Modal extends React.Component {
         timeout: hasTransition ? this.props.backdropTransition.timeout : 0,
       };
 
-      const Backdrop = hasTransition ?
+      const Backdrop = backdrop && (hasTransition ?
         (<Fade
           {...backdropTransition}
           in={isOpen && !!backdrop}
           cssModule={cssModule}
           className={mapToCssModules(classNames('modal-backdrop', backdropClassName), cssModule)}
         />)
-        : <div className={mapToCssModules(classNames('modal-backdrop', 'show', backdropClassName), cssModule)} />;
+        : <div className={mapToCssModules(classNames('modal-backdrop', 'show', backdropClassName), cssModule)} />);
 
       return (
         <Portal node={this._element}>
@@ -373,7 +373,7 @@ class Modal extends React.Component {
               {external}
               {this.renderModalDialog()}
             </Fade>
-            {backdrop && Backdrop}
+            {Backdrop}
           </div>
         </Portal>
       );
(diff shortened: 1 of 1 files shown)


jeff-nz:
**Suggestion:**
The fix should probably be done in Backdrop.

bestguy:
Sure, was avoiding nested ternary but let me know if updated works.

jeff-nz:
I wasn't implying that you do it that way but your fix should do the job. You probably just need to format it better since it is now a nested ternary statement.

Nested ternary is highly not recommended on other programming language coding practices. Totally agree with that.

However on JavaScript, this is acceptable as long as you format it well. 

Here are some articles:
https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340
https://hackernoon.com/the-untapped-potential-of-nested-ternaries-2cb98079b634

I hope it helps.

jeff-nz:
> though it looks like nested ternary is not allowed with your eslint config, was getting a warning

I'm pretty sure the rule "no-nested-ternary" is not set on this repo, which is good. This is why your commit has passed the CI test.

ESLint recommends the rule no-nested-ternary being off.

Here's the link to their list of rule recommended:
https://github.com/eslint/eslint/blob/master/conf/eslint-recommended.js

I hope it helps. I will now leave you to the member who can merge this for you. Your commit looks fine.

not getting any warning about nested ternaries

bestguy:
Thanks, yes sorry I meant nested ternary in this form:

```javascript
const Backdrop = !backdrop ? null 
    : hasTransition ? <Fade ... />
    : <div ... />;
```

I receive this eslint error
`/reactstrap/src/Modal.js
  351:24  error  Do not nest ternary expressions                no-nested-ternary`

Issue #1274, opened 2018-10-25
Add possibility to set custom node where to mount modal


dmitrykrylov:
### Issue description

In some cases you need to mount modals in custom element instead of body.

### What is happening?

I have a react-app with some kind of incapsulated bootstrap styles that mounts into a host-site with its own styles. Therefore when I use modal it mounts in the body of the host-site and loses it's styles.

### What should be happening?

Provide a mountNode prop like it's done in semantic-ui-react

dmitrykrylov:
I tried to implement a feature myself, but I can't get it working. Devserver doesn't seem to work
https://github.com/dmitrykrylov/reactstrap/commit/8490a73aa6d831db50c967977278163c4c5e55e2

jamesremuscat:
Just come here looking for a solution to a similar problem; this looks like a duplicate of #1011 to me, which in turn was part of #113 that got missed before it got closed.

PR #1275, opened 2018-10-25
Feature/modal mount node

dmitrykrylov:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

Relates to #1274
Should I add any unit tests? I'm not sure how to better test this feature

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -77,7 +77,6 @@ const defaultProps = {
     mountOnEnter: true,
     timeout: TransitionTimeouts.Fade, // uses standard fade transition
   },
-  mountNode: document.body,
 };
 
 class Modal extends React.Component {
@@ -237,7 +236,7 @@ class Modal extends React.Component {
   }
 
   init() {
-    const { mountNode } = this.props;
+    const mountNode = this.props.mountNode || document.body;
 
     try {
       this._triggeringElement = document.activeElement;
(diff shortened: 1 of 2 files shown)


src/Modal.js
```diff
@@ -54,6 +54,7 @@ const propTypes = {
     PropTypes.string,
     PropTypes.func,
   ]),
+  mountNode: PropTypes.any,
```
TheSharpieOne: We should probably limit this like we do for the `target`s in tooltip/popover and allow for a DOMElement, function (which will return a DOMElement), or a string (which can be used to find a DOMElement).

There is a util which handles all of those cases, returning the DOMElement if found.

Probably should call it `container` to match similar functionality/props elsewhere.

src/Modal.js
```diff
@@ -235,6 +236,8 @@ class Modal extends React.Component {
   }
 
   init() {
+    const mountNode = this.props.mountNode || document.body;
```
TheSharpieOne: could probably use `defaultProps` to default it to `document.body`

dmitrykrylov: Look at the my first commit. I used `defaultProps` but when I run `npm start` it threw an error where `document` was `undefined`.

dmitrykrylov: Probably webpack-dev-server works in kind of SSR mode?

TheSharpieOne: Yeah, I suppose that will happen on the server when document is not defined. What you have here should be fine then.

virgofx:
This won't work with the forthcoming changes as well on server side. Everything in the init() will need to move to client side rendering as this breaks when {open} = true on SSR. Please do not merge this.

iamandrewluca:
Fixed in #1817

PR #1296, opened 2018-11-15
feat(Table): add innerRef prop

amiuhle:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

`innerRef` prop for `Table`


amiuhle:
Thanks for the quick merge. Can we get a `6.5.1` release soon please?

I tried installing from GitHub, but that doesn't work because it needs to be built. I don't know how other packages having a build step handle this, if you want I can open an issue to discuss building the source (if it's not there yet) in the `install` script.

PR #1298, opened 2018-11-16
fix(CardTitle,CardSubtitle): div as default tag

### LINKED ISSUE #1297: Do not use h* in Card*#defaultProps#tag?
The `Card*defaultProp#tag` has been in place from the start since #44 (updated in #794) but I don't think `reactstrap` should conflate these any more.

1. HTML heading elements (`<h1>`-`<h6>`) are for document structure
1. BS4 `card-*` classes (`card-title`, `.card-subtitle`, ...) are for applying presentational style (ref [BS4 docs](https://getbootstrap.com/docs/4.1/components/card/#titles-text-and-links)) and **do not** suppose heading levels

Rationale:

> My project uses lots of `Cards` with `CardTitles` but most do not have any structural meaning in the HTML document (and if they did, they wouldn't be semantic h5s). So I pass a `tag` prop in _every_ one of my CardTitles 😢.

Resolution? Maybe of the following:

1. Do not suppose structure. Use a generic container (`div`) by default instead of a heading element

1. Provide an API to map defaultTags to components á la `utils#[setGlobal/mapTo]CssModule` or context or something

I can easily imagine projects where `Cards` _are_ highly correlated to document structure, in which case `CardTitle`/`CardSubtitle` defaultTags of `h<N>` would be really helpful. But even then, I'd want to be able to set the `h*` levels b/c its probably not `h5`/`h6`. @TheSharpieOne if you have an opinion here I'm happy to start working on a PR

eckdanny-osi:
Closes #1297

### DIFF
--- a/src/CardSubtitle.js
+++ b/src/CardSubtitle.js
@@ -10,7 +10,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  tag: 'h6'
+  tag: 'div'
 };
 
 const CardSubtitle = (props) => {
--- a/src/CardTitle.js
+++ b/src/CardTitle.js
@@ -10,7 +10,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  tag: 'h5'
+  tag: 'div'
 };
 
 const CardTitle = (props) => {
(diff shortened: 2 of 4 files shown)


eckdanny-osi:
note:

does not address similar defaultProp tag headings. Present in:

- `PopoverHeader.js#13`,
- `ModalHeader.js#19`, 
- `ListGroupItemHeading.js#13`

PR #1311, opened 2018-11-29
fix(CarouselIndicators): li key values generate from provided item va…

### LINKED ISSUE #1310: [CarouselIndicators] Warning: Encountered two children with the same key
- components: `CarouselIndicators`
- reactstrap version `^6.1.0`
- react version `^16.4.1`
- bootstrap version `^4.1.0`

### What is happening?

When you pass your data to 
```
<CarouselIndicators items={items} activeIndex={activeIndex} onClickHandler={this.goToIndex} />
```

It provoke

screenshot 2018-11-28 at 15 46 26

### What should be happening?

CarouselIndicators shouldn't rely on dataset implementation (which is not using src)

### Error message in console

screenshot 2018-11-28 at 15 46 26

### Code

```
<CarouselIndicators items={items} activeIndex={activeIndex} onClickHandler={this.goToIndex} />
```

wlockiv:
…lues

Generation of key items for `li` components in `CarouselIndicators` is a
concatenation of specific, optional values (`src`, `caption`, and
`altText`) in each item of the `items` object array. Lacking these
key/value pairs will cause each `li` element to have a key along the
lines of `undefinedundefinedundefined`, resulting two or more of the
`li`'s to be identical and causing an error to log in the console.
While either including any one of the specified keys with a unique
value, or adding a unique `key` field in each of the items objects will
remedy this issue, that solution is not well documented. The proposed
solution is a transparent fix: preventing the aforementioned error while
requiring no change of code for the end-user. The new `key` prop value
for each `li` will be a concatenation of all the values of its
respective item object.

Closes #1310

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/CarouselIndicators.js
+++ b/src/CarouselIndicators.js
@@ -13,7 +13,7 @@ const CarouselIndicators = (props) => {
     ), cssModule);
     return (
       <li
-        key={`${Object.values(item).join('')}`}
+        key={`${item.key || Object.values(item).join('')}`}
         onClick={(e) => {
           e.preventDefault();
           onClickHandler(idx);
--- a/src/CarouselIndicators.js
+++ b/src/CarouselIndicators.js
@@ -13,7 +13,7 @@ const CarouselIndicators = (props) => {
     ), cssModule);
     return (
       <li
-        key={`${item.key || item.src}${item.caption}${item.altText}`}
+        key={`${Object.values(item).join('')}`}
         onClick={(e) => {
           e.preventDefault();
           onClickHandler(idx);


src/CarouselIndicators.js
```diff
@@ -13,7 +13,7 @@ const CarouselIndicators = (props) => {
     ), cssModule);
     return (
       <li
-        key={`${item.key || item.src}${item.caption}${item.altText}`}
+        key={`${Object.values(item).join('')}`}
```
TheSharpieOne: It would still be good to explicitly allow the key to be the key specified if specified. In the case where other values change but the user wants to ensure its the same instance of the component, an explicit key would be the only way to ensure that.

wlockiv: 100% Agree. What about something like this?

```suggestion
        key={`${item.key || Object.values(item).join('')}`}
```

TheSharpieOne: Yes, exactly.

Issue #1336, opened 2018-12-24
cant find doc on forms


crap-coder:
i cant visit the "form" component's page on the doc . somehow the page is down

saleem-git:
Its been down for days, can someone please look into this, project deadline closing in.

Xachman:
Xachman@11d9f1a

The issue is with the plaintext attribute. I dont have time to write a unit test for this right now so I cannot create a PR for this change now. But if you are working locally and would like to read the docs you can either remove the plaintext input or use this update.

TheSharpieOne:
Fixed.

Issue #1332, opened 2018-12-21
Forms documentation component is broken. [out of scope]


unfaceit:
- components: `Form`
- documentation website

### What is happening?

In your documentation, when you go to [https://reactstrap.github.io/components/form/](FORM), your react application breaks. Page reload to a different tab required, to make it render again. 

### What should be happening?

We should be able to access FORM component documentation.

### Steps to reproduce issue

1. Go to your docs - [https://reactstrap.github.io/components/alerts/](url)
2. Click on **Form**

### Error message in console

```
Error: Minified React error #137; 
```

bwanicur:
When I try to run `npm start` locally, I get the following error:

` ERROR in Invariant Violation: input is a void element tag and must neither have 'children' nor use 'dangerouslySetInnerHTML'.`.

I'm new to webpack and I could not get a meaningful stack trace, or I would post it.

Xachman:
https://github.com/Xachman/reactstrap/commit/11d9f1a68a90e02f6f7c15de7fc1b2adb7113e75

The issue is with the plaintext attribute.  I dont have time to write a unit test for this right now so I cannot create a PR for this change now.  But if you are working locally and would like to read the docs you can either remove the plaintext input or use this update.

PR #1339, opened 2018-12-28
Plaintext fix

Xachman:
- [x ] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [ x] There is an open issue which this change addresses
- [ x] I have read the **[CONTRIBUTING](./CONTRIBUTING.md)** document.
- [x ] My code follows the code style of this project.
- [x ] My change requires a change to the documentation.
- - [x ] I have updated the documentation accordingly.
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.

https://github.com/reactstrap/reactstrap/issues/1336

https://github.com/reactstrap/reactstrap/issues/1332

### DIFF
--- a/docs/lib/examples/InputType.js
+++ b/docs/lib/examples/InputType.js
@@ -7,7 +7,8 @@ export default class Example extends React.Component {
       <Form>
         <FormGroup>
           <Label for="exampleEmail">Plain Text (Static)</Label>
-          <Input plaintext>Some plain text/ static value</Input>
+          <Input plaintext value="Some plain text/ static value" />
+          <Input plaintext value="Some plain text/ static value">Children override value attribute</Input>
         </FormGroup>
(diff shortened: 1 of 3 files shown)


TheSharpieOne:
Closing in favor of #1335
If someone wanted to use `tag="p"` to have it output a `<p>`, this change would not allow children.

Xachman:
Alright we should fix some of the tests that include children in inputs as well.  Since these tests all throw the same error.
https://github.com/reactstrap/reactstrap/blob/master/src/__tests__/Input.spec.js#L7

Also this test passed so i assumed that the tag attribute would work.  If this is not the case then test coverage for your case should be added.
https://github.com/reactstrap/reactstrap/blob/master/src/__tests__/Input.spec.js#L43

Issue #1349, opened 2019-01-05
Support backward-compatible popover behavior


FelixHenninger:
- components: `Popover`
- reactstrap version `#7.0.x`
- import method `es`
- react version `#16.4.1`
- bootstrap version `#16.4.1`

### What is happening?

In earlier versions, popovers were dismissed if users clicked outside of the current popover (but not if the click was inside the popover). With the new version, popovers are either dismissed manually via their toggle (`trigger="click"`) or on the next click (with `trigger="focus"`), wherever it lands (even if on the popover itself).

### What should be happening?

This behavior appears to match the [bootstrap defaults](https://getbootstrap.com/docs/4.2/components/popovers/#dismiss-on-next-click), therefore I'm guessing that it is consistent. Thus, this is probably not a bug as much as a humble request for pointers or hints toward a backwards-compatible solution.

### Steps to reproduce issue

The changes are visible when comparing versions [6.5.0](https://reactstrap-ewpznv.stackblitz.io/) ([code](https://stackblitz.com/edit/reactstrap-ewpznv?file=Example.js)) and [7.0.2](https://reactstrap-vgbr1w.stackblitz.io/) ([code](https://stackblitz.com/edit/reactstrap-vgbr1w?file=Example.js)) by clicking on multiple popover toggles in sequence.
In the first case, only one popover is visible at a time, while in the second, several popovers compete for screen estate.

### Error message in console

(none)

### Code

(see above)

----

Thanks so much for your effort with this library! I've found it a joy to work with, and would like to apologize for reporting a semi-issue here. I'm guessing that I've missed a flag I should be setting, and would be happy to extend the documentation if that helps. Sorry for the trouble, and thanks in advance for your help!

TheSharpieOne:
Yes, these changes were made to be consistent with bootstrap, but I do see that the previous behavior was also desired. Also, some how that commit message got lost and the breaking change was not included on the list in the change log making this even more problematic.
`trigger="focus"` was suppose to mimic what was previously there, but it still doesn't allow you to click on the popover as the button/triggering element would lose focus.

I think it would be a good thing to bring back as an opt-in (maybe via another trigger value (not sure what to call it, please suggest) or a prop in combination with the "focus" trigger).

septs:
```html
trigger="classic"
```

The is `reactstrap@6` classic behavior.

kinke:
[...]

PR #1360, opened 2019-01-09
Tackle issue #1349 (backward-compatible Popover behavior)

kinke:
Via `<Popover trigger="manual" isInteractive />`.

Changed semantics for `trigger="manual"`: hide popover when clicking outside target element *and*, depending on the new `isInteractive` prop, outside the popover itself.

Would probably need some tests & documentation.

### DIFF
--- a/src/TooltipPopoverWrapper.js
+++ b/src/TooltipPopoverWrapper.js
@@ -36,7 +36,8 @@ export const propTypes = {
     PropTypes.string,
     PropTypes.object
   ]),
-  trigger: PropTypes.string
+  trigger: PropTypes.string,
+  isInteractive: PropTypes.bool,
 };
 
 const DEFAULT_DELAYS = {
@@ -51,7 +52,12 @@ const defaultProps = {
   delay: DEFAULT_DELAYS,
   toggle: function () {},
   trigger: 'click',
+  isInteractive: false,
 };
+
+function isInDOMSubtree(element, subtreeRoot) {
+  return subtreeRoot && (element === subtreeRoot || subtreeRoot.contains(element));
+}
 
 class TooltipPopoverWrapper extends React.Component {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


TheSharpieOne:
"manual" (based on the definition from bootstrap itself) seems like it would mean that the user of the library is controlling the state of `isOpen` themselves outside of the anything popover is doing, I don't think it would make sense to use that in this context and it seems like it work diverge from bootstrap.
It would be nicer to have something like `trigger="classic"` or `trigger="reactstrap"` to invoke the previous behavior.

kinke:
> It would be nicer to have something like trigger="classic" or trigger="reactstrap" to invoke the previous behavior.

Pre-v7, there was no trigger, so `isOpen` was controlled outside. But the popover toggled itself via a click anywhere except its target's subtree and its own. Not a real trigger IMHO, but I found something like `trigger="manual" autoHide={true}` too inconvenient; something like `trigger="legacy-auto-hide"` may not be that bad (and may make even sense in combination with other triggers like `click` => show on target click, but don't hide again on 2nd target click, instead require a click somewhere else).

Btw, the `isInteractive` prop is 'inspired' by [react-tippy](https://github.com/tvkhoa/react-tippy)'s `interactive`.

TheSharpieOne:
`trigger="legacy-auto-hide"` or even just `trigger="legacy"` sounds good to me.

kinke:
Alright, reworked so that `<Popover trigger="legacy-auto-hide" isInteractive delay={0} />` should restore full legacy behavior. A combined `trigger="click legacy-auto-hide" [isInteractive]` should work as well and change which clicks hide the popover.

TheSharpieOne:
Awesome, thanks for tackling this. If you can, please add some tests and update the documentation to point out this trigger which is outside of bootstrap's spec. Or, I can merge it and add the tests and update the docs.

kinke:
I'd prefer the latter ;), as I'd have to look up all the needed stuff. - Thanks!

Issue #363, opened 2017-03-15
Ability to have Fade animation on tooltips and popover


JohanBengtsson:
It would be great to have the ability to add a fade animation to Reactstrap Tooltip and Popover, similar to bootstrap tooltip option `animation`

![image](https://cloud.githubusercontent.com/assets/3408400/23955758/2bac69c2-099b-11e7-9d39-fae6117351ae.png)

Are there any plans of introducing this? Thanks!

KamiGim:
I'm also waiting 
Thanks

Geczy:
Yes!

TheSharpieOne:
@amikhalev might you be able to help with this on the v5 branch? From the looks of it, tooltips and popovers should always render the PopperContent and the PopperContent can use  the `Fade` around the [`ReactPopper`](https://github.com/reactstrap/reactstrap/blob/9b80d11447320ef4cb64d2aa97b2cc95672b686e/src/PopperContent.js#L159-L162) similar to this example: https://github.com/souporserious/react-popper/blob/ef520ee1213f30d3af3f10c6f5b4a04a888deb73/example/index.jsx#L170-L188
what do you think?

amikhalev:
Yeah, I think I could do that. Might take me a few days to get to it though

Geczy:
Any update? @TheSharpieOne @amikhalev

DorienMay:
I would love to have this feature too.

PR #1364, opened 2019-01-12
feat(Popper): Add fade support for ToolTip and Popover (#363)

esoh:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

Hi, this is my first PR. I hope I'm doing everything right! This is to address issue #363.

### Changes in this PR
- default behavior for PopperContent set to fade in and out, toggleable with prop `fade={false}` in Tooltip/Popover
- default hide delay changed from 250 to 0 since now it fades out as consistent with bootstrap
- separated the popperClasses (e.g. `tooltip show`) from any extra classNames as to apply these classes to the ReactPopper child component for display purposes, and the extra classes will be applied to the root Fade component. I was not sure if this is the intended behavior, or if the extra classNames should be applied to ReactPopper instead.
- updated failing TooltipPopoverWrapper isOpen toggle test to account for the 150ms fade animation and added a test with `fade={false}`

### DIFF
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -11,8 +11,7 @@ const defaultProps = {
 const Popover = (props) => {
   const popperClasses = classNames(
     'popover',
-    'show',
-    props.className
+    'show'
   );
 
   const classes = classNames(
@@ -24,7 +23,7 @@ const Popover = (props) => {
   return (
     <TooltipPopoverWrapper
       {...props}
-      className={popperClasses}
+      popperClassName={popperClasses}
       innerClassName={classes}
     />
   );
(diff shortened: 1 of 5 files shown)


TheSharpieOne:
Thank you for tackling this issue. It's something that has been asked for several times in the past but keeps getting overlooked.

I did notice a few things. First, all of the tooltips and popovers always appear as they have placement of bottom, no matter what the placement value is (See https://deploy-preview-1364--reactstrap.netlify.com/components/tooltips/#Tooltips-List for example)
Also, with `autohide={false}` on the tooltip, it should allow the tooltip to be remain visible while hovering over the tooltip (to select the text or something). While it does do is when moving very fast, once the fade starts, hovering over the tooltip doesn't stop it from fading. I feel as if the fade should stop and the tooltip should stay open (at 100 opacity) in this case, if possible.

esoh:
Thank you for pointing out the issues. I believe I've fixed the ones that you've mentioned, but please let me know if I'm missing anything.

TheSharpieOne:
travis is messed up, closing and reopening to trigger a build

AlexanderSribny:
I have a problem. I've install reactstrap from npm, but there is no changes from this commit. The lastest version of reactstrap is 7.1.0., am I right? So why this changes are not in installation package? Or I maked some mistakes?

Issue #1346, opened 2019-01-02
Add support for Toast messages (Bootstrap 4.2.1)


cvalmonte:
Please add support for toast messages which were recently added in version 4.2.1:

https://getbootstrap.com/docs/4.2/components/toasts/

bpas247:
I might open a PR if I have enough time to get around to doing this. I'm guessing we would handle showing and hiding toast messages via state?

TheSharpieOne:
I imagine they would work much like `Alert`s do. They are very similar in may way, just that these toasts will have more content.
I would let the use define all of the content (such as the "11 mins ago" seen in the examples), the user can pull in a library like react-timeago (or similar) his/herself if needed rather than supporting that feature and bloating reactstrap.

bpas247:
Would it be useful to use a library like react-timeago in the Toast example? Or just reference how you would implement it via a library?

TheSharpieOne:
An example in the docs could be good, make sure when the package is installed, it's installed as a devDev.

chez14:
Any news with this new feature? 

Or just quick question, when will the docs for this feature (along with other new BS release like Switch and etc) that have been implemented be released?

TheSharpieOne:
switch and spinner have already been released (a week ago) and the docs were updated when the PRs for those features were merged (switch is on the forms page with the custom inputs and spinner has its own page).

bpas247:
> Any news with this new feature?

I haven't had the chance to implement this yet, I've been quite busy with school.

I'll open a PR once I get the first revision done.

chez14:
Oh ok, got it. Thanks everyone xD

charlex:
I started work on this in case @bpas247 hasn't yet started. Didn't want to step on toes, but I would love to see this out in the world:

https://github.com/charlex/reactstrap/tree/toasts

It's still a work in progress, but I could maybe use some suggestions on how to handle the little icon bug in the top left (and in general). Should it be a prop like `icon="warning"`?

 ```jsx
<Toast>
  <ToastHeader icon="warning">Reactstrap</ToastHeader>
  <ToastBody>Lorem ipsum</ToastBody>
</Toast>
```

bpas247:
> It's still a work in progress, but I could maybe use some suggestions on how to handle the little icon bug in the top left (and in general). Should it be a prop like icon="warning"? Or should we just not bother to support it and leave it up to lib users to implement it how they see fit?

[...]

PR #1384, opened 2019-01-29
[WIP] Add support for Bootstrap 4.2's Toasts feature (unfinished PR. feedback welcome)

charlex:
WIP for issue #1346

Screenshot of progress: https://user-images.githubusercontent.com/1542740/51679897-2eb69900-1fae-11e9-85eb-01a7ba9242de.png

--

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/docs/lib/examples/ToastDismiss.js
+++ b/docs/lib/examples/ToastDismiss.js
@@ -1,0 +1,39 @@
+/* eslint react/no-multi-comp: 0, react/prop-types: 0 */
+
+import React from 'react';
+import { Button, Toast, ToastBody, ToastHeader } from 'reactstrap';
+
+class ToastDismissExample extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      show: false
+    };
+
+    this.toggle = this.toggle.bind(this);
+  }
+
+  toggle() {
+    this.setState({
+      show: !this.state.show
+    });
+  }
+
+  render() {
+    return (
+      <div>
+        <Button color="danger" onClick={this.toggle}>{this.props.buttonLabel}</Button>
+        <br />
+        <br />
+        <Toast isOpen={this.state.show}>
+          <ToastHeader toggle={this.toggle}>Toast title</ToastHeader>
+          <ToastBody>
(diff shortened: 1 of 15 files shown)


src/ToastHeader.js
```diff
@@ -0,0 +1,82 @@
...
+      <button type="button" onClick={toggle} className={mapToCssModules('close', cssModule)} aria-label={closeAriaLabel}>
+        <span aria-hidden="true">{closeIcon}</span>
+      </button>
+    );
+  }
+
+  return (
+    <WrapTag {...attributes} className={classes}>
+      {icon && (
+        <svg
```
TheSharpieOne: I think this svg was just for the examples in the bootstrap docs. It would be good to allow the user to specific something custom.
Though, I do like the idea of specifying a color to have a square of that color. Maybe allow a component or a string. In the case of a component, the component is used as is. When it is a string, this svg can be used.

src/index.js
```diff
@@ -1,3 +1,5 @@
+import ToastHeader from "./ToastHeader";
```
bpas247: You're importing the `ToastHeader`? Why?

docs/lib/examples/ToastDismiss.js
```diff
@@ -0,0 +1,39 @@
...
+    this.state = {
+      show: false
+    };
+
+    this.toggle = this.toggle.bind(this);
+  }
+
+  toggle() {
+    this.setState({
+      show: !this.state.show
```
rkumar-auro: better to use prevState here.

GoPro16:
Was just looking to add this feature in and found this open PR. I was wondering if we could add in an additional prop such as `position` with possible values of `top, top-right, bottom, bottom-right` etc. 

Either this can be added or we can do something like the below:

Create a `Toaster` container that creates toasts:
```javascript
<Toaster position="top-right">
..Normal Application
</Toaster>
```

Then export a `context` and `useToaster` like the below:
```javascript
const toast = useToaster();
``` 

Then in any underlying component you can call the toaster to create by just using the `toast` method from the provider.

Wanted to get everyone's thoughts on this and possibly add this in as a feature after this has been merged in.

samuelcolvin:
Any progress on this?

Would be wonderful to get this deployed.

charlex:
@samuelcolvin I haven't had any time to work on this, but anyone is welcome to pick up where I left off. I think I did much of the gruntwork of putting together the right files, but there needs to be much more finessing.

seansfkelley:
I just opened https://github.com/reactstrap/reactstrap/pull/1447 to address the feedback on this PR and hopefully get it over the line.

PR #1410, opened 2019-02-22
feat(pagination-links): added props for first and last. changed carets…

GoPro16:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

Currently the `next` and `previous` props on the `PaginationLink` component have default carets that symbolize `first` and `last`. I added new props for `first` and `last` in order to have both features for navigating to any of these.

One change that is not breaking but does impact the view on the UI is the now the `next` and `previous` props are displaying single carets instead of double caret. I am not sure if this should be treated as a BREAKING CHANGE or not as its very minor in this case.

### DIFF
--- a/docs/lib/Components/PaginationPage.js
+++ b/docs/lib/Components/PaginationPage.js
@@ -58,6 +58,8 @@ PaginationLink.propTypes = {
   cssModule: PropTypes.object,
   next: PropTypes.bool,
   previous: PropTypes.bool,
+  first: PropTypes.bool,
+  last: PropTypes.bool,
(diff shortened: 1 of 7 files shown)


TheSharpieOne:
Thanks!

Yes, it would be considered a breaking change as the output is now different. Any build which pulls in minor releases and has snapshot tests would potentially break if they were using `previous` and `next` before.

I think the next release was already queued up for a major anyways, just waiting on a few extra breaking changes to be completed and merged to group them together.

GoPro16:
Awesome sounds good!

PR #1465, opened 2019-04-21
fix(Tooltip): Support for multiple target elements

### LINKED ISSUE #1185: Tooltip only show for first element
reactstrap version 6.4
react verison 16.4.2
bootstrap 4

In cases where there are elements with the same ID created through a loop or rendered from reusable components, the tooltip only show for the first element with the ID.

I try to fix it myself but I lack the knowledge of reading other people's code. The issue, I suspect lies here

```
function getTarget(target) {
  var els = findDOMElements(target);
  if (isArrayOrNodeList(els)) {
    return els[0];
  }
  return els;
}
```

It's always returning the first element of the list, even when there are more than one elements. In the original bootstrap 4, it initializes tooltip for any elements with the attribute `data-toggle` (from the example), thus allowing same tooltip to show on different same elements. Consider using a loop to loop through `NodeList` and apply the `addEventListener`.

akhlesh:
It fixes #1185 
Here is stackblitz demo with fix - https://stackblitz.com/edit/react-k8rkzv

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/TooltipPopoverWrapper.js
+++ b/src/TooltipPopoverWrapper.js
@@ -61,11 +61,16 @@ function isInDOMSubtree(element, subtreeRoot) {
   return subtreeRoot && (element === subtreeRoot || subtreeRoot.contains(element));
 }
 
+function isInDOMSubtrees(element, subtreeRoots = []) {
+  return subtreeRoots && subtreeRoots.length && subtreeRoots.find(subTreeRoot=> isInDOMSubtree(element, subTreeRoot));
+}
+
 class TooltipPopoverWrapper extends React.Component {
   constructor(props) {
     super(props);
 
-    this._target = null;
+    this._targets = [];
+    this.currentTargetElement = null;
     this.addTargetEvents = this.addTargetEvents.bind(this);
     this.handleDocumentClick = this.handleDocumentClick.bind(this);
     this.removeTargetEvents = this.removeTargetEvents.bind(this);
@@ -90,6 +95,7 @@ class TooltipPopoverWrapper extends React.Component {
 
   componentWillUnmount() {
     this.removeTargetEvents();
+    this._targets = null;
   }
 
   static getDerivedStateFromProps(props, state) {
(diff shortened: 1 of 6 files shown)


Kliton:
Up

src/TooltipPopoverWrapper.js
```diff
@@ -231,54 +250,55 @@ class TooltipPopoverWrapper extends React.Component {
...
+      this.removeEventOnTargets('focusout', this.hide, true);
     }
 
     document.removeEventListener('click', this.handleDocumentClick, true)
   }
 
   updateTarget() {
-    const newTarget = getTarget(this.props.target);
-    if (newTarget !== this._target) {
+    const newTarget = getTarget(this.props.target, true);
```
felixfbecker: This PR broke support for refs as `target` prop. When `props.target` is a React ref or an HTMLElement, this will return an HTMLElement, not an array. `Array.from(newTarget)` will then return an empty array, and Popper will error that `target` was set to `undefined`.

src/TooltipPopoverWrapper.js
```diff
@@ -231,54 +250,55 @@ class TooltipPopoverWrapper extends React.Component {
...
+    if (newTarget !== this._targets) {
```
felixfbecker: Using reference equality here to compare two arrays doesn't look right

src/TooltipPopoverWrapper.js
TheSharpieOne: fixed in #1687

src/TooltipPopoverWrapper.js
TheSharpieOne: Yes, it seems like this would always be `true` since we would be getting a new array from `getTarget` every time. Probably need to perform a deep comparison here.

felixfbecker: And then only remove the event listeners from the targets that actually got removed (vs removing from all the targets if one was removed)

TheSharpieOne: it might be easier _and quicker_ to just remove them all and add the newTarget back 🤷‍♂

felixfbecker: True

PR #1513, opened 2019-05-29
feat(Dropdown) Add `container` prop to Dropdown, like what's already in PopperContent

hsource:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

When a Dropdown is inside a small element with overflow: hidden, it's useful to have the dropdown display directly inside of body or another container. I added the same `menuContainer` prop to `Dropdown`. Using a React Portal, this allows the dropdown to have `body` or any other tag as the parent.

Note that the prop needs to be on `Dropdown` instead of `DropdownMenu` since `Dropdown` needs to know about what the container is for some of its click handlers.

## Old behaviour
Previous

## New behaviour
**Code:**
```
<Dropdown menuContainer="body">
  <DropdownMenu modifiers={{ preventOverflow: { boundariesElement: 'viewport' }}}>...</DropdownMenu>
</Dropdown>
```

New: with `<DropdownMenu modifiers={{ preventOverflow: { boundariesElement: 'viewport' }}}>`

### DIFF
--- a/docs/lib/Components/DropdownsPage.js
+++ b/docs/lib/Components/DropdownsPage.js
@@ -69,6 +69,9 @@ export default class DropdownPage extends React.Component {
   tag: PropTypes.string, // default: 'div' unless nav=true, then 'li'
   toggle: PropTypes.func,
   setActiveFromChild: PropTypes.bool
+  // Where to inject the popper menu DOM node. Can be used to help menu
+  // escape from a div with overflow, for example by setting it to "body"
+  menuContainer: PropTypes.oneOfType([PropTypes.string, PropTypes.func, DOMElement]),
 };
 
 DropdownToggle.propTypes = {
--- a/docs/lib/Components/DropdownsPage.js
+++ b/docs/lib/Components/DropdownsPage.js
@@ -97,7 +97,8 @@ DropdownMenu.propTypes = {
   cssModule: PropTypes.object,
   // Custom modifiers that are passed to DropdownMenu.js, see https://popper.js.org/popper-documentation.html#modifiers
   modifiers: PropTypes.object,
-  persist: PropTypes.bool // presist the popper, even when closed. See #779 for reasoning
+  persist: PropTypes.bool, // presist the popper, even when closed. See #779 for reasoning
+  positionFixed: PropTypes.bool,
 };
 
 DropdownItem.propTypes = {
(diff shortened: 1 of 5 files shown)


mariocasciaro:
I see a warning when using the new prop

DevTools_-_localhost_3000_Xl47BKAs7CrqyQG_testSuites_p1Ayl17xCkCLE9lL_testCases

Also would it make more sense to use a `container` option on the `DropdownMenu` component instead of using the `menuContainer` option on the `Dropdown` component?

hsource:
Note that I removed the `container` prop. The only prop that should be used is `menuContainer` on `Dropdown`.

This is because `Dropdown` needs to know what the container is to detect clicks outside of the dropdown to close the dropdown. As such, `menuContainer` is passed to `Dropdown`, which uses the context to pass it to any child `DropdownMenu`.

I also added a new positionFixed parameter to DropdownMenu

hsource:
The new commit should also fix the console errors shown above

hsource:
I updated the API again. Now, the API is:

1. DropdownMenu has new prop `container`. `container` can be used to move the Dropdown to have another parent like "body" or "#myCustomDiv"
2. DropdownMenu also has new prop `positionFixed`. This allows to keep it fixed on page instead of absolute
2. Dropdown has new prop `menuClickTarget`. Dropdown needs to know where the menu is so that when it detects any clicks outside of the menu, it closes the menu. `menuClickTarget` allows the developer to set a custom click target

All 3 have been tested and documented

mariocasciaro:
Hi @hsource thanks for the PR and the updates. I tried your new changes and it seems that the `onClick` event on a `MenuItem` is not invoked when the container is `body`. It was working before your last update.

hsource:
If you're setting container="body", then you must set some kind of ID or class on `<DropdownMenu>` and then set that in `<Dropdown menuClickTarget>`. For example, write:

```
<Dropdown menuClickTarget="myMenuId">
  <DropdownMenu id="myMenuId">...</DropdownMenu>
</Dropdown>
```

hsource:
I'm not a huge fan of this API right now. Ideally, the `Dropdown` would know where the `DropdownMenu` is rendered, but there's no way to pass context up the chain. I think integrating `DropdownMenu` into `Dropdown` or moving the click handler that toggles the Dropdown into `DropdownMenu` may be a good solution, but I don't have the time to do that right now.

mariocasciaro:
I agree, this is error prone. An alternative would be to pass the reference of `Dropdown` down to `DropdownMenu` using a context. However all those solutions would probably make the code more complicated than what it needs to be. Considering this I would be in favour for using the old `menuContainer` prop on `Dropdown`.

hsource:
I made a new change (#2016) fixing the architecture issues! Take a look

PR #1545, opened 2019-06-22
chore(release): adding 8.0.3

apperside:
- [x ] Bug fix Fixed Carousel to scroll in wrong direction when scrolling to an arbitrary position
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

When using carousel, calling next for example from 0 to 3, it was scrolling left. I believe that regardless of the fact the indicator was clicked, if the target index is bigger than the next one, it should scroll right


TheSharpieOne:
What happened to 8.0.1 and 8.0.2?

PR #1571, opened 2019-07-11
fix: Dropdown disabled toggle issue

### LINKED ISSUE #1542: Dropdown disabled style inconsistent
- components: `Dropdown`
- reactstrap version `8.0.0`
- import method `umd/csj/es` 
- react version `16.8.6`
- bootstrap version `4.3.1`

### What is happening?

On a `Dropdown` component, there is a prop to disable it. But when the disabled prop is applied, there is no indication that the dropdown is disabled.

The `Dropdown` can still be opened even when it is disabled and this does not match the rest of the disabled behaviors on components like `Button`, `Input`, etc..

From the docs

```js
Dropdown.propTypes = {
  disabled: PropTypes.bool,
  ...other props
};
```

### What should be happening?

The `Dropdown` should have the disabled state applied and be inaccessible.

### Steps to reproduce issue

Just apply the `disabled` prop to any `Dropdown` component

### Error message in console

N/A

### Code

Check out the [codesandbox here](https://codesandbox.io/s/reactstrap-lli8e)

bpas247:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [xI have read the **[CONTRIBUTING](./CONTRIBUTING.md)** document.

This fix addresses the issue where Dropdown will be passed a
`disabled` prop, yet it is still clickable. This is due to context
being passed the user-defined `toggle` prop directly, rather than
our defined `this.toggle` function that includes a check for
`disabled`.

Fixes #1542

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -52,10 +52,11 @@ class Dropdown extends React.Component {
 
   getContextValue() {
     return {
-      toggle: this.props.toggle,
+      toggle: this.toggle,
       isOpen: this.props.isOpen,
       direction: (this.props.direction === 'down' && this.props.dropup) ? 'up' : this.props.direction,
       inNavbar: this.props.inNavbar,
+      disabled: this.props.disabled
     };
   }
 
--- a/src/DropdownContext.js
+++ b/src/DropdownContext.js
@@ -7,6 +7,7 @@ import React from 'react';
  *  isOpen: PropTypes.bool.isRequired,
  *  direction: PropTypes.oneOf(['up', 'down', 'left', 'right']).isRequired,
  *  inNavbar: PropTypes.bool.isRequired,
+ *  disabled: PropTypes.bool
  * }
  */
 export const DropdownContext = React.createContext({});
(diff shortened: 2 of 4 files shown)


bpas247:
hmm, I'm not exactly sure why those tests are failing, they seem unrelated to the changes I committed.

Anyone have any ideas what could be causing this?

TheSharpieOne:
I think you found some bad tests. It looks like it was expecting the Dropdown toggle to not be called... when it should have been getting called. Since you fixed it to call Dropdown's toggle (which in turn calls the user's toggle), those tests fail.

In reality, those tests probably should have been failing this whole time and this PR would have make them pass.

Can you change `expect(Dropdown.prototype.toggle.mock.calls.length).toBe(0);` to `expect(Dropdown.prototype.toggle.mock.calls.length).toBe(1);` in both of those tests to get them to pass as they should be expecting Dropdown's toggle to be called 1 time.

bpas247:
> Can you change `expect(Dropdown.prototype.toggle.mock.calls.length).toBe(0);` to `expect(Dropdown.prototype.toggle.mock.calls.length).toBe(1);` in both of those tests to get them to pass as they should be expecting Dropdown's toggle to be called 1 time.

I had a feeling they were wrong :wink: will do 👍

PR #1572, opened 2019-07-11
feat: migrate doc examples to function components

bpas247:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

This PR aims to provide consistency for all current doc examples for reactstrap, by migrating all existing example components to function components via hooks.

~~**Currently in draft**, as this will require a large amount of work to migrate all existing doc examples.~~

**Now ready for review** 🚀

### DIFF
--- a/docs/lib/examples/Badge.js
+++ b/docs/lib/examples/Badge.js
@@ -1,17 +1,17 @@
 import React from 'react';
 import { Badge } from 'reactstrap';
 
-export default class Example extends React.Component {
-  render() {
-    return (
-      <div>
-        <h1>Heading <Badge color="secondary">New</Badge></h1>
-        <h2>Heading <Badge color="secondary">New</Badge></h2>
-        <h3>Heading <Badge color="secondary">New</Badge></h3>
-        <h4>Heading <Badge color="secondary">New</Badge></h4>
-        <h5>Heading <Badge color="secondary">New</Badge></h5>
-        <h6>Heading <Badge color="secondary">New</Badge></h6>
-      </div>
-    );
-  }
+const Example = (props) => {
+  return (
+    <div>
+      <h1>Heading <Badge color="secondary">New</Badge></h1>
+      <h2>Heading <Badge color="secondary">New</Badge></h2>
+      <h3>Heading <Badge color="secondary">New</Badge></h3>
+      <h4>Heading <Badge color="secondary">New</Badge></h4>
+      <h5>Heading <Badge color="secondary">New</Badge></h5>
+      <h6>Heading <Badge color="secondary">New</Badge></h6>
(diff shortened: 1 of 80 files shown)


bpas247:
Taking this out of draft phase, since most of the doc examples have been re-written by now :smiley:

Let me know if I missed any examples :v:

docs/lib/examples/Badge.js
```diff
@@ -1,17 +1,17 @@
...
-        <h2>Heading <Badge color="secondary">New</Badge></h2>
-        <h3>Heading <Badge color="secondary">New</Badge></h3>
-        <h4>Heading <Badge color="secondary">New</Badge></h4>
-        <h5>Heading <Badge color="secondary">New</Badge></h5>
-        <h6>Heading <Badge color="secondary">New</Badge></h6>
-      </div>
-    );
-  }
+const Example = (props) => {
+  return (
```
GoPro16: Since you are doing a functional components you can always just directly return the `jsx` here rather than having a return method. Just a recommendation for all of these refactors but otherwise LGTM.
```jsx
const Example = props => (<div>stuff</div>)
```

bpas247: the reason why I’m not implicitly returning is because I wanted to make the docs as explicit as possible, to avoid confusion for new JS developers (hence why I’m also including “props” as an argument for components that don’t use it). 

I should’ve went more explicit with the syntax and used “function” declarations instead of arrow function declarations but I wanted the doc examples to be consistent and I really didn’t feel like refactoring the other function component examples 😛the diff for this PR is really big already.

GoPro16: Sounds good. Agreed she’s a big one 🎣

bpas247:
What's currently preventing this from merge? I'm hoping to get this merged ASAP, since it's been stale for quite awhile now

bpas247:
Looks like a few new doc examples were added since I last worked on this PR that are class components 👀 I think I'll document them in a new issue and work on them in a later PR, the diffs for this one are too large at this point to introduce any new changes.

kiley0:
Really appreciate this PR -- when we're onboarding a new developer, it was confusing for them to have to switch between function components in our codebase and class components in the reactstrap docs. Wanted to say thank you! 🙌

bpas247:
> Really appreciate this PR -- when we're onboarding a new developer, it was confusing for them to have to switch between function components in our codebase and class components in the reactstrap docs. Wanted to say thank you! 🙌

Yeah that's how I felt about the docs personally when I was using this library for a project, very annoying to have to translate the class component examples to function components. No problem 😄 and please let me know if you have any suggestions for any future improvements in the docs!

Issue #1577, opened 2019-07-15
Testing migration for less testing on implementation details


bpas247:
# Issue
The current state of our testing codebase has numerous examples of testing implementation details, which [can break our code during refactoring](https://kentcdodds.com/blog/testing-implementation-details).

# Example
A good example of what I mean can be found in `UncontrolledCollapse`. The current implementation goes as follows:
https://github.com/reactstrap/reactstrap/blob/b4edeb8c417aaa7e8d6370e3fd242612c2eae73c/src/UncontrolledCollapse.js#L18-L54

If I wanted to refactor it to be a function component, like so:
```jsx
const UncontrolledCollapse = ({
  toggler,
  defaultOpen, 
  toggleEvents,
  ...props
}) => {
  const [isOpen, setIsOpen] = useState(defaultOpen || false);

  const toggle = e => {
    e.preventDefault();
    setIsOpen(!isOpen);
  }

  useEffect(() => {
    const togglers = findDOMElements(toggler);
    const removeEventListeners = addMultipleEventListeners(
      togglers,
      toggle,
      toggleEvents
    );
    return () => {
      if (togglers.length && removeEventListeners) {
        removeEventListeners();
      };
    }
  }, []);

  return <Collapse isOpen={isOpen} {...omit(props, omitKeys)} />;
}
```

The test suite would eventually fail due to this test (among other tests):
https://github.com/reactstrap/reactstrap/blob/b4edeb8c417aaa7e8d6370e3fd242612c2eae73c/src/__tests__/UncontrolledCollapse.spec.js#L41-L48
Which is testing an implementation detail :)

# Solution

We should re-design our testing to test less implementation details, using [react-testing-library](https://github.com/testing-library/react-testing-library) rather than enzyme.

For example:
```js
// at the top of `UncontrolledCollapse.spec.js`
import {
  render,
  fireEvent,
  cleanup,
} from '@testing-library/react';

import { toHaveClass } from '@testing-library/jest-dom';

expect.extend({toHaveClass});
```

```jsx
  // ... inside describe
  afterEach(cleanup);
  it('should trigger Collapse to expand when it is clicked', () => {
    const { getByText } = render(<UncontrolledCollapse defaultOpen={false} toggler="#toggler">Yo!</UncontrolledCollapse>);

    fireEvent.click(getByText("Click Me"));

    expect(getByText("Yo!")).toHaveClass("collapsing");
  });
```

Would pass, whether we used the class or function component implementation 😎

[...]

PR #1578, opened 2019-07-16
[WIP] feat: Migrate Testing to React-Testing-Library

bpas247:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

This is the first pass-through of migrating our unit tests to `react-testing-library`, in an attempt to remove testing implementation details.

Relevant to #1577

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
   "homepage": "https://github.com/reactstrap/reactstrap#readme",
   "dependencies": {
     "@babel/runtime": "^7.2.0",
+    "@testing-library/react": "^8.0.5",
     "classnames": "^2.2.3",
     "lodash.isfunction": "^3.0.9",
     "lodash.isobject": "^3.0.2",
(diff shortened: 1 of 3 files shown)


bpas247:
Hmm, the more I think about this the less I feel like this is an improvement for anything. The main motivation for this migration was to prevent testing implementation details, but I'm not sure if a migration to RTL would necessarily be a good solution for this. [This convo from React Bootstrap](https://github.com/react-bootstrap/react-bootstrap/issues/4693#issuecomment-542662417) seems to highlight why this might not be a good approach to solving it, as we'd be re-writing essentially every test to achieve almost the same effect. 

It might just be easier to re-write the tests that are testing implementation details than migrating our entire testing suite.

Anyone have further thoughts on this?

TheSharpieOne:
Personally, "if it ain't broke, don't fix it". Unless the current test are impeding progress or preventing things from being fully tested, I don't see a need to make this change globally. Sure, there are cases where RTL would have benefits, and we might want to bring it in for those cases and use it side-by-side where it is needed, but I don't think we need to redo testing as a whole.

bpas247:
> Personally, "if it ain't broke, don't fix it". Unless the current test are impeding progress or preventing things from being fully tested, I don't see a need to make this change globally. Sure, there are cases where RTL would have benefits, and we might want to bring it in for those cases and use it side-by-side where it is needed, but I don't think we need to redo testing as a whole.

I completely agree, just seems like this PR is creating more work with not much payoff. I'll go ahead and close this PR then :)

PR #1633, opened 2019-09-26
feat(NavbarToggler): add aria-label to NavbarToggler

### LINKED ISSUE #1632: NavbarToggler Accessibility
reactstrap should add the `aria-label="Toggle navigation"` as seen in bootstrap's docs. It would be overridable by the props passed in.
reactstrap should also include information about the `aria-controls` and `aria-expanded` for additional a11y as those would need to be provided by the user (`NavbarToggler` doesn't actually know if it is toggled or not (maybe it could/should?))

glenpadua:
This adds the aria-label to the NavbarToggler as
seen in the bootstrap docs.

Fixes #1632

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/NavbarToggler.js
+++ b/src/NavbarToggler.js
@@ -31,7 +31,7 @@ const NavbarToggler = (props) => {
   ), cssModule);
 
   return (
-    <Tag {...attributes} className={classes}>
+    <Tag aria-label="Toggle navigation" {...attributes} className={classes}>
       {children || <span className={mapToCssModules('navbar-toggler-icon', cssModule)} />}
     </Tag>
   );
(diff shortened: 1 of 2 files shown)


glenpadua:
Working on the breaking tests.

src/NavbarToggler.js
```diff
@@ -31,7 +31,7 @@ const NavbarToggler = (props) => {
   ), cssModule);
 
   return (
-    <Tag {...attributes} className={classes}>
+    <Tag {...attributes} className={classes} aria-label="Toggle navigation">
```
TheSharpieOne: I would place it before `{...attributes}` so that the attributes can override it.

glenpadua: Makes sense. Done 👍

macsplan:
hi @TheSharpieOne any ideas when's this getting to release?

Issue #1602, opened 2019-08-18
Carousel component needs update to conform to new react lifecyle methods


gone-skiing:
- components: `Carousel `
- reactstrap version `8.0.1`
- react version `16.9.0`
- bootstrap version `4.3.1`

### What is happening?

backend.js:1 Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

### What should be happening?

Deprecated method should not be used.

### Steps to reproduce issue

1. Any use of the component.

### Error message in console

```
backend.js:1 Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
* Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: Carousel
```

drewskiii:
I'm having the same issue.

Tulean:
Did you guys find any fix for it? I am also having the same issue.

gone-skiing:
I never heard from maintainers. One option is to simply add prefixes to methods to suppress the warnings. Full credit fix requires changing component lifecycle methods to supported, I believe.

ahmeed83:
Im still having the same issue. Are they planning to include it to the new version of reacstrap?

stevesioli:
is anybody already working on this?  if not, I'd like to resolve for hacktoberfest.

MarkEhr:
@stevesioli looks like #1638 fixes it for this specific component, but I'm sure you can find other comonents with the same problem if you wanna get Hacktoberfest points.

stevesioli:
@MarkEhr I left a comment on that PR.  The object comparison they are using will not work as expected.

onlywicked:
Anybody working on it. I would love to work on this issue.

TheSharpieOne:
I believe @stevesioli might be working on this.

onlywicked:
No worry.

stevesioli:
[...]

PR #1638, opened 2019-10-03
fix(Carousel): Fix deprecated lifecycle method componentWillReceiveProps

CDimonaco:
Fix https://github.com/reactstrap/reactstrap/issues/1602

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

My approach for solving this issue was to delete the componentWillReceiveProps in favor of a componentDidUpdate implementation with the same logic, another approach was to switch to the new static method getDerivedStateFromProps, I tried the switch but the final result was really messy and full of workaround in order to not refactoring all the component and in particular the timeout handling logic. Using the componentDidUpdate results in a more clean and smooth logic implementation.

Refers to the issue https://github.com/reactstrap/reactstrap/issues/1602

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -31,19 +31,24 @@ class Carousel extends React.Component {
     document.addEventListener('keyup', this.handleKeyPress);
   }
 
-  componentWillReceiveProps(nextProps) {
-    this.setInterval(nextProps);
-    // Calculate the direction to turn
-    if (this.props.activeIndex + 1 === nextProps.activeIndex) {
-      this.setState({ direction: 'right' });
-    } else if (this.props.activeIndex - 1 === nextProps.activeIndex) {
-      this.setState({ direction: 'left' });
-    } else if (this.props.activeIndex > nextProps.activeIndex) {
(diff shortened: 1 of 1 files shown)


src/Carousel.js
```diff
@@ -31,19 +31,24 @@ class Carousel extends React.Component {
...
-      this.setState({ direction: this.state.indicatorClicked ? 'left' : 'right' });
-    } else if (this.props.activeIndex !== nextProps.activeIndex) {
-      this.setState({ direction: this.state.indicatorClicked ? 'right' : 'left' });
-    }
-    this.setState({ indicatorClicked: false });
+  componentDidUpdate(prevProps){
+    // we try to simulate the componentWillReceive props
+    // purpose, if the props are the same we are not interested to
+    // proceed anymore in the function.
+    if(prevProps === this.props) return;
```
stevesioli: this type of comparison does not work with objects.  this will always result in `false`

You need to create some kind of function to iterate through each object property and check for equality.

iamandrewluca:
@TheSharpieOne this PR is outdated, can be closed.

PR #1650, opened 2019-10-07
Fix key in UncontrolledCarousel

### LINKED ISSUE #1536: UncontrolledCarousel should not use src of an item as key
If you want to use the same image in 2 different slides, you have a react warning :

    Warning: Encountered two children with the same key, `https://xxx.png`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

The problem lies in https://github.com/reactstrap/reactstrap/blob/master/src/UncontrolledCarousel.js#L67

Here, the key is defined to be img.src

I suggest to use the index in the array as the key (which could be the second param of the map function on line 62)

My current workaround is to copy / paste the entire UncontrolledCarousel call to change that.

baktiaditya:
fix #1536

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/docs/lib/examples/CarouselUncontrolled.js
+++ b/docs/lib/examples/CarouselUncontrolled.js
@@ -3,18 +3,24 @@ import { UncontrolledCarousel } from 'reactstrap';
 
 const items = [
   {
+    // by default, `src` would be the key for an item.
+    // if there are similar `src` value in item, you should define `key` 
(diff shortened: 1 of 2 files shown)


GoPro16:
superseded by #1536

Issue #1664, opened 2019-10-15
Tooltip crash the current open tab in the browser when the space is not availaible for the tooltip;


Aashishgtbit:
- components: `Tooltip`
- reactstrap version `#8.0.1`
- import method `umd/csj/es` 
- react version `#16.9.0`
- bootstrap version `#4.x`

### What is happening?

When I use the tool tip in my web app if the browser has not enough space for the tooltip then it will crash the current open tab . For example I have a side bar which has some data rows and I want to use tooltip for each item and position it right  , It is working perfectly fine for the top position .
### What should be happening?

it should adjust its position automatically when enough space is not available .
### Steps to reproduce issue

1. ...
2. ...

### Error message in console

```
paste error message with stacktrack here
```

nylon22:
Looks like duplicate of #1488 and #1482

Issue #1488, opened 2019-05-10
Popover causes the browser to crash if window size is changed


liemdo:
- components: `popover`
- reactstrap version `8.0.0`

### Steps to reproduce the issue

1. Open the popover in Chrome or Safari (I didn't test with other browsers)
2. Change the window size
3. The browser crashes. I have to force quit the browser.

TheSharpieOne:
I cannot reproduce. Does this happen for you on https://reactstrap.github.io/components/popovers/ ?

liemdo:
You have to resize the window quickly so that part of the popover is out of the window view. The browser crashes right after that.

![reactstrap](https://user-images.githubusercontent.com/989817/57551219-1fc1df80-732e-11e9-9d36-ee58bf59d528.gif)

If you open the Chrome Devtools while the popover is open, you'll get the same issue. 

This issue may be related to this one https://github.com/reactstrap/reactstrap/issues/1482

morel333:
> I cannot reproduce. Does this happen for you on https://reactstrap.github.io/components/popovers/ ?

Open your samples in minimal browser size and click on any popover - it will reproduce.
Looks like ther's an infinite state placement change in PopperContent in handlePlacementChange (placement changes top/bottom -> etc).

vitorblz:
The same happens with tooltips

morel333:
As a hotfix changed
`        <ReactPopper
          referenceElement={this.targetNode}
          modifiers={extendedModifiers}
          placement={placement}
        >`
to
`<ReactPopper
                    referenceElement={this.targetNode}
                    modifiers={extendedModifiers}
                    placement={attrs.placement}
                >`
in PopperContent

usman-subhani:
![Tooltiperror](https://user-images.githubusercontent.com/25919458/58682585-9892e500-838a-11e9-9268-90c5e9831be0.gif)

My browser crashes as well while using tooltips.Its a bit tricky to reproduce but as far as i know it often happens when mouse moves to quickly over the tooltip hover element or mouse hovers over a tooltip element during/after scrolling. I am using uncontrolled tooltips with delays mentioned below.
I am using v8.0.0 of reactstrap.
```
const tooltipDelay = { show: 500, hide: 300 };

const UncontrolledTooltipWithDelay = props => (
  <UncontrolledTooltip delay={tooltipDelay} trigger="hover" {...props} />
);
```

Phibedy:
[...]

Issue #1482, opened 2019-05-06
Popover/Tooltip breaks if parent is smaller(height) causing memory leak.


rabrenio:
- components: `Popover/Tooltip`
- reactstrap version `8.0.0`

Here is how you can replicate the issue.

Tooltip's works if you set placement.

This can also happens if the window size is smaller than the Popover/Tooltip.

[codesanbox](https://w7yj17kw87.codesandbox.io/)

jisaf:
I will add to this - we're running into this problem as well.

After updating to 8.0.0 one of our tooltips (specifically, one used on an AgGrid row) is causing the CPU to jump to 100% and lock up the browser.

Reverting to pre-8 solves the problem (for anyone reading this, note that whatever dev-server you're running will probably need to be re-started as well).

Increasing the height of the row, however, does not seem to solve the problem.

leocorder:
Same problem here after update from 6.5.0 to 8.0.0.

val1984:
Getting this issue too when using `reactstrap@8.0.0`.
Seems that the browser is locked in an async deadlock.

YassienW:
Same thing here, does anyone have any pointers as to what specifically is causing this? Or to any possible workarounds?

leocorder:
@YassienW I ended up rolling back to version 6.5.0.

choweiyuan:
> Same thing here, does anyone have any pointers as to what specifically is causing this? Or to any possible workarounds?

@YassienW  https://github.com/reactstrap/reactstrap/issues/1488#issuecomment-492527252 try setting placement as suggested from the linked comment

YassienW:
It does fix it since it stops the infinite render loop. I am not sure whether this is an issue with reactstrap or react-popper though.

val1984:
Considering [#1488 (comment)](https://github.com/reactstrap/reactstrap/issues/1488#issuecomment-492527252) fixes this issue, it seems that what happens is that `this.state.placement` keeps flip-flopping when there's not enough space, resulting in a deadlock since ReactPopper return a different `placement` each time than the original one which cannot be honored.

val1984:
A fix which doesn't require modifying `reactstrap`, until it is fixed, is to add the following `modifiers` prop to your `Popover` or `Tooltip`:
```typescript
<Popover
placement="bottom"
modifiers={{ flip: { behavior: ['bottom'] } }}
// additional props
>
/// children
</Popover>
```

YassienW:
[...]

PR #1669, opened 2019-10-17
PopperContent memory leak

phwebi:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

When a tooltip or popover does not fit in the containing element, this causes the `PopperContent` component to get into an infinite loop of switching between `placement`s in its state. This is because we attempt to update `this.state.placement` everytime popper.js flips the placement to find a better fit and pass it back as the `placement` property to `ReactPopper`.

This is a problem because the `PopperContent` component has two sources of truth for its placement. One coming from its own `placement` property, and one coming from the internal data of `ReactPopper`.

This commit fixes the issue by using `PopperContent`'s `placement` property as the initial placement of the `ReactPopper` component and allowing `ReactPopper` to manage its own updates (i.e. if the popover needs to be flipped because there's not enough room for the specified placement.)

I manually tested this by reproducing the issues linked below at https://reactstrap.github.io/components/popovers/, and making sure that the page wouldn't crash when I followed the same steps on the changes in this branch.

Fixes https://github.com/reactstrap/reactstrap/issues/1482
Fixes https://github.com/reactstrap/reactstrap/issues/1488
Fixes https://github.com/reactstrap/reactstrap/issues/1664

### DIFF
--- a/src/PopperContent.js
+++ b/src/PopperContent.js
@@ -51,7 +51,6 @@ class PopperContent extends React.Component {
   constructor(props) {
     super(props);
 
-    this.handlePlacementChange = this.handlePlacementChange.bind(this);
     this.setTargetNode = this.setTargetNode.bind(this);
     this.getTargetNode = this.getTargetNode.bind(this);
     this.getRef = this.getRef.bind(this);
@@ -88,13 +87,6 @@ class PopperContent extends React.Component {
     this._element = ref;
   }
 
-  handlePlacementChange(data) {
-    if (this.state.placement !== data.placement) {
-      this.setState({ placement: data.placement });
-    }
-    return data;
-  }
-
   onClosed() {
     this.props.onClosed();
     this.setState({ isOpen: false });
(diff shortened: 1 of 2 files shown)


src/PopperContent.js
```diff
@@ -120,28 +112,22 @@ class PopperContent extends React.Component {
...
     const arrowClassName = mapToCssModules(classNames(
       'arrow',
       _arrowClassName
     ), cssModule);
-    const placement = this.state.placement || attrs.placement;
-    const placementFirstPart = placement.split('-')[0];
     const popperClassName = mapToCssModules(classNames(
       _popperClassName,
-      placementPrefix ? `${placementPrefix}-${placementFirstPart}` : placementFirstPart
+      placementPrefix ? `${placementPrefix}-auto` : ''
```
phwebi: This allows the arrow to inherit its style from the computed popper placement instead of the `placement` property passed into this component.

See https://github.com/twbs/bootstrap/blob/c1ee395f80c05de8317588b07f34a65c5b95c42c/scss/_popover.scss#L136

PR #1678, opened 2019-10-20
Removes condition to call close when modal is unmounting

### LINKED ISSUE #1626: Class 'modal-open' is persisted on body even when the Modal component is unmounted
### Issue description

- components: `Modal`
- reactstrap version `#8.0.1`
- react version `#16.9.0`
- bootstrap version `#4.3`

### What is happening?

Class 'modal-open' is persisted on body even when the Modal component is unmounted.
I think this happens because the modal is not displayed entirely and unmounted.

### What should be happening?

Class 'modal-open' should be removed of the body when the Modal component is unmounted.

### Steps to reproduce issue

Instanciate a modal with `isOpen=true` by default, unmount the component before it was even displayed. (see #code)

### Error message in console

No error message.

### Code

https://codesandbox.io/embed/buggy-modal-vnyml

artdevgame:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

Fixes #1626: Class 'modal-open' is persisted on body even when the Modal component is unmounted.

I'm not sure why close is conditionally called after destroying a modal but it feels like the cleanup code should run unconditionally.

I'm also not sure if my test is written in the best way. It demonstrates the problem but it looks a little clunky. Any advice is appreciated.

I might need a bit of help getting this PR into a state that into the appropriate state for merge as I got confused whilst reading the contribution guidelines.

#1626

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -147,9 +147,7 @@ class Modal extends React.Component {
 
     if (this._element) {
       this.destroy();
-      if (this.state.isOpen) {
-        this.close();
-      }
+      this.close();
     }
 
     this._isMounted = false;
--- a/src/__tests__/Modal.spec.js
+++ b/src/__tests__/Modal.spec.js
@@ -746,6 +746,40 @@ describe('Modal', () => {
     expect(document.body.className).toBe('');
   });
 
+  it.only('should not keep modal-open class on the body if the modal is not rendered', () => {
+    class ModalWrapper extends React.Component {
+      constructor (props) {
+        super(props);
+
+        this.state = {
+          displayModal: true
+        };
+      }
+
+      componentDidMount() {
+        this.setState({ displayModal: false });
+      }
+
+      render () {
+        const { displayModal } = this.state;
+
+        return (
+          <div>
+            {displayModal && (
+              <Modal isOpen={true} toggle={toggle}>
+                Hello!
+              </Modal>
+            )}
+          </div>
+        )
+      }
+    }
+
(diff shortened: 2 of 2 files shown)


TheSharpieOne:
> I'm not sure why close is conditionally called after destroying a modal but it feels like the cleanup code should run unconditionally.
I believe it is to better handle the case when the modal is already closed and has already removed _its_ instance of the `modal-open` from the global counter.
When the change suggested here, a modal which is closed, then removed later, it would decrease the value of the counter by 2 instead of the desired 1. When multiple modals are open, this would cause problems.

artdevgame:
@TheSharpieOne Ah cool, so we need something in between then. Something that conditionally calls the close function, but also removes the `modal-open` class when appropriate.

TheSharpieOne:
Yes. Ideally, mounting / unmounting itself should not be adding/removing the modal-open class nor incrementing/decrementing the counter. It should only really be happening with the modal is opened and closed. Of course, mounting / initializing a modal with `isOpen` set to `true` would count as "opening" and unmounting an open modal would count as "closing" so thus mounting and unmounting are in the mix.

I'm not 100% sure, but I believe because the component mounts and unmounts so quickly that the async status update doesn't have time to run and thus that if statement you removed evaluates to `false` even when the modal is "open". I forget why we needed to track the open status in state... maybe it's no longer needed and replacing `this.state.isOpen` with `this.props.isOpen` throughout `Modal` will fix this issue?

Issue #1686, opened 2019-10-24
the targets prop passed to child component is undefined


zachary822:
- components: `TooltipPopoverWrapper`
- reactstrap version `#8.1.1`
- import method `umd/csj/es` 
- react version `#16.10.2`
- bootstrap version `#4.3.1`

### What is happening?

In the [`updateTarget`](https://github.com/reactstrap/reactstrap/blob/45775c01da58e4e29d752367b43a730258398d78/src/TooltipPopoverWrapper.js#L296-L304) method. If `this.props.target` is a function that returns a single element (not array), then the [`Array.from(newTarget)`](https://github.com/reactstrap/reactstrap/blob/45775c01da58e4e29d752367b43a730258398d78/src/TooltipPopoverWrapper.js#L300) will return `[]` (empty array). This will cause `this._targets` to be set to an empty array. In turn [`this.currentTargetElement` ](https://github.com/reactstrap/reactstrap/blob/45775c01da58e4e29d752367b43a730258398d78/src/TooltipPopoverWrapper.js#L301) to be set to `undefined`. Which will cause `undefined` to be passed to `PopperContent` as its `target` prop. Resulting in no tooltips showing up. 

### What should be happening?

Tooltip should show up if `target` prop is valid.

### Steps to reproduce issue

1. passing a function to the `target` prop of `Tooltip`
2. ...

### Error message in console

```
backend.js:6 Warning: Failed prop type: The prop `target` is marked as required in `PopperContent`, but its value is `undefined`.
    in PopperContent (created by TooltipPopoverWrapper)
    in TooltipPopoverWrapper (created by Popover)
```

### Code

https://stackblitz.com/edit/reactstrap-uynews?embed=1&file=Example.js

zachary822:
I made a pull request to a possible fix #1687

bradchristensen:
@zachary822 I've noticed the issue you mentioned when trying to upgrade from Reactstrap 8.0.1 (last working version for tooltips using refs for me) to 8.2.0 as per the screenshot below.

<img src="https://user-images.githubusercontent.com/6006249/71350484-4e977800-25d6-11ea-94d4-8671a54cfac8.png" width="514.66" />

I agree that filtering out falsy values in the `getTarget` function seems like the way to fix this.

PR #1687, opened 2019-10-24
return array like when allElements parameter in getTarget is true

zachary822:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

#1686 The component expects the getTarget to return array like when allElement param is set to `true`.

### DIFF
--- a/src/__tests__/TooltipPopoverWrapper.spec.js
+++ b/src/__tests__/TooltipPopoverWrapper.spec.js
@@ -364,6 +364,7 @@ describe('Tooltip', () => {
 
   it('should not throw when passed a ref object as the target', () => {
     const targetObj = React.createRef();
+    targetObj.current = createSpyObj('div', ['addEventListener', 'removeEventListener']);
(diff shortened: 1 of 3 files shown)


zachary822:
The failed test comes from when the `ref.current` is `null`. Which happens before the component with ref is mounted. The only reason the test previously passed is because when [`Array.from`](https://github.com/reactstrap/reactstrap/blob/47843998b627aadcccfe71b092a194197fd2ce15/src/TooltipPopoverWrapper.js#L300) is called with a non-array-like `newTarget`, regardless of value, always returns an empty array.

TheSharpieOne:
Could you set `ref.current` to a dummy object with `addEventListener` as a mock fn. That should allow the test to pass. You can even expect the mock function to be called with some arguments in the test.

TheSharpieOne:
This has been merged right? GitHub seems to think this is still an open PR... but it doesn't give me options to merge it... only close
image

PR #1708, opened 2019-11-17
Allow usage of object for parameter className

auipga:
Parameter `className` can be an object for `classNames()` function.
See lines 55-56 (`const classes =`...)

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -10,7 +10,7 @@ const propTypes = {
   caret: PropTypes.bool,
   color: PropTypes.string,
   children: PropTypes.node,
-  className: PropTypes.oneOfType(PropTypes.string, PropTypes.object)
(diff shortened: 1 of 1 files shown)


TheSharpieOne:
`classnames` accepts even more types (such as an array), plus most (if not all) of the components use it and this PR only affects a single component.
We try to avoid exposing the usage of `classnames` by just excepting the standard type for `className`, which is just a string. You can use `classnames` yourself to generate the string (it shouldn't bloat your bundle since it's already included).
I have seen other libs simply not specify the PropType from `className`, allowing you to pass whatever (and if whatever is passed in works, then it works.) We might want to look into that?

auipga:
> not specify

sounds good to me

PR #1715, opened 2019-11-27
feat(ListGroup): Add horizontal prop

tcbegley:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

This PR adds a `horizontal` prop to `ListGroup` in order to add support for [horizontal list groups](https://getbootstrap.com/docs/4.4/components/list-group/#horizontal)

Horizontal list groups can't be mixed with flush list groups, so I made the decision to ignore the `horizontal` prop if `flush` is `true`. Happy to take guidance on this point though.

The `getHorizontalClass` function I based on the [analogous function](https://github.com/reactstrap/reactstrap/blob/master/src/Nav.js#L25-L33) in the source for `Nav` in order to be consistent with existing code.

### DIFF
--- a/src/ListGroup.js
+++ b/src/ListGroup.js
@@ -8,10 +8,20 @@ const propTypes = {
   flush: PropTypes.bool,
   className: PropTypes.string,
   cssModule: PropTypes.object,
+  horizontal: PropTypes.oneOfType([PropTypes.bool, PropTypes.string])
 };
 
 const defaultProps = {
   tag: 'ul'
+};
+
+const getHorizontalClass = horizontal => {
+  if (horizontal === false) {
+    return false;
+  } else if (horizontal === true || horizontal === "xs") {
+    return "list-group-horizontal";
+  }
+  return `list-group-horizontal-${horizontal}`;
 };
 
 const ListGroup = (props) => {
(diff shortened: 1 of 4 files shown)


src/ListGroup.js
```diff
@@ -8,24 +8,37 @@ const propTypes = {
...
+    horizontal,
     ...attributes
   } = props;
   const classes = mapToCssModules(classNames(
     className,
     'list-group',
-    flush ? 'list-group-flush' : false
+    // list-group-horizontal cannot currently be mixed with list-group-flush
+    // we only try to apply horizontal classes if flush is false
+    flush ? 'list-group-flush' : getHorizontalClass(horizontal)
```
TheSharpieOne: When `flush` and `horizontal` are not provided, they are `undefined`. This code would go into `getHorizontalClass(undefined)`. Since `undefined === false` is `false`, it would go to 
```
return `list-group-horizontal-${horizontal}`;
```
and produce "list-group-horizontal-undefined".

tcbegley: Oops, good catch thanks. I had meant to add a default value for `horizontal`. Added that now, are you happy with that as a fix?

PR #1720, opened 2019-12-03
feat(Row): Add row columns support

kyletsang:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

This is the new row columns feature from Bootstrap 4.4:
https://getbootstrap.com/docs/4.4/layout/grid/#row-columns

### DIFF
--- a/docs/lib/Components/LayoutPage.js
+++ b/docs/lib/Components/LayoutPage.js
@@ -4,7 +4,10 @@ import { PrismCode } from 'react-prism';
 import PageTitle from '../UI/PageTitle';
 import SectionTitle from '../UI/SectionTitle';
 import LayoutExample from '../examples/Layout';
+import LayoutRowColsExample from '../examples/LayoutRowCols';
+
 const LayoutExampleSource = require('!!raw-loader!../examples/Layout');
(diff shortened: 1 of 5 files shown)


TheSharpieOne:
Thoughts on removing the `col` prefix? Basically `<Row colMd="3">` vs `<Row md="3">`, but it would also simplify the code a bit so that the [mapping](https://github.com/reactstrap/reactstrap/pull/1720/files#diff-5bacd7cba2b21fde05e22db7b632cb3eR7) is not needed.
As it is now, you can pass in `widths`, but it always maps to those mappings. The reason `widths` exists it to allow for users to more easily define custom breakpoints like `xxs` or `xxl` (bootstrap's sass allows for this easily). Settings `widths` to `['xxs', 'xs', 'sm', 'md', 'lg', 'xl', 'xxl']` would allow this, but in the current code it would break since `i` would be of the `rowColsPropsWidthMapping` indexes (as well as mismatch)

kyletsang:
Sure, I will make those changes right now.

TheSharpieOne:
looks good. Thanks!

PR #1750, opened 2020-01-11
feat(Modal): Added static backdrop animation defined in Bootstrap 4.4

kyletsang:
- [x] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

An animation was added for static backdrop modals in Bootstrap 4.4.  This can be demoed at the following url:
https://getbootstrap.com/docs/4.4/components/modal/#static-backdrop

In addition to this, I fixed the esc key from closing a modal when the backdrop is set to static as per docs:
https://getbootstrap.com/docs/4.4/components/modal/#options

`Includes a modal-backdrop element. Alternatively, specify static for a backdrop which doesn't close the modal on click or on escape key press.`

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -100,6 +100,7 @@ class Modal extends React.Component {
     this.onOpened = this.onOpened.bind(this);
     this.onClosed = this.onClosed.bind(this);
     this.manageFocusAfterClose = this.manageFocusAfterClose.bind(this);
+    this.clearBackdropAnimationTimeout = this.clearBackdropAnimationTimeout.bind(this);
 
     this.state = {
       isOpen: false,
@@ -144,6 +145,8 @@ class Modal extends React.Component {
   }
 
   componentWillUnmount() {
+    this.clearBackdropAnimationTimeout();
+
     if (this.props.onExit) {
       this.props.onExit();
     }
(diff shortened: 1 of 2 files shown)


src/Modal.js
```diff
@@ -243,13 +251,26 @@ class Modal extends React.Component {
...
+        return;
+      }
+
       this.props.toggle(e);
     }
   }
 
+  handleStaticBackdropAnimation() {
+    this.setState({ showStaticBackdropAnimation: true });
+    setTimeout(() => {
```
TheSharpieOne: This is problematic if the component is unmounted before the timeout fires. It would cause `setState` to be executed on an unmounted component.
Maybe you can store the timeout on the component instance and clear it when the component unmounts.

kyletsang: Thanks. I made changes to address this.

ALEUT:
The behavior was changed since version 4.2.0: before I was able to close the modal using Esc button when keyboard={true} and backdrop="static", but now I can't. Is it expected?

TheSharpieOne:
Looks like it should allow `esc` to close the modal when `keyboard` is `true`. I opened a new issue to track this. Thanks you.

Also for reference: https://github.com/twbs/bootstrap/issues/29978

PR #1760, opened 2020-01-17
fix: revert breaking change

### LINKED ISSUE #1758: 8.2.0 -> 8.3.x was a breaking change
- components: `DropdownToggle`, `DropdownItem`, `DropdownMenu`
- reactstrap version `#8.3.1`
- import method `es`
- react version `#16.12.0`
- bootstrap version `#4.4.1`

I've got two issues:
1. My snapshots changed due to `<ForwardRef />`
2. I can't `extend` from `DropdownItem` (it may not be the only class that has this problem)

#### In browser error:
```
Uncaught TypeError: Super expression must either be null or a function
```

#### In jest tests error:

Snapshots failing:
```
    -       </DropdownToggle>
    -       <DropdownMenu
    -         flip={true}
    -         tag="div"
    -       >
    -         <DropdownItem
    +       </ForwardRef>
    +       <ForwardRef>
    +         <ForwardRef
```

Runtime Error:
```
TypeError: Class extends value #<Object> is not a constructor or null
```

### Code

```
import { DropdownItem } from 'reactstrap';
export class PopoverMenuItem extends DropdownItem {}
```

I *think*, that https://github.com/reactstrap/reactstrap/commit/64847df4a858c4ad7fc4c0f876fcff9a456ee55b caused this issue.

TheSharpieOne:
…text prop"

This reverts commit 64847df4a858c4ad7fc4c0f876fcff9a456ee55b.

64847df contained a breaking change, per semvar it needs to be removed and another version published on the same major ver which does not contain the breaking change.

Fixes #1758

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -261,6 +261,5 @@ class Dropdown extends React.Component {
 
 Dropdown.propTypes = propTypes;
 Dropdown.defaultProps = defaultProps;
-Dropdown.Context = DropdownContext;
 
 export default Dropdown;
--- a/src/DropdownItem.js
+++ b/src/DropdownItem.js
@@ -41,7 +41,7 @@ class DropdownItem extends React.Component {
     }
 
     if (this.props.toggle) {
-      this.props.context.toggle(e);
+      this.context.toggle(e);
     }
   }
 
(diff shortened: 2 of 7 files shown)


ajacksified:
Thank you!

PR #1771, opened 2020-01-28
feat(Progress): Allow passing in custom style object

DSBalaban:
Allows for more flexibility in styling

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

Ran into a scenario where I could use passing in styles in this manner. 
My use case involves mapping over a collection and creating slightly rounded progress bars which overlap (to cover the space created at the corners by border-radius), and I could use the ability of passing in `zIndex` values programmatically.

I realize this makes `Progress` "special", seeing as none of the other components accept the style prop. If you want me to do this for all components, let me know.

EDIT: `DropdownMenu › should not render multiple children when isOpen is false` fails, but it shouldn't be because of this PR.

### DIFF
--- a/src/Progress.js
+++ b/src/Progress.js
@@ -22,12 +22,14 @@ const propTypes = {
   className: PropTypes.string,
   barClassName: PropTypes.string,
   cssModule: PropTypes.object,
+  style: PropTypes.object,
 };
 
 const defaultProps = {
   tag: 'div',
   value: 0,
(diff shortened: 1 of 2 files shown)


kyletsang:
Unfortunately, this broke the ability to set style into the parent progress container.  Going to expose this feature under a new prop called `progressBarStyle`.

Issue #1496, opened 2019-05-21
0 value not displaying in DropdownToggle


tagyoureit:
- components: `DropdownToggle`
- reactstrap version `#8.0.0`
- import method `umd/csj/es`  es
- react version `#16.8.6`
- bootstrap version `#4.0.0`

### What is happening?
I'm passing props to a custom component that sometimes has the value of 0.  The render in my custom file looks like this:
```
    return (
      <div>
        <ButtonDropdown isOpen={this.state.dropdownOpen} toggle={this.toggleDropDown} disabled={this.state.disabled} >
          <DropdownToggle caret disabled={this.state.disabled}>
            {this.props.currentMinute}
          </DropdownToggle>
          <DropdownMenu>
            {positions()}
          </DropdownMenu>
        </ButtonDropdown>
      </div >
    )
```
When the value of `currentMinute=0` the display shows this:
![2019-05-20_21-17-49](https://user-images.githubusercontent.com/1646840/58068087-c5522b80-7b44-11e9-8939-29d7f2afc3ab.jpeg)
(In the yellow box the 0 comes from the parent file.)

### What should be happening?

0 should be displayed as a value just like the other values (

### Steps to reproduce issue

It appears that reactstrap is treating this 0 as a boolean value.  If I change the code to display based on an if/else `{this.props.currentMinute === 0 ? '0' : this.props.currentHour}` it does display properly.

![2019-05-20_21-25-24](https://user-images.githubusercontent.com/1646840/58068340-d8193000-7b45-11e9-8204-66d84d47ac28.jpeg)

Interestingly, it seems if I put _any_ other code as a child here it does render properly.  For example, `{this.props.currentMinute}{'anything else'}` will show `0anything else`.

### Error message in console

```
No error
```

### Code

https://stackblitz.com/edit/react-7zzcjw

TheSharpieOne:
Probably due to https://github.com/reactstrap/reactstrap/blob/c844ab1f0c4e02c2169e8ad19b18cc59bf4bd3c6/src/DropdownToggle.js#L63

For now, you can just use `this.props.currentMinute.toString()`

tagyoureit:
Thanks.  I settled on `{this.props.currentMinute}{''}` for now and it is working fine.

kyletsang:
Closing as fixed

PR #1780, opened 2020-02-04
fix(DropdownToggle): fix 0 value #1496

hedgecox:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

Replaced OR operator with null-coalescing operator to allow 0 value but not nullish values

#1496

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -60,7 +60,7 @@ class DropdownToggle extends React.Component {
         'nav-link': nav
       }
     ), cssModule);
-    const children = props.children || <span className="sr-only">{ariaLabel}</span>;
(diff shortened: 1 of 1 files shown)


hedgecox:
Just seen the travis SyntaxError:
"Support for the experimental syntax 'nullishCoalescingOperator' isn't currently enabled (63:37)"

Will close this for now

hedgecox:
Looking at the support it seems even though it's not fully supported in the browsers there is support for it in Babel: https://github.com/babel/proposals/issues/14

Babel transpiles it as so:
`const b = a ?? "else";`
to
`const b = a !== null && a !== void 0 ? a : "else";`

However, this might be overkill for the problem at hand

PR #1918, opened 2020-07-03
in solidarity

bpas247:
https://twitter.com/kwuchu/status/1267436741692133376
https://github.com/reactjs/reactjs.org
https://react-bootstrap.github.io/

Adds a banner for the Black Lives Matter movement, to spread
awareness for the cause.


TheSharpieOne:
Due to how the docs release is tied to the lib release (so that it is in lock-step and doesn't show features not yet released), a new version would need to be released in order to automatically update the docs.

PR #1941, opened 2020-08-05
Modal: Trap focus on opened component (FIX #1497 and FIX #1679)

### LINKED ISSUE #1497: Nested modals shift+tab focuses items of the underlying modal
- components: `modal`
- reactstrap version `#8.0.0`
- import method `es`
- react version `#16.8.6`
- bootstrap version `#4.3.1`

### What is happening?

As seen in the [Nested Modals](https://reactstrap.github.io/components/modals/) example of the documentation, if you open the nested modal `tab` into the focusable items ("Done" and "All Done") and then `shift+tab` back items of the previous modal are focused.

### What should be happening?

When pressing `shift+tab` the previously focused item should be focused.

### Steps to reproduce issue

1. Go to the [Modals](https://reactstrap.github.io/components/modals/) page
2. Find the "Nested Modals" example
3. Open the example modal
4. Open its nested modal
5. Press `tab` twice to focus the "All Done" button
6. Press `shift+tab` to focus the "Done" button
7. The "Cancel" button of the underlying modal will be focused

### LINKED ISSUE #1679: Shift and shift+tab focus items of the body when modal is open
- components: `modal`
- reactstrap version `#6.4.0`
- import method `es`
- react version `#16.4.0`
- bootstrap version `#4.1.1`

### What is happening?

If you open the modal, focus any browser element (e.g. address bar), and start to press the tab key to move the focus forward body elements get the focus as usual.

### What should be happening?

As described in https://www.w3.org/TR/wai-aria-practices/#dialog_modal elements outside the modal dialog should not be interactive:

> Windows under a modal dialog are inert. That is, users cannot interact with content outside an active dialog window. 

Additionally, here is an example of implementation: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

### Steps to reproduce the issue

1. Go to the http://reactstrap.github.io/components/modals/
2. Open the modal dialog
3. Set focus to the address bar of the browser
4. Press tab until focus goes inside the page

watinha:
- [X] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [X] There is an open issue which this change addresses


watinha:
Hey guys, if you require any change to these fixes, please let me know.

kyletsang (changes requested):
Shift tabbing on the nested modal examples doesn't work with this fix.  A better approach might be to only handle keyboard events on the topmost modal and ignore those below it.

watinha:
I will try to look into this issue as well.

watinha:
Hey @kyletsang, handleTab listeners were being processed for all Modals, as you mentioned. Changing the listeners to run on the top most Modal component only, seems to do the trick.

src/Modal.js
```diff
@@ -162,9 +165,34 @@ class Modal extends React.Component {
...
+            return ;
+    }
+
+    if (children.length > 0) {
+      ev.preventDefault();
+      ev.stopPropagation();
+      children[0].focus();
+    }
+  }
+
```
kyletsang, line 195: I don't believe we need this `trapFocus` function now that you're only handling keyboard on topmost modal?

ie. I think that your below fix of the early return in `handleTab` should be enough to fix the issue?

src/Modal.js
```diff
@@ -309,6 +338,7 @@ class Modal extends React.Component {
     }
 
     Modal.openCount += 1;
+    this.modalCount = Modal.openCount;
```
kyletsang, line 341: Let's rename`this.modalCount` to `this.modalIndex` to be more clear.  Also move the assignment above `Modal.openCount += 1` to make it 0-based.

src/Modal.js
watinha: The trapFocus function was implemented considering #1679. This issue reports that the user can set focus to elements outside of the Modal after focusing the address bar. I tried addressing both issues in the same commit, sorry for that lack of organization.

kyletsang: Ah ok makes sense now.  Can you fix the formatting here?  Indents are a bit off.   Also add some comments next to each branch inside the `trapFocus` function to let us know how it works and what each branch is preventing.

watinha:
@kyletsang just fixed the indentation, renaming modalCount -> modalIndex, and included comments in the trapFocus function.

src/Modal.js
```diff
@@ -162,9 +166,34 @@ class Modal extends React.Component {
       }
     }
 
+    document.removeEventListener('focus', this.trapFocus);
```
kyletsang, line 169:
```suggestion
    document.removeEventListener('focus', this.trapFocus, true);
```

kyletsang: Sorry missed this one.  Need to add `true` here since you added the original handler using capture.

kyletsang (changes requested):
Sorry one more change mentioned above.  Everything else looks good.

src/Modal.js
watinha: Hey @kyletsang, just included this change and included a test case for assuring the listeners are added/removed while mounting and unmounting the component.

kyletsang (approved):

kyletsang:
Thanks!

PR #1970, opened 2020-09-18
fix(Progress): Fix parent element props being spread to children

### LINKED ISSUE #1969: Duplicate id being rendered for Progress component
- components: `Progress`
- reactstrap version `8.6.0`

### What is happening?

If you pass an id to `Progress`, this will be added to both the outer progress element along with the inner progress bar element.  This is caused by a change in #1919.

### What should be happening?

id should not be duplicated.

### Code

`<Progress id="test" />`

kyletsang:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

Fixes #1969

This also adds to ability to customize the progress bar element via tag

### DIFF
--- a/src/Progress.js
+++ b/src/Progress.js
@@ -75,30 +75,34 @@ const Progress = (props) => {
     striped || animated ? 'progress-bar-striped' : null
   ), cssModule);
 
-  const ProgressBar = multi ? children : (
-    <div
-      {...attributes}
-      className={progressBarClasses}
-      style={{
-        ...style,
-        width: `${percent}%`,
-      }}
-      role="progressbar"
-      aria-valuenow={value}
-      aria-valuemin={min}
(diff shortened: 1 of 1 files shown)


src/Progress.js
```diff
@@ -75,30 +75,34 @@ const Progress = (props) => {
...
+        {...attributes}
+        {...progressBarProps} 
+      />
+    );
   }
 
   return (
-    <Tag {...attributes} className={progressClasses} children={ProgressBar} />
+    <Tag {...attributes} className={progressClasses}>
+      {multi ? children : <div {...progressBarProps} />}
```
kyletsang, line 104: I don't think having props `bar` and `multi` together would be valid, so this condition is moved down here.

PR #1973, opened 2020-10-05
Fix #1960 aria-haspopup prop types

### LINKED ISSUE #1960: DropdownToggle aria-haspopup should accept string value
- components: `DropdownToggle`
- reactstrap version `#8.5.1`
- react version `#16.13.1`
- bootstrap version `#4.5.2`
- typescript version `#3.9.7`
- @types/react version `#16.9.49`
- @types/reactstrap version `#8.5.1`

### What is happening?

I'm using typescript and when I try to set a string value (`listbox`) to `aria-haspopup` prop I got this error: 
```
Warning: Failed prop type: Invalid prop `aria-haspopup` of type `string` supplied to `DropdownToggle`, expected `boolean`.
```

I know this is related to the typings library (`@types/reactstrap`) but the prop types of `DropdownToggle` has wrong typing as well:

```js
#reactstrap/src/DropdownToggle.js

...
const propTypes = {
  ...
  'aria-haspopup': PropTypes.bool,
  ...
};
```

### What should be happening?

`aria-haspopup` prop should accept string value according to the spec https://www.w3.org/TR/wai-aria/#aria-haspopup

hedgecox:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

#1960

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -14,7 +14,7 @@ const propTypes = {
   cssModule: PropTypes.object,
   disabled: PropTypes.bool,
   onClick: PropTypes.func,
-  'aria-haspopup': PropTypes.bool,
+  'aria-haspopup': PropTypes.oneOfType(PropTypes.bool, PropTypes.oneOf(["menu", "listbox", "tree", "grid", "dialog"])),
   split: PropTypes.bool,
   tag: tagPropType,
   nav: PropTypes.bool,
@@ -33,6 +33,9 @@ class DropdownToggle extends React.Component {
   }
 
   onClick(e) {
+    console.log("Should i toggle here")
+    console.log(e)
+
     if (this.props.disabled || this.context.disabled) {
       e.preventDefault();
       return;
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -33,9 +33,6 @@ class DropdownToggle extends React.Component {
   }
 
   onClick(e) {
-    console.log("Should i toggle here")
-    console.log(e)
-
     if (this.props.disabled || this.context.disabled) {
       e.preventDefault();
       return;


kyletsang:
Don't think the values of tree, grid or dialog should be valid in the context of bootstrap's dropdown.  The listbox might be a valid one, but there are more issues that would need to be resolved to support that.

https://www.w3.org/TR/wai-aria/#aria-haspopup

The ARIA specs mention:
> Authors MUST ensure that the role of the element that serves as the container for the popup content is menu, listbox, tree, grid, or dialog, and that the value of aria-haspopup matches the role of the popup container.

- The dropdown menu has its role hardcoded to `menu`

> List boxes contain children whose role is option.

- Dropdown items has their roles set as `menuitem`

hedgecox:
@kyletsang Good point in terms of tree, grid and dialog. 

If we were to implement listbox as an option on the dropdownToggle, we could add `role` as a prop to dropdownMenu and DropdownItem, allowing the user to pass in `"listbox"` and `"option"`. However this isn't ideal as the user would have to set the values in 3+ different places to change the aria-haspopup - not sure if there is a better solution?

kyletsang:
`Dropdown` does contain a `DropdownContext` that we can use to pass props down.  Maybe adding a `menuRole` prop on the `Dropdown` and passing that down is the way to go.

src/DropdownToggle.js
```diff
@@ -33,6 +33,9 @@ class DropdownToggle extends React.Component {
   }
 
   onClick(e) {
+    console.log("Should i toggle here")
+    console.log(e)
+
```
TheSharpieOne, line 38:
```suggestion
```

kyletsang:
@hedgecox are you still interested in working on this?

andyseracuse:
perhaps I'm mistaken, but It is my understanding that using the dropdown context will set all the dropdowns in the project to whatever value you set.  In other words, if I set a value in the dropdown context using a dropdown's props, and I have multiple dropdowns with different uses in my application, they will all be set to whatever value I set in the one dropdown.  This doesn't seem like desired behavior.  Am i misunderstanding something?  Thanks for any assistance!

kyletsang:
@andyseracuse, the context will only affect anything in the tree within the `<DropdownContext.Provider>`.  They won't affect each other.

andyseracuse:
@kyletsang aha!  got it.  thanks for the clarification.  I'll try to get that implemented.

Issue #1340, opened 2018-12-28
Usage of legacy context API


thetric:
- components: `TabPane`, `Dropdown(,Item,Menu,Toggle)`, `Manager`, `PopperTargetHelper`, `CarousellItem` (see https://github.com/reactstrap/reactstrap/search?q=contextTypes&unscoped_q=contextTypes)
- reactstrap version `6.5.0`
- import method: `es`
- react version `16.7.0`
- bootstrap version `4.2.1`

### What is happening?
When using `<React.StrictMode>` Reacts advises against using the legacy context API:

```
Warning: Legacy context API has been detected within a strict-mode tree: 
    in StrictMode (at App.tsx:27)
    in App (at src/index.tsx:7)

Please update the following components: Dropdown, DropdownItem, DropdownMenu, DropdownToggle, Manager

Learn more about this warning here:
https://fb.me/react-strict-mode-warnings
```

> _From https://reactjs.org/docs/context.html#legacy-api:_
>
> React previously shipped with an experimental context API. The old API will be supported in all 16.x releases, but applications using it should migrate to the new version. The legacy API will be removed in a future major React version. [...]

### What should be happening?
Reactstrap should be using the new Context API.

### Steps to reproduce issue
Wrap an example app using some of the components mentions at the beginning in `React.StrictMode` and start it in development mode.

### Code

https://stackblitz.com/edit/reactstrap-saxswj?file=Example.js

thetric:
The `Modal` component uses deprecated lifecycle hooks:

```
componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Modal

componentWillUpdate: Please update the following components to use componentDidUpdate instead: Modal
```

gergely-nagy:
I'll submit a PR for it!

iegik:
Duplicate of https://github.com/reactjs/react-transition-group/issues/429

armenzg:
@gergely-nagy any progress?

jonathan-richer:
Same thing on `Collapse`:
![image](https://user-images.githubusercontent.com/36356746/83686057-9378ff00-a5b7-11ea-92f9-ef3691d2c1db.png)

sumanthratna:
I got this with `Tooltip`, if that helps.

MA-Ahmad:
I got this warning with Carousel.

iamandrewluca:
Anyone interested in fixing this?
[...]

PR #1992, opened 2020-10-21
Fix strict mode warnings for Collapse component

GabrielMajeri:
Updates the `react-transition-group` dependency to help with compatibility with React strict mode.

- [X] Bug fix for part of #1340 
- [X] Chore since it also updates a dependency

Helps towards fixing #1340 
Supersedes #1907

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -62,7 +62,7 @@ class Collapse extends Component {
       dimension: null,
     };
 
-    this.nodeRef = React.createRef();
+    this.nodeRef = props.innerRef || React.createRef();
 
     ['onEntering', 'onEntered', 'onExit', 'onExiting', 'onExited'].forEach(
       (name) => {
@@ -72,8 +72,7 @@ class Collapse extends Component {
   }
 
   getNode() {
-    const nodeRef = this.props.innerRef || this.nodeRef;
-    return nodeRef.current;
+    return this.nodeRef.current;
   }
 
   onEntering(_, isAppearing) {
@@ -125,8 +124,6 @@ class Collapse extends Component {
       ...otherProps
     } = this.props;
 
-    const nodeRef = innerRef || this.nodeRef;
-
     const { dimension } = this.state;
 
     const transitionProps = pick(otherProps, TransitionPropTypeKeys);
@@ -135,7 +132,7 @@ class Collapse extends Component {
       <Transition
         {...transitionProps}
         in={isOpen}
-        nodeRef={nodeRef}
+        nodeRef={this.nodeRef}
         onEntering={this.onEntering}
         onEntered={this.onEntered}
         onExit={this.onExit}
@@ -162,7 +159,7 @@ class Collapse extends Component {
               {...childProps}
               style={{ ...childProps.style, ...style }}
               className={classes}
-              ref={nodeRef}
+              ref={this.nodeRef}
             >
               {children}
             </Tag>
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -71,10 +71,6 @@ class Collapse extends Component {
     );
   }
 
-  getNode() {
-    return this.nodeRef.current;
-  }
-
   onEntering(_, isAppearing) {
     const node = this.getNode();
     this.setState({ dimension: this.getDimension(node) });
@@ -105,6 +101,10 @@ class Collapse extends Component {
     const node = this.getNode();
     this.setState({ dimension: null });
     this.props.onExited(node);
+  }
+
+  getNode() {
+    return this.nodeRef.current;
   }
 
   getDimension(node) {
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -29,11 +29,7 @@ const propTypes = {
   navbar: PropTypes.bool,
   /** Change underlying component's CSS base class name */
   cssModule: PropTypes.object,
-  innerRef: PropTypes.oneOfType([
-    PropTypes.func,
-    PropTypes.string,
-    PropTypes.object,
-  ]),
+  innerRef: PropTypes.shape({ current: PropTypes.object }),
 };
 
 const defaultProps = {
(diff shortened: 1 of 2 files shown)


kyletsang:
Would also need use the nodeRef API, since RTG still uses findDOMNode under the hood.

GabrielMajeri:
> Would also need use the nodeRef API, since RTG still uses findDOMNode under the hood.

I've tested the specific component with strict mode and no warnings were shown. How can I can reproduce the situation to get strict mode warnings with RTG?

kyletsang:
You'll need to trigger an animation to see the warning, like clicking the toggle button on the collapse examples.

GabrielMajeri:
@kyletsang Thanks for the tips, I've updated the code to use the new `nodeRef` API.

src/Collapse.js
```diff
@@ -55,34 +55,46 @@ class Collapse extends Component {
...
     ['onEntering', 'onEntered', 'onExit', 'onExiting', 'onExited'].forEach((name) => {
       this[name] = this[name].bind(this);
     });
   }
 
-  onEntering(node, isAppearing) {
+  getNode() {
+    const nodeRef = this.props.innerRef || this.nodeRef;
+    return nodeRef.current;
+  }
```
kyletsang, line 68: This would need to consider callback refs as well.  Also, looking at the proptypes, it looks like it also allows legacy refs.

package.json
```diff
@@ -262,7 +262,7 @@
     "classnames": "^2.2.3",
     "prop-types": "^15.5.8",
     "react-popper": "^1.3.6",
-    "react-transition-group": "^2.3.1"
+    "react-transition-group": "^4.4.1"
```
kyletsang, line 265: Just realized that the latest RTG requires a higher peer dependency of react 16.6, whereas reactstrap is at 16.3.  This would have to be considered a breaking change.

GabrielMajeri: Should I pick a version of RTG which doesn't require anything higher than 16.3, or should I go through the process of bumping the major version of Reactstrap?

src/Collapse.js
GabrielMajeri: I will look into supporting callback refs.

As for legacy refs, should I check if the type of `props.innerRef` is a string, and add a workaround for it?

package.json
kyletsang: I believe only 4.4.x resolves the issues with findDOMNode, so we're stuck there.  Will need to get @TheSharpieOne's feedback here on major version bumps.

src/Collapse.js
kyletsang: If we do the major version bump, I think we should drop legacy refs.

package.json
TheSharpieOne: major version bumps are fine. But we should try to group all breaking changes. I know there are other breaking changes that could be made (such as forward refs).

src/Collapse.js
GabrielMajeri: The new RTG `nodeRef` API only supports object refs (I've tried passing it a function ref and it triggered a proptypes warning).

I've updated the prop types on `Collapse` to only allow object refs. Is this a good solution?

package.json
GabrielMajeri: Should I bump the major version in this PR, or is this something to be done in another PR?

src/Collapse.js
kyletsang: Ugh... didn't realize it only took ref objects.  We should still allow both callback and object refs to be passed in through `innerRef` though... Unfortunately I only know of how to merge refs via hooks.  Unsure what approach to take here.

cesarve77:
Why is not merged?

GabrielMajeri:
@cesarve77 the changes would increase the minimum required React version from `16.3` to `16.6`. I can make the changes but I requested some feedback from the maintainers to know if this is the right thing to do.

stdavis:
Is there anything that I can do to help get this merged? The findDOMNode deprecation warning in the console makes me worry.

GabrielMajeri:
@stdavis I've updated the PR to the latest `master`. I've noticed that the library has been updated to use `react-transition-group` version 4.4.2, hence this isn't exactly a "breaking change" anymore.

I'd ask the library's maintainers to take a look at the changes again and tell me if there's anything else that needs to be done to get this merged. It should fix the deprecation warnings, as far as I can tell.

stdavis (approved):

stdavis:
This looks good to me! Hopefully it gets some attention. Thanks, @GabrielMajeri !

karna41317:
Any updates on this ?

RichardLindhout:
Please merge this.

davidacevedo:
Seems like the Collapse documentation isn't working after the above changes: https://deploy-preview-1992--reactstrap.netlify.app/?path=/docs/components-collapse--collapse

src/Collapse.js
```diff
@@ -53,6 +49,8 @@ class Collapse extends Component {
       dimension: null
     };
 
+    this.nodeRef = React.createRef();
```
davidacevedo, line 52:
```suggestion
    this.nodeRef = props.innerRef || React.createRef();
```

src/Collapse.js
```diff
@@ -62,29 +60,39 @@ class Collapse extends Component {
     return this.props.horizontal ? node.scrollWidth : node.scrollHeight;
   }
 
-  onEntering(node, isAppearing) {
+  getNode() {
+    const nodeRef = this.props.innerRef || this.nodeRef;
+    return nodeRef.current;
```
davidacevedo, line 65: From the above recommendation, we can then do
```suggestion
  return this.nodeRef.current;
```
For here and in the render method

src/Collapse.js
```diff
@@ -102,6 +110,8 @@ class Collapse extends Component {
       ...otherProps
     } = this.props;
 
+    const nodeRef = innerRef || this.nodeRef;
```
davidacevedo, line 113: Wouldn't be needed anymore after the above changes.

davidacevedo:
Everything looks good, just a few changes and maybe a rebase is needed as the documentation seems broken for the Collapse component.

GabrielMajeri:
Thank you for the feedback @davidacevedo. I've committed the corresponding changes and rebased the pull request against the latest `master`.

GabrielMajeri:
I've pushed a new commit, addressing the lint and formatting issues encountered in CI.

davidacevedo (approved):

PR #1995, opened 2020-10-26
Fix click handlers on menu items not working when Dropdown is used in shadow DOM

yordan-pavlov:
Currently when a Dropdown is used in shadow DOM, the click handlers on menu items do not fire.
This happens because the dropdown menu closes too soon, because the condition contained in the `handleDocumentClick` function which is intended to prevent the dropdown from closing in this case doesn't trigger.
This happens because the click handler is registered on the document (outside the shadow DOM) so any events originating from shadow DOM which bubble up to the document will have the target property set to the shadow DOM container rather than the actual element from which the event started.

This PR changes the `handleDocumentClick` function to use the `composedPath` method (if available) to correctly identify the target element even when the event originated from shadow DOM, very similar to how it's done in the `show` function here  
https://github.com/reactstrap/reactstrap/blob/master/src/TooltipPopoverWrapper.js#L167

Unfortunately, writing a test for this requires a newer version of jsdom which would require upgrading the create react app scripts to the just released version 4.0.0.

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -113,8 +113,12 @@ class Dropdown extends React.Component {
   handleDocumentClick(e) {
     if (e && (e.which === 3 || (e.type === 'keyup' && e.which !== keyCodes.tab))) return;
     const container = this.getContainer();
-
-    if (container.contains(e.target) && container !== e.target && (e.type !== 'keyup' || e.which === keyCodes.tab)) {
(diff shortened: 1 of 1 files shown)


kyletsang (approved):
Looks ok to me

PR #2007, opened 2020-11-02
Fix reference to CSSModule

jmandel:
The 8.7.0 release has a broken import of CSSModuile at https://github.com/reactstrap/reactstrap/blob/8.7.0/types/lib/Toast.d.ts#L2

- [x] Bug fix

### DIFF
--- a/types/lib/Toast.d.ts
+++ b/types/lib/Toast.d.ts
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { CSSModule } from '../index';
+import { CSSModule } from './index';
 import { FadeProps } from './Fade';
 
 export interface ToastProps extends React.HTMLAttributes<HTMLElement> {


kyletsang (approved):

kyletsang:
Thanks

jmandel:
Sure! (Would be great if there were tests to automatically check validity of these definitions -- I wasn't sure where/how you'd want to add this though.)

kyletsang:
We have a test file reactstrap-tests.tsx inside the types folder where you can add type tests.

Unfortunately this bug one slipped through due to the duplicate index.d.ts in the types folder that contained another CSSModule def.  I think I had this dupe file in there to get the dtslint to work properly.  Going to look into cleaning this up to prevent this from slipping again.

PR #2016, opened 2020-11-09
feat(Dropdown): Added container prop to DropdownMenu using React.portal

hsource:
- [ ] Bug fix 
- [X] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [X] My change requires a change to the documentation.
- - [X] I have updated the documentation accordingly.

This is an improvement of #1513

When a Dropdown is inside a small element with overflow: hidden, it's useful to have the dropdown display directly inside of body or another container. I added the same `container` prop to DropdownMenu. Using a React Portal, this allows the dropdown to have body or any other tag as the parent.

The DropdownMenu passes its own `ref` back to the parent dropdown via an `onMenuRef` callback in the Dropdown context so that clicks outside the DropdownMenu will properly close the menu.

| Without `container` | With `container |
| --- | --- |
| Before container | Testing container |

### DIFF
--- a/docs/lib/Components/DropdownsPage.js
+++ b/docs/lib/Components/DropdownsPage.js
@@ -16,11 +16,13 @@ import DropdownSizingExample from '../examples/DropdownSizing';
 import CustomDropdownExample from '../examples/CustomDropdown';
 import DropdownUncontrolledExample from '../examples/DropdownUncontrolled';
 import DropdownSetActiveFromChildExample from '../examples/DropdownSetActiveFromChild';
+import DropdownContainerExample from '../examples/DropdownContainer';
 
 const DropdownExampleSource = require('!!raw-loader!../examples/Dropdown');
 const CustomDropdownExampleSource = require('!!raw-loader!../examples/CustomDropdown');
 const DropdownUncontrolledExampleSource = require('!!raw-loader!../examples/DropdownUncontrolled');
 const DropdownSetActiveFromChildSource = require('!!raw-loader!../examples/DropdownSetActiveFromChild');
+const DropdownContainerSource = require('!!raw-loader!../examples/DropdownContainer');
 
 export default class DropdownPage extends React.Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


hsource:
I just realized this broke some tests! I can take a quick look later today.

hsource:
Fixed the tests! Ready for review again

kyletsang:
@TheSharpieOne, is this something you'd like to add to the API?  If so, I can review + test this.

TheSharpieOne:
@kyletsang yes

docs/lib/Components/DropdownsPage.js
```diff
@@ -406,6 +412,19 @@ DropdownItem.propTypes = {
             {DropdownSetActiveFromChildSource}
           </PrismCode>
         </pre>
+
+        <SectionTitle>container</SectionTitle>
+        <p>
+          Use the <code>container</code> prop to allow the dropdown menu to be placed inside an alternate container through a React Portal. This can be used to allow the dropdown menu to escape a container with the style `overflow: hidden`.
```
kyletsang, line 418: Let's wrap "overflow: hidden" in code tags.

kyletsang (changes requested):
`getMenuItems` inside `Dropdown` will need to account for the change in the location of the menu so it can query the menu items.  Currently, it won't find anything, so keyboard controls don't work.

hsource:
> `getMenuItems` inside `Dropdown` will need to account for the change in the location of the menu so it can query the menu items. Currently, it won't find anything, so keyboard controls don't work.

Fixed this! Here's experiments with tab, enter, and the down arrow.

Keyboard check

src/Dropdown.js
```diff
@@ -88,14 +97,18 @@ class Dropdown extends React.Component {
...
+
   getMenuCtrl() {
     if (this._$menuCtrl) return this._$menuCtrl;
     this._$menuCtrl = this.getContainer().querySelector('[aria-expanded]');
     return this._$menuCtrl;
   }
 
   getMenuItems() {
-    return [].slice.call(this.getContainer().querySelectorAll('[role="menuitem"]'));
+    return [].slice.call(this.getMenu().querySelectorAll('[role="menuitem"]'));
```
kyletsang, line 111:
```suggestion
    const menuContainer = this.getMenu() || this.getContainer();
    return [].slice.call(menuContainer.querySelectorAll('[role="menuitem"]'));
```

Tests are failing here, but it's working in the docs site.  Seems like an issue with timing in the test environment.  Normally, the dropdown is supposed to toggle, then render the popper element (which assigns the menu ref), then trigger the focus code. In the tests, this menu ref is not yet assigned...

Anyways, I think we can safely add the old container as a fallback here just to satisfy the tests.  I don't see this being triggered when testing on the docs site.  This seems hacky, but it does follow whatever existing behavior was there previously anyway.

kyletsang:
@hsource, aside from the 1 issue above, this looks pretty good!  Should be good to merge after you make the last change.

src/Dropdown.js
hsource: Thanks again for the review! I pushed your fix and added a comment explaining why we do it. I agree it seems hacky, but I don't think it should ever break in real-life usage.

kyletsang (approved):

Issue #2028, opened 2020-12-01
Carousel related components do not attach any attribute provided as a prop to them


afzaalahmad:
### Issue description

- components: `Carousel`, `CarouselCaption`, `Carousel Control` and `Carousel Item`
- reactstrap version `#8.7.1`
- import method `import Carousel from 'reactstrap/lib/Carousel'` 
- react version `#17.0.1`
- bootstrap version `#4.5.2`

### What is happening?
Assigning an id attribute to Carousel and CarouselControls but it is not setting to their respective top level div inside component

### What should be happening?
it should attach any attribute provided as a prop directly to the top level div.

### Steps to reproduce issue
Use Carousel and set id prop, you will see in DOM that there is no id being assigned to div of carousel

### Code
https://stackblitz.com/edit/reactstrap-1jk2hy?file=Example.js

PR #2029, opened 2020-12-01
fix(#2028): pass extra attributes after destructuring to carousel

afzaalahmad:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [x ] There is an open issue which this change addresses

#2028

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -166,7 +166,7 @@ class Carousel extends React.Component {
   }
 
   render() {
-    const { cssModule, slide, className } = this.props;
+    const { cssModule, slide, className, ...attributes } = this.props;
     const outerClasses = mapToCssModules(classNames(
       className,
       'carousel',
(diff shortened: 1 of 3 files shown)


kyletsang (approved):

src/CarouselIndicators.js
```diff
@@ -13,6 +13,7 @@ const CarouselIndicators = (props) => {
     ), cssModule);
     return (
       <li
+        {...item.props}
```
kyletsang, line 16: The main purpose of this change is to introduce ids into the carousel indicators?

afzaalahmad: @kyletsang ids and data-testid to every indicator, to get indicator through testid and fire events on them.

afzaalahmad: @kyletsang when this PR will gets merged and new release will be published?

kyletsang: I guess my concern here would be the naming of `props`.  The usage of carousel is pretty open-ended, so could potentially pass an array of components as `items` to `CarouselIndicators` which will conflict.  Maybe it's better to allow users to pass children into `CarouselIndicators` so they can control the rendering?

Thoughts @TheSharpieOne?

Issue #2031, opened 2020-12-06
List inline


johnathanmiller:
Any chance of supporting [list inline](https://getbootstrap.com/docs/4.0/content/typography/#inline)? Horizontal list group doesn't quite cut it.

kyletsang:
Sounds like a good idea.  Interested in sending a PR?

johnathanmiller:
Sure! I'll get one going.

PR #2033, opened 2020-12-09
feat(ListInline): add List and ListInlineItem components

johnathanmiller:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

#2031


src/ListInlineItem.js
```diff
@@ -0,0 +1,36 @@
...
+    tag: tagPropType,
+    className: PropTypes.string,
+    cssModule: PropTypes.object
+};
+
+const defaultProps ={
+    tag: 'li'
+};
+
+const ListInlineItem = (props) => {
```
kyletsang, line 16: Let's also use a `forwardRef` here

src/ListInline.js
```diff
@@ -0,0 +1,36 @@
...
+    tag: tagPropType,
+    className: PropTypes.string,
+    cssModule: PropTypes.object
+};
+
+const defaultProps ={
+    tag: 'ul'
+};
+
+const ListInline = (props) => {
```
kyletsang, line 16: Let's make this component a little bit more generic so we can cover `.list-unstyled` here too, for example:
`<List type="unstyled">` and `<List type="inline">`

Also use a `forwardRef` here

kyletsang (changes requested):

johnathanmiller:
Summary of changes:
- Renamed `ListInline` to `List`.
- `List` now accepts a prop of `type`
- `List` and `ListInlineItem` both accept `ref`
- Docs now have a `List` page which contains both `unstyled` and `inline` list examples
- Updated tests

src/List.js
```diff
@@ -0,0 +1,40 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { mapToCssModules, tagPropType } from './utils';
+
+const propTypes = {
+  tag: tagPropType,
+  className: PropTypes.string,
+  cssModule: PropTypes.object,
+  type: PropTypes.string,
+  innerRef: PropTypes.oneOfType([PropTypes.object, PropTypes.string, PropTypes.func])
```
kyletsang, line 11: Let's use `forwardRef` instead of `innerRef` since this is more suited for function components:
https://reactjs.org/docs/forwarding-refs.html

So below would be
`const List= React.forwardRef((props, ref) => ...`

kyletsang:
Looks good.  One final fix for the `forwardRef` for both components and we should be good to go.

kyletsang (approved):
Looks good, thanks!

Issue #1990, opened 2020-10-20
Incorrect position of popover on button with icon 


Mati365:
Popover is not attached to center of button but on inner icon instead. It is propably caused by wrong target picking in method below.

https://github.com/reactstrap/reactstrap/blob/2a47626dc088f1983239f606c91f6294c7e756d7/src/TooltipPopoverWrapper.js#L166

Pic rel:
![bug](https://imgur.com/iBVIaXx.png)

kyletsang:
Can you provide a code repro?

Mati365:
> Can you provide a code repro?

Just edit example on popover page (for example add div inside button and click on newly created div)

Infonautica:
Same goes with Tooltip component on Button with Icon

kyletsang:
Fixed in #2038

PR #2038, opened 2020-12-17
fix(#1990): changes priority of target element passed to popper component

watinha:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/TooltipPopoverWrapper.js
+++ b/src/TooltipPopoverWrapper.js
@@ -160,10 +160,19 @@ class TooltipPopoverWrapper extends React.Component {
     return delay;
   }
 
+  getCurrentTarget(target) {
+    if (!target)
+      return null;
+    const index = this._targets.indexOf(target);
+    if (index >= 0)
+      return this._targets[index];
+    return this.getCurrentTarget(target.parentElement);
+  }
+
   show(e) {
     if (!this.props.isOpen) {
       this.clearShowTimeout();
-      this.currentTargetElement = e ? e.currentTarget || e.target : null;
+      this.currentTargetElement = e ? e.currentTarget || this.getCurrentTarget(e.target) : null;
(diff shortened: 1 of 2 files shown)


src/TooltipPopoverWrapper.js
```diff
@@ -351,7 +351,7 @@ class TooltipPopoverWrapper extends React.Component {
     return (
       <PopperContent
         className={className}
-        target={this.currentTargetElement || this._targets[0]}
+        target={this._targets[0] || this.currentTargetElement}
```
kyletsang, line 354: Unfortunately, this would reintroduce #1185.

watinha: @kyletsang, I am thinking I could try to use the focused element to identify which of the _targets is associated to it. Do you think this approach might solve this issue?

kyletsang: The issue here is with the button click.  `TooltipPopoverWrapper` adds the click listener to `document`, so the code below

https://github.com/reactstrap/reactstrap/blob/a97a8345f3998e9388d42550b8e70aaa62a5be50/src/TooltipPopoverWrapper.js#L166

would result in `this.currentTargetElement` being set to `e.target` due to `e.currentTarget` being null.

I don't believe we should ever be using `e.target` here, so we should modify the code to scan the current list of targets and determine which target `e.target` falls under. This should only be done if `e.currentTarget` is null.

watinha: Hey @kyletsang, I tried implementing exactly what you said. Please, let me know if this was the implementation you were expecting.

kyletsang: Looks good!  Thanks

kyletsang (approved):

PR #2039, opened 2020-12-19
feat: add positionFixed prop to Tooltip and Popover

hsource:
- [ ] Bug fix 
- [X] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [X] My change requires a change to the documentation.
  - [X] I have updated the documentation accordingly.
- [X] My change requires a change to [Typescript typings](./types/lib).
  - [X] I have updated the typings accordingly.

react-popper has a positionFixed prop that allows the popover itself to have the `position: fixed` style, so that it doesn't flicker when applied to an element that itself is position-fixed. This change allows users of Tooltip/Popover components to specify this and pass it on to prevent flickering.

| Before | After |
| --- | --- |
| Before | After |

### DIFF
--- a/src/PopperContent.js
+++ b/src/PopperContent.js
@@ -24,6 +24,7 @@ const propTypes = {
   container: targetPropType,
   target: targetPropType.isRequired,
   modifiers: PropTypes.object,
+  positionFixed: PropTypes.bool,
   boundariesElement: PropTypes.oneOfType([PropTypes.string, DOMElement]),
   onClosed: PropTypes.func,
   fade: PropTypes.bool,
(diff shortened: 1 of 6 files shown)


src/PopperContent.js
```diff
@@ -108,6 +108,7 @@ class PopperContent extends React.Component {
       tag,
       container,
       modifiers,
+      positionFixed,
```
kyletsang, line 111: Let's add this to `propTypes` too

hsource: Done!

hsource:
Added `positionFixed` to PropTypes! Should be ready for review now

kyletsang (approved):

PR #2041, opened 2020-12-22
fix: Make tooltips fade out

hsource:
- [X] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

Tooltips currently don't actually fade out, because as soon as `isOpen` is set to false, the render function in TooltipPopoverWrapper returns null.

This change makes it so that it returns something in the render function even when `isOpen` is false. We still only call `this.updateTarget()` when the tooltip is open. This ensures that the component still works for tooltips that open on a target appearing only after initial page rendering.

| Before | After |
| --- | --- |
| https://user-images.githubusercontent.com/2937410/102849462-0fa82800-43cc-11eb-92bb-b8f1d7a7443b.mov | https://user-images.githubusercontent.com/2937410/102849480-1767cc80-43cc-11eb-8030-40cab32414ed.mov |

### DIFF
--- a/src/TooltipPopoverWrapper.js
+++ b/src/TooltipPopoverWrapper.js
@@ -317,11 +317,9 @@ class TooltipPopoverWrapper extends React.Component {
   }
 
   render() {
-    if (!this.props.isOpen) {
-      return null;
+    if (this.props.isOpen) {
+      this.updateTarget();
     }
-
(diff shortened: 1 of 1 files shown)


kyletsang (approved):

kyletsang:
Thanks for the PRs!

PR #2059, opened 2020-12-28
fix: Prevent popover/tooltip rendering if target not resolved

### LINKED ISSUE #2058: Tooltip error with ref target in 8.8.0
- components: `Tooltip`
- reactstrap version `#8.8.0`
- import method `es` 
- react version `#17.0.1`
- bootstrap version `#4`

### What is happening?

When `ref` is passed to the Tooltip component, it displays a warning:
```
Warning: Failed prop type: The prop `target` is marked as required in `PopperContent`, but its value is `undefined`.
```

### How to reproduce

```javascript
import React, { useRef } from 'react';
import { UncontrolledTooltip } from 'reactstrap';

const tooltipRef = useRef();

<i ref={tooltipRef} />                                                           
<UncontrolledTooltip  target={tooltipRef}>                      
  Example tooltip       
</UncontrolledTooltip> 
```

kyletsang:
Fixes #2058

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/TooltipPopoverWrapper.js
+++ b/src/TooltipPopoverWrapper.js
@@ -331,6 +331,11 @@ class TooltipPopoverWrapper extends React.Component {
       this.updateTarget();
     }
 
+    const target = this.currentTargetElement || this._targets[0];
+    if (!target) {
+      return null;
+    }
+
     const {
       className,
       cssModule,
@@ -360,7 +365,7 @@ class TooltipPopoverWrapper extends React.Component {
     return (
       <PopperContent
         className={className}
-        target={this.currentTargetElement || this._targets[0]}
+        target={target}
         isOpen={isOpen}
         hideArrow={hideArrow}
         boundariesElement={boundariesElement}


willvlad:
I've just run into the same issue, but I don't think the above fix will solve it. The error looks like it might be caused by wrong use of an array, when the target is just a string: https://github.com/reactstrap/reactstrap/issues/1686

The target IS being passed, but the error is triggered by it not being an array

kyletsang:
@willvlad, are you seeing this exact message or something else?
`Warning: Failed prop type: The prop `target` is marked as required in `PopperContent`, but its value is `undefined`.`

The issue is passing `undefined` to `PopperContent` which is used internally for popovers and tooltips.

The bug is a regression from #2041.  In previous versions, if the popover/tooltip is not open, `PopperContent` is not rendered.  There is a function `updateTarget` that resolves the correct target prop based on the target prop you pass into the popover or tooltip, but this didn't get called yet prior to rendering `PopperContent` and this is what is causing the error. 

The fix here guards against that because if we have no target yet, there's no point rendering `PopperContent`.  `updateTarget` will get called in `componentDidMount` and `render` when the popover or tooltip is opened.

willvlad:
The message I see is "Failed prop type: The prop `target` is marked as required in `PopperContent`, but its value is `undefined`." even though the target is being passed

kyletsang:
Ok, should be the same bug.  Target is passed to tooltip, but internally, it passes something else to PopperContent, so it's nothing wrong with your code.

Anyways, I've released 8.8.1 that should fix this.  Give it some time to propagate to npm.

willvlad:
Thanks, confirm that 8.8.1 fixed it for me!

PR #2077, opened 2021-01-15
feat(Dropdown, DropdownMenu, DropdownToggle): allow Dropdown to have …

andyseracuse:
…listbox or menu role

Dropdown Menu should allow for listbox or menu roles.  Before this commit, the Dropdown could only have the menu role.  After this commit, a user can add a menuRole property to the Dropdown component which will set a corresponding aria-haspopop property on the DropdownToggle, role property on the DropdownMenu, and role property on the Dropdown Items

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 


src/Dropdown.js
```diff
@@ -35,7 +36,8 @@ const defaultProps = {
   active: false,
   addonType: false,
   inNavbar: false,
-  setActiveFromChild: false
+  setActiveFromChild: false,
+  menuRole: true
```
kyletsang, line 40: Don't think we need this.  Just to make this behavior opt-in

src/DropdownToggle.js
```diff
@@ -49,6 +47,16 @@ class DropdownToggle extends React.Component {
     this.context.toggle(e);
   }
 
+  getRole() {
+    const menuRole = this.context.menuRole
+    if(menuRole === true || menuRole === 'menu') {
```
kyletsang, line 52: We can just return `menuRole` here since all the supported values in `menuRole` are also applicable to `aria-haspopup`.  So this whole function can be simplified to return `context.menuRole` if it exists, otherwise return the `aria-haspopup` prop.

kyletsang (changes requested):

src/Dropdown.js
```diff
@@ -112,7 +115,7 @@ class Dropdown extends React.Component {
     // be null, but it is sometimes null in tests. To mitigate that, we just
     // use `this.getContainer()` as the fallback `menuContainer`.
     const menuContainer = this.getMenu() || this.getContainer();
-    return [].slice.call(menuContainer.querySelectorAll('[role="menuitem"]'));
+    return [].slice.call(menuContainer.querySelectorAll('[role="menuitem"], [role="option"]'));
```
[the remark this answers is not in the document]
andyseracuse: totally possible I'm wrong, but I did assign menuRole to default to true, so won't it always be defined?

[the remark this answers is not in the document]
kyletsang: I'm thinking we make menuRole opt-in just so we can leave the behavior as it was previously for existing users.  I think I need to amend one of my review comments in dropdown toggle.

src/DropdownToggle.js
kyletsang: Scratch my comment above.  `menuRole` could be undefined here.

Should do something like this... 

```
return this.context.menuRole || this.props.['aria-haspopup'];
```

andyseracuse:
Okay!  I made all the suggested changes.  Thanks so much for all your suggestions and guidance @kyletsang

src/Dropdown.js
```diff
@@ -107,11 +109,21 @@ class Dropdown extends React.Component {
     return this._$menuCtrl;
   }
 
+  getItemType() {
+    if(this.context.menuRole === 'listbox') {
+      return 'option'
+    }
+    return 'menu'
```
kyletsang, line 116:
```suggestion
    return 'menuitem'
```

src/Dropdown.js
```diff
@@ -107,11 +109,21 @@ class Dropdown extends React.Component {
...
+
   getMenuItems() {
     // In a real menu with a child DropdownMenu, `this.getMenu()` should never
     // be null, but it is sometimes null in tests. To mitigate that, we just
     // use `this.getContainer()` as the fallback `menuContainer`.
     const menuContainer = this.getMenu() || this.getContainer();
+    if(this.context.menuRole) {
+      return [].slice.call(menuContainer.querySelectorAll('[role="' + this.getItemType() + '"]'));
+    }
     return [].slice.call(menuContainer.querySelectorAll('[role="menuitem"]'));
```
kyletsang, line 127:
```suggestion
    return [].slice.call(menuContainer.querySelectorAll(`[role="${this.getItemType()}"]`));
```

getItemType will return the default menuitem in case when menuRole doesn't exist, so we can simplify this.  Also, string interpolation is good to use here too

src/DropdownItem.js
```diff
@@ -31,6 +31,16 @@ class DropdownItem extends React.Component {
...
   }
 
+  getRole() {
+    const menuRole = this.context.menuRole
+    if(menuRole === true || menuRole === 'menu' || menuRole === undefined) {
+      return 'menuitem'
+    }
+    if(menuRole === 'listbox') {
+      return 'option'
+    }
```
kyletsang, line 41: Can simplify this function similar to getRole in DropdownMenu

src/DropdownToggle.js
```diff
@@ -49,7 +49,12 @@ class DropdownToggle extends React.Component {
     this.context.toggle(e);
   }
 
+  getRole() {
+    return this.context.menuRole || this.props['aria-haspopup'];
+  }
+
   render() {
+    console.log(this.getRole())
```
kyletsang, line 57:
```suggestion
```

kyletsang (changes requested):
Almost there!  Few more nitpicks and we're good to go.

andyseracuse:
gotcha, I'm on it!  Thanks so much for all your help.  This has been super educational!!!

kyletsang (approved):

kyletsang:
Looks good, thanks!

PR #2083, opened 2021-01-21
feat(TabPane): Add renderOnlyIfActive prop

svekko:
- [x] New feature 

The reason for this PR is: 

I have a page with multiple tab panes. In each tab pane there is a separate table that requires some data from database. That means, if there are 4 tab panes then 4 separate requests are made to the server, although only 1 one of those tab panes is active. To prevent that I propose `renderOnlyIfActive` property for `TabPane` that allows me to prevent tab panes from rendering if not active.

### DIFF
--- a/src/TabPane.js
+++ b/src/TabPane.js
@@ -8,17 +8,20 @@ const propTypes = {
   tag: tagPropType,
   className: PropTypes.string,
   cssModule: PropTypes.object,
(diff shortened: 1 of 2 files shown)


kyletsang:
Thanks for the PR.  I believe the conditional rendering is best handled by the user though.

PR #2085, opened 2021-01-24
Update Dropdowns

gthomas-appfolio:
- Update Dropdown, DropdownMenu, ButtonDropdown with start/end vs left/right
- Update tests
- Update docs

TODO dropup docs seem to be broken


gthomas-appfolio:
Ignore, PR meant for a v5 fork which will PR back to here when complete

kyletsang:
Probably worth mentioning that reactstrap will very likely be deprecated in favor of react-bootstrap when v5 lands.  The v5 work there is pretty far along now and I think it will be better to unify the community.

Issue #1748, opened 2020-01-10
bootstrap 5 upgrade plan?


thecodejack:
Now that bootstrap 5 nearing the release(https://github.com/twbs/bootstrap/projects/11), any plans to start working on supporting same?

TheSharpieOne:
I haven't been tracking it. Personally, I haven't had the time to give to reactstrap for a while now. Plus (FWIW) react-bootstrap now supports v4. I would love for some help from the community and would be happy to pass the torch if someone wants to completely take over.

thecodejack:
At present, I can try starting branch around bootstrap 5. I would love to takeover the torch but don't think I am ready as of now. I will start working on bootstrap 5 upgrade.

Ashniu123:
I would also love to assist with the work on bootstrap 5.

TheSharpieOne:
I made a branch `bootstrap5`, make PRs to that. Feel free to open new issues to keep track of bootstrap5 related efforts so that efforts are not duplicated.

thecodejack:
Thanks. That really helps. Waiting for a dev release tag of bootstrap to start on this. I will review after and add all todos once available. Later can start picking the tasks.

iamandrewluca:
@TheSharpieOne 

> I would love for some help from the community and would be happy to pass the torch if someone wants to completely take over.

I can step in to help with the library maintainance.

> Plus (FWIW) react-bootstrap now supports v4.

Maybe slowly move `reactstrap` API to match `react-bootstrap` and deprecate this library in favor of `react-bootstrap`?

As I see they have a roadmap for bootstrap v5

mxschmitt:
Hey folks,
I'm one of the react-bootstrap maintainers, we would really appreciate to work together for Bootstrap 5. We have updated our codebase fully too hooks, have tests in place and soon (PR open) we are moving it to TypeScript for a better type safety.
What do you think about that? (if not, thats also fully okay. Would be cool to provide for the community one solution which is in place for using Bootstrap with React)

kyletsang:
> 
> 
> Hey folks,
> I'm one of the react-bootstrap maintainers, we would really appreciate to work together for Bootstrap 5. We have updated our codebase fully too hooks, have tests in place and soon (PR open) we are moving it to TypeScript for a better type safety.
> What do you think about that? (if not, thats also fully okay. Would be cool to provide for the community one solution which is in place for using Bootstrap with React)
> 
[...]

PR #2105, opened 2021-02-18
Feat(bs5) accordion

phwebi:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

This PR adds the new Accordion feature introduced in Bootstrap 5 and updates the minimum react version to 16.8.0 in order to support hooks.


src/Accordion.js
```diff
@@ -0,0 +1,53 @@
...
+  } = props;
+  const classes = mapToCssModules(classNames(
+    className,
+    'accordion',
+  ), cssModule);
+
+  const [openId, setOpenId] = useState();
+
+  return (
+    <AccordionContext.Provider value={{
```
kyletsang, line 39: Let's `useMemo` the value for this context

src/AccordionBody.js
```diff
@@ -0,0 +1,57 @@
...
+    className,
+    cssModule,
+    tag: Tag,
+    innerRef,
+    children,
+    id,
+    ...attributes
+  } = props;
+
+  return (
```
kyletsang, line 36: Let's use `useContext` hook instead of `***.Consumer`

docs/lib/routes.js
```diff
@@ -42,7 +43,8 @@ const routes = (
     <Route path="/" component={UI.Layout}>
       <IndexRoute component={Home} />
       <Route path="/components/" component={Components}>
-        <IndexRedirect to="alerts/" />
```
kyletsang, line 45: Let's add the alerts page back :)

src/AccordionBody.js
```diff
@@ -0,0 +1,57 @@
...
+    <AccordionContext.Consumer>
+      {({ openId }) => {
+        const classes = mapToCssModules(classNames(
+          className,
+          'accordion-collapse',
+        ), cssModule);
+
+        return (
+          <Collapse id={id} tag={Tag} {...attributes} className={classes} ref={innerRef} isOpen={openId === id}>
+            <Tag className="accordion-body">{children}</Tag>
```
kyletsang, line 46: Probably best to make it so `tag` doesn't control both the collapse and accordion-body elements.  Just for more flexibility if needed

src/Accordion.js
```diff
@@ -0,0 +1,53 @@
...
+    tag: Tag,
+    innerRef,
+    ...attributes
+  } = props;
+  const classes = mapToCssModules(classNames(
+    className,
+    'accordion',
+  ), cssModule);
+
+  const [openId, setOpenId] = useState();
```
kyletsang, line 36: I think we should make Accordion a controlled component and add an uncontrolled version similar to the others in the lib

kyletsang (changes requested):

src/AccordionBody.js
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <AccordionContext.Consumer>
+      {({ openId }) => {
+        const classes = mapToCssModules(classNames(
+          className,
+          'accordion-collapse',
+        ), cssModule);
+
+        return (
+          <Collapse id={id} tag={Tag} {...attributes} className={classes} ref={innerRef} isOpen={openId === id}>
```
kyletsang, line 45: Had another thought about this.  Probably best to not tie the key of the accordion body to an html id.  Just to provide more flexibility like if there are multiple accordions on the page.

docs/lib/routes.js
phwebi: I think accordion should be the new redirect, no? If everything is alphabetic 🤔

kyletsang: Ah I mean adding the alerts page back since it was deleted entirely.  The accordion page can stay as the main redirect, but there's no more alerts page on this routes list.

src/Accordion.js
phwebi: Curious why we should do this - the value doesn't really require expensive computation to come up with 🤔 I'd think that adding a useMemo hook here would make things more complex without much benefit.

docs/lib/routes.js
kyletsang: Actually disregard.  Didn't realize alerts was declared below

phwebi: OH gotcha - didn't realize this `IndexRedirect` was the only link to the alerts page. 👍  Will do

phwebi: Actually - isn't it here? https://github.com/reactstrap/reactstrap/pull/2105/files#diff-68a33def9ed9e0ff643d9b783aac860a707fd6f5652fffe863dd5123bdac502dR69

Is there something else I'm missing? 😅

src/Accordion.js
kyletsang: It's to avoid re-rendering of the context consumers in the case the accordion needs to re-render and the context object values don't change.  Without `useMemo`, the object reference changes each time, so the `AccordionContext.Consumer`s will re-render unnecessarily.

docs/lib/routes.js
kyletsang: Yeah sorry my bad.  Ignore my comment above - didn't notice alerts page was already there

src/Accordion.js
phwebi: Ahh sorry - was assuming we were going to convert to `useContext` per the other comment. Re-rendering shouldn't be a concern if we convert to the hook, right?

kyletsang: We would use a combination of `useMemo` (on the provider side) and `useContext` (on the consumer side).  `useContext` mainly because you can avoid the `Context.Consumer` declaration and in general it's a much cleaner look :)

For example you can reduce something like
```
return (
  <AccordionContext.Consumer>
    {{ openId } => {
      return ... 
    }
  <AccordionContext.Consumer>
);
```

to

```
const { openId } = useContext(AccordionContext);

return ...
```

phwebi:
Thanks for reviewing btw!

docs/lib/Components/AccordionPage.js
```diff
@@ -0,0 +1,149 @@
...
+  tag: tagPropType,
+  className: PropTypes.string,
+  cssModule: PropTypes.object,
+  innerRef: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.string,
+    PropTypes.func,
+  ]),
+  children: PropTypes.node,
+  id: PropTypes.string.isRequired,
```
kyletsang, line 116: Change to `accordionId`

types/lib/AccordionBody.d.ts
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import { CSSModule } from './index';
+
+export interface AccordionBodyProps extends React.HTMLAttributes<HTMLElement> {
+  tag?: React.ElementType;
+  cssModule?: CSSModule;
+  innerRef?: React.Ref<HTMLElement>;
+  id: string;
```
kyletsang, line 8: Change to `accordionId`

kyletsang (changes requested):

kyletsang:
> 
> 
> Thanks for reviewing btw! (I force pushed since I had to rebase to get the updated bootstrap version for testing my changes)

No problem!   I think a couple more changes and it should be good to go.

kyletsang (approved):

singhmadhukar11:
Not able to export Accordion

phwebi:
Hi @singhmadhukar11 ! The Accordion is part of the bootstrap 5 branch that hasn't been officially released yet. We do have a pre-release version `9.0.0-0` if you want to test it out.

singhmadhukar11:
> Hi @singhmadhukar11 ! The Accordion is part of the bootstrap 5 branch that hasn't been officially released yet. We do have a pre-release version `9.0.0-0` if you want to test it out.

Hey @phwebi thanks for information. If this is the part of bootstrap 5 then why i able to see on website yet if you not  released 9.0.0-0 version.
Because of this develop will be confused.

phwebi:
Oh - we had a hiccup with the release bot tagging the prerelease as our latest version. We fixed that, but didn't realize the website was updated. We need to revert that as well. Thanks for the heads up!

PR #2107, opened 2021-02-18
Feat(input group) remove input group addon

phwebi:
- [ ] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [x] Breaking change 

This PR removes the `input-group-addon` class which is no longer supported in Bootstrap 5. It also removes the `InputGroupButtonDropdown` component since it's no longer relevant with the Bootstrap 5 changes. `InputGroupButtonDropdown` can now be replaced with a normal `Dropdown`.


src/Dropdown.js
```diff
@@ -141,8 +139,9 @@ class Dropdown extends React.Component {
     const container = this.getContainer();
     const menu = this.getMenu();
     const clickIsInContainer = container.contains(e.target) && container !== e.target;
+    const clickIsInInput = container.classList.contains('input-group') && container.classList.contains('dropdown') && e.target.tagName === 'INPUT';
     const clickIsInMenu = menu && menu.contains(e.target) && menu !== e.target;
-    if ((clickIsInContainer || clickIsInMenu) && (e.type !== 'keyup' || e.which === keyCodes.tab)) {
+    if (((clickIsInContainer && !clickIsInInput) || clickIsInMenu) && (e.type !== 'keyup' || e.which === keyCodes.tab)) {
```
phwebi, line 144: The removal of the `input-group-addon` class/component interfered with the ability for a a dropdown attached to an input group to automatically close when the input is clicked. This is because we no longer have the input group addon as a container. Adding a placeholder container interferes with the way Bootstrap 5 calculates button radius for buttons at the start/end of the input group, so this is a workaround that addon buttons are still rendered correctly while behavior of dropdowns within in input group remain correct.

PR #2119, opened 2021-02-18
feat(bs5): change no-gutters class to gx-0

phwebi:
- [ ] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

This PR updates `no-gutters` class to `gx-0` per Bootstrap 5 changes.

### DIFF
--- a/src/Row.js
+++ b/src/Row.js
@@ -1,14 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { mapToCssModules, tagPropType } from './utils';
+import { mapToCssModules, tagPropType, deprecated } from './utils';
 
 const rowColWidths = ['xs', 'sm', 'md', 'lg', 'xl'];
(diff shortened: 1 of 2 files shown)


src/Row.js
```diff
@@ -52,7 +52,7 @@ const Row = (props) => {
 
   const classes = mapToCssModules(classNames(
     className,
-    noGutters ? 'no-gutters' : null,
+    noGutters ? 'gx-0' : null,
```
kyletsang, line 55: I wonder if we should convert this to a convenience prop for all gutter sizes.  Alternatively could remove the prop entirely and leave class handling in user land.  Thoughts?

phwebi: I think it's nice to avoid breaking changes if possible. Perhaps we can keep the prop for now, but add a deprecation warning? Any more fine tined adjustments to the gutter can still be achieved through normal classes.

kyletsang: Sure let's add the warning.  Thanks!

phwebi: 👍

kyletsang (approved):

PR #2179, opened 2021-05-11
Adding MDBReact To Themes Premium List

bwsky-a:
- [ ] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/docs/lib/PremiumThemes/premium.js
+++ b/docs/lib/PremiumThemes/premium.js
@@ -130,4 +130,21 @@ export default [
       },
     ],
   },
+  {
+    creator: {
+      name: 'MDBootstrap',
+      tagLine:
+        'Check out some examples that our partners from MDBootstrap created using the Reactstrap components library.',
+      link: 'https://mdbootstrap.com/',
+    },
+    products: [
+      {
+        title: 'Material Design for Bootstrap 5 & React 17',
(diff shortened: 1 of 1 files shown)


bwsky-a:
@phwebi Any feedback on this PR or possibly when would it be merged?

phwebi (approved):

bwsky-a:
@phwebi Do you know why I can't see my changes made on the website? They should probably be visible here - https://reactstrap.github.io/premium-themes/, but unfortunately, I can't see anything

bwsky-a:
@phwebi Do you know why I can't see my changes made on the website?

bwsky-a:
@phwebi any updates?

TheSharpieOne:
The docs only get updated when a new version is published. This is to keep the docs in "lock-step" with the latest release, which is needed to ensure what you see in the docs can be used. Previously we would publish the docs with every change (which would reflect `master`), but we would get feedback that `x` feature found in the docs doesn't work since that feature was not released (and was only in `master`).

Ideally, extra pages, such as premium themes, would be able to be published independently. That and/or we could have all of the docs for multiple, specific versions published at the same time (similar to how you can see older versions of bootstrap on bootstrap's docs site).

TheSharpieOne:
@bwsky-a I just manually updated the docs premium themes

PR #2236, opened 2021-08-21
Add horizontal collapse

bestguy:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -6,6 +6,7 @@ import { mapToCssModules, omit, pick, TransitionTimeouts, TransitionPropTypeKeys
 
 const propTypes = {
   ...Transition.propTypes,
+  horizontal: PropTypes.bool,
   isOpen: PropTypes.bool,
   children: PropTypes.oneOfType([
     PropTypes.arrayOf(PropTypes.node),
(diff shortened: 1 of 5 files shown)


src/Collapse.js
```diff
@@ -117,10 +120,11 @@ class Collapse extends Component {
           let collapseClass = getTransitionClass(status);
           const classes = mapToCssModules(classNames(
             className,
+            horizontal && 'collapse-horizontal',
             collapseClass,
             navbar && 'navbar-collapse'
           ), cssModule);
-          const style = height === null ? null : { height };
+          const style = dimension === null ? null : { [horizontal ? 'width' : 'height']: dimension };
```
phwebi, line 127: Can we add a test that the actual style is set correctly depending on whether the `horizontal` prop is true or false?

phwebi (commented):
Nice!

src/Collapse.js
bestguy: Good idea, however this is not easy to test as the style attribute is removed after transition is over - the height and width are only present during the transition.

phwebi: Gotcha - that's fine then!

phwebi (approved):

PR #2294, opened 2021-09-14
feat(Navbar): add container options

phwebi:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -12,14 +12,17 @@ const propTypes = {
   color: PropTypes.string,
   role: PropTypes.string,
   tag: tagPropType,
+  container: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
   className: PropTypes.string,
   cssModule: PropTypes.object,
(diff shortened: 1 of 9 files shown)


src/Navbar.js
```diff
@@ -60,7 +62,11 @@ const Navbar = (props) => {
   ), cssModule);
 
   return (
-    <Tag {...attributes} className={classes} />
+    <Tag {...attributes} className={classes}>
+      <div className="container-fluid">
```
bestguy, line 66: I wonder if this should be a prop that handles this vs hardcoding?  E.g. like: https://github.com/bestguy/sveltestrap/blob/master/src/Navbar.d.ts#L7

bestguy (commented):
One question

src/Navbar.js
phwebi: 👍🏼  that looks good. And we can default to fluid

bestguy (approved):

PR #2416, opened 2021-12-10
fix(FormGroup): dont pass switch prop to tag #2415

### LINKED ISSUE #2415: FormGroup: Using switch prop results in warning: Received `true` for a non-boolean attribute `switch`
- components: `name`
- reactstrap version `#9.0.1`
- import method `typescript, import { FormGroup } from 'reactstrap' - whichever that is :)`
- react version `#17.0.2`
- bootstrap version `#5.1.3`

### What is happening?

When using the `switch` prop with `FormGroup`:

```
<FormGroup switch>
...
</FormGroup>
```

I get this warning in the browser:

```
Warning: Received `true` for a non-boolean attribute `switch`.
      
      If you want to write it to the DOM, pass a string instead: switch="true" or switch={value.toString()}.
          at div
          at FormGroup (C:\dev\Portal\src\Coeo.Portal\clientapp\node_modules\reactstrap\src\FormGroup.js:25:5)
         ...
```

### What should be happening?

No warnings should be displayed.

### Steps to reproduce issue

1. Create a `FormGroup` and set the `switch` prop.
2. Run the app

### Error message in console

See above.

### Code

https://reactstrap-py5tda.stackblitz.io

(Check the console pane to see the warning)

mikegoatly:
- [X] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [X] There is an open issue which this change addresses
- [X] My change requires a change to [Typescript typings](./types/lib).
  - [X] I have updated the typings accordingly.

Fixes #2415 - I also noticed that the typings didn't include the `switch` prop, so I've added it for completeness.

### DIFF
--- a/src/FormGroup.js
+++ b/src/FormGroup.js
@@ -30,16 +30,17 @@ const FormGroup = (props) => {
     inline,
     floating,
     tag: Tag,
+    switch: switchProp,
     ...attributes
   } = props;
 
-  const formCheck = check || props.switch;
+  const formCheck = check || switchProp;
 
   const classes = mapToCssModules(classNames(
     className,
     row ? 'row' : false,
     formCheck ? 'form-check' : 'mb-3',
-    props.switch ? 'form-switch' : false,
+    switchProp ? 'form-switch' : false,
     formCheck && inline ? 'form-check-inline' : false,
     formCheck && disabled ? 'disabled' : false,
     floating && 'form-floating'
--- a/types/lib/FormGroup.d.ts
+++ b/types/lib/FormGroup.d.ts
@@ -5,6 +5,7 @@ export interface FormGroupProps extends React.HTMLAttributes<HTMLDivElement> {
   [key: string]: any;
   row?: boolean;
   check?: boolean;
+  switch?: boolean;
   inline?: boolean;
   floating?: boolean;
   disabled?: boolean;


src/FormGroup.js
```diff
@@ -30,16 +30,17 @@ const FormGroup = (props) => {
     inline,
     floating,
     tag: Tag,
+    switch: switchProp,
```
davidacevedo, line 33: Can you leave this as `switch` instead of `switchProp`?

mikegoatly: Unfortunately not - doing so means you end up with ambiguity between the `switch` variable and JavaScript  `switch`/`case` statement - I can call it something else if you'd prefer, e.g. `_switch`?

jfhenon: Hello, It would be nice to see this ugly error message fixed.
Maybe a comment to explain the variable name would be good to approve this merge? 
The other way is to keep the props.switch notation?
Thanks

mikegoatly: We can't leave it as `props.switch` - destructuring the prop like this fixes the issue - without it `switch` remains in the `attributes` object that is passed through to the underlying control.

As I say, I'm very happy to use whatever name is deemed appropriate, but as far as I know this PR hasn't even been looked at by any of the maintainers, so I don't think that's what's causing it to be left unmerged.

jfhenon: workaround: you can remove the switch prop from FormGroup and replace with the bootstrap classname:
` <FormGroup className="form-check form-switch">`

davidacevedo: Had forgotten to respond, the PR looks good 👍

davidacevedo (approved):

Issue #2431, opened 2022-01-07
<FormGroup> renders as <div class="mb-3">


joshuajameshunt:
- components: `FormGroup`
- reactstrap version `9.0.1`
- import method `scss`
- react version `17.0.2`
- bootstrap version `5.1.3`

### What is happening?

<FormGroup> is rendering as <div class="mb-3">

### What should be happening?

<FormGroup> should be rendering as <div class="form-group">

### Steps to reproduce issue

1. Add a <FormGroup> component
2. Look at the render output

### Error message in console

```
N/A
```

### Code

<FormGroup></FormGroup>

scivarolo:
According to [Bootstrap 5's migration docs](https://getbootstrap.com/docs/5.1/migration/#forms), form-specific layout classes (`.form-group`, `.form-row`, `.form-inline`) were dropped in favor of using the grid system

joshuajameshunt:
@scivarolo you are correct, it just seems like adding `.form-group { margin-bottom: 1rem; }` to your global CSS is a lot more convenient than having to remove every single non-check FormGroup from your project (or replace all references with your own FormGroup component) because you styled it differently (a bottom margin of only 0.75rem, for example).

juandelacruz23:
It's also a bad idea for me that the `mb-3` is added after classnames, making it impossible to remove that margin with utility classes nor with a personalized bootstrap variables file.

davidacevedo:
`form-group` won't be added back in as it doesn't exist anymore in the bootstrap context, but what about having a prop that disables the default margin class?

shtbik:
In this case, wouldn't be a bad idea to update the docs:
https://reactstrap.github.io/?path=/docs/components-forms--input#form-inline

AJax2012:
> According to [Bootstrap 5's migration docs](https://getbootstrap.com/docs/5.1/migration/#forms), form-specific layout classes (`.form-group`, `.form-row`, `.form-inline`) were dropped in favor of using the grid system

Is this also why the `inline` property doesn't seem to work in the LabelProps or the FormProperties interfaces anymore as well? If so, would you mind if I submit a PR to remove that property from both? I don't know if your team allows outside parties to submit PRs, so I figured I'd ask, but as @shtbik pointed out, it doesn't see to work in the storybook demo either.

illiteratewriter:
@AJax2012 Feel free to make a PR.

AJax2012:
[...]

PR #2432, opened 2022-01-07
<FormGroup> render with class "form-group" instead of "mb-3" when not "form-check"

joshuajameshunt:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

https://github.com/reactstrap/reactstrap/issues/2431


davidacevedo:
If you need `form-group` you can add it by passing in the `className` prop. Bootstrap 5 has removed form-group, so it doesn't make sense for reactstrap to include this anymore. https://getbootstrap.com/docs/5.0/migration/#forms

Issue #2425, opened 2021-12-23
Badge text color not applied for light color backgrounds


lilasquared:
- components: `Badge`
- reactstrap version `#9.0.1`
- react version `#17.0.0`
- bootstrap version `#5.1.3`

### What is happening?

for `<Badge color="warning">Some Text</Badge>` the text-color is `white` when it probably should be `dark`.  This is a change from when there were `badge-*` classes, where now the badges use `bg-*` utility classes which only set the background. [source](https://getbootstrap.com/docs/5.1/components/badge/#background-colors)

### What should be happening?

I would expect reactstrap to take care of the text coloring too - though I don't know the best way to do it given you can customize the `theme-colors` in bootstrap.

### Code
```jsx
<Badge color="warning">Some Text</Badge>
```

BenJenkinson:
Bootstrap 5.2 has added some helper classes for this scenario

> Set a background-color with contrasting foreground color with [our `.text-bg-{color}` helpers](https://getbootstrap.com/docs/5.2/helpers/color-background/). Previously it was required to manually pair your choice of [`.text-{color}`](https://getbootstrap.com/docs/5.2/utilities/colors/) and [`.bg-{color}`](https://getbootstrap.com/docs/5.2/utilities/background/) utilities for styling, which you still may use if you prefer.
> 
> 📚 https://getbootstrap.com/docs/5.2/components/badge/#background-colors

PR #2443, opened 2022-01-25
badge change text color

msobalvarro:
- [x] Bug fix 
- [X] There is an open issue which this change addresses. Fixed issue [Issue #2425](https://github.com/reactstrap/reactstrap/issues/2425#issue-1087822368)

### DIFF
--- a/src/Badge.js
+++ b/src/Badge.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { mapToCssModules, tagPropType } from './utils';
+import { mapToCssModules, tagPropType,  } from './utils';
 
 const propTypes = {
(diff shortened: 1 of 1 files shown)


src/Badge.js
```diff
@@ -34,6 +34,7 @@ const Badge = (props) => {
     className,
     'badge',
     'bg-' + color,
+    color === "light" && "text-secondary",
```
davidacevedo, line 37: We should use `text-dark` instead to be more inline with the bootstrap documentation https://getbootstrap.com/docs/5.1/components/badge/#background-colors

davidacevedo:
So, this is a good next step, but we should also include `[warning, info]` as those other colors also require `text-dark` https://getbootstrap.com/docs/5.1/components/badge/#background-colors

PR #2483, opened 2022-03-31
convert button to use hooks

illiteratewriter:
- [ ] Bug fix 
- [ ] New feature 
- [X] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { mapToCssModules, tagPropType } from './utils';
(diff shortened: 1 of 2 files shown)


davidacevedo (dismissed):
Looks good 👍

TheSharpieOne:
Would this be considered a breaking change since it would break anyone that might be trying to get the instance of the component via `ref` (not the `innerRef`)

illiteratewriter:
The Button component was using `innerRef` previously also.

davidacevedo (approved):
👍

davidacevedo:
> Would this be considered a breaking change since it would break anyone that might be trying to get the instance of the component via `ref` (not the `innerRef`)

@TheSharpieOne I'm curious how many people grab and use the instance of the component. It's not something that is documented here, so we should not let that influence whether we should convert or not. I also don't think it's something we should be supporting.

Also, I would assume the button component will most likely not be affected, as someone can just roll their own button if they wanted to call the onClick function from another button!

PR #2484, opened 2022-04-01
convert button toggle to use hooks

illiteratewriter:
- [X] Chore 
- [ ] Breaking change 


src/__tests__/ButtonToggle.spec.js
```diff
@@ -10,12 +10,6 @@ describe("ButtonToggle", () => {
     expect(wrapper.first(Button).shallow().text()).toBe("Ello world");
   });
 
-  it('should have button already toggled for defaultValue true', () => {
```
davidacevedo, line 13: Why was this removed?

src/ButtonToggle.js
```diff
@@ -15,72 +15,52 @@ const defaultProps = {
   defaultValue: false,
```
davidacevedo, line 15: To avoid checking for `onClick`, `onBlur`, and `onFocus` you can add defaultProps to the component:
```
const noop = () => {}
const defaultProps = {
   ...,
   onBlur: noop,
   onFocus: noop,
   ...
}

Comment (davidacevedo) [reply src/ButtonToggle.js:15]: Although for 1:1 translation this works as the old implementation did this.

File: src/__tests__/ButtonToggle.spec.js (hunk 1 above)
Comment (illiteratewriter) [reply src/__tests__/ButtonToggle.spec.js:13]: This was removed since the state can be accessed only inside class components. Should we add `react-testing-library` to solve this, or are there any other suggestions?

Comment (davidacevedo) [reply src/__tests__/ButtonToggle.spec.js:13]: You can do this with enzyme by:
```
wrapper.find(Button).props().active
```
and run expectations on that.

Although I do recommend going with react-testing-library instead as enzyme is technically a dead project now.

davidacevedo (approved):

PR #2536, opened 2022-05-29
feat(CloseButton): add close button

illiteratewriter:
- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/CloseButton.js
+++ b/src/CloseButton.js
@@ -1,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { mapToCssModules } from './utils';
+
+const propTypes = {
+  active: PropTypes.bool,
+  'aria-label': PropTypes.string,
+  onClick: PropTypes.func,
+  variant: PropTypes.oneOf(['white', 'black'])
+}
+
+const defaultProps = {
+  variant: 'black',
+  'aria-label': 'close'
+}
+
+const CloseButton = (props) => {
+  const {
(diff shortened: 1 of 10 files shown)


src/CloseButton.js
```diff
@@ -0,0 +1,45 @@
...
+
+const propTypes = {
+  active: PropTypes.bool,
+  'aria-label': PropTypes.string,
+  onClick: PropTypes.func,
+  variant: PropTypes.oneOf(['white', 'black'])
+}
+
+const defaultProps = {
+  variant: 'black',
```
davidacevedo, line 14: I don't think bootstrap has the `black` variant. This is just the `btn-close`. The only variant we should support is `white`.

Btw, `Button` has a `close` prop right now. We might want to use this `Close` button if that prop is passed.

stories/CloseButton.stories.js
```diff
@@ -0,0 +1,20 @@
...
+  
+  A generic close button for dismissing content like modals and alerts.
+        `,
+      }
+    }
+  }
+};
+
+export { default as CloseButton } from './examples/CloseButton';
+export { default as CloseButtonWhite } from './examples/CloseButtonWhite';
```
davidacevedo, line 19: Can we add a disabled state example. It should follow the bootstrap documentation if possible https://getbootstrap.com/docs/5.2/components/close-button/

davidacevedo (approved):
Looks good 👍

Issue #2410, opened 2021-12-05
ButtonDropdown, Dropdown in InputGroup has a broken rounded corners


janicekt:
- components: ` ButtonDropdown, Dropdown in InputGroup`
- reactstrap version `#9.0`
- react version `any`
- bootstrap version `#5.1`

### What is happening?
<img width="452" alt="Snímek obrazovky 2021-12-05 v 12 47 27" src="https://user-images.githubusercontent.com/25146453/144745228-e45178b5-27b9-4e36-a004-85e5080f4d49.png">

if there is a ButtonDropdown or Dropdown in the InputGroup, then the wrapper div of the button breaks the border round look (div class="btn-group" or div class="dropdown")..

### What should be happening?

if the parent is an InputGroup, then the dropdown button should not be wrapped in another div. See official documentation https://getbootstrap.com/docs/5.1/forms/input-group/#buttons-with-dropdowns

### Steps to reproduce issue

can also be seen in the yours storybook documentation https://reactstrap.github.io/?path=/docs/components-inputgroup--input-group-button#input-group-button

PR #2540, opened 2022-06-04
fix(#2410): broken rounded corners for dropdown in input group

illiteratewriter:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 


illiteratewriter:
The issue was the selector for dropdown in the input group changed for the new version of Bootstrap. The new selector selected the immediate child of InputGroup, whereas the previous version had a div between InputGroup and the Button.

I removed the extra `div` but then all of the remaining code was dependent on selecting via that particular `div`. So have made changes to fix them.

davidacevedo:
I like the format you were doing before for commits, `fix(#1410): your commit message here`

src/Dropdown.js
```diff
@@ -101,14 +110,18 @@ class Dropdown extends React.Component {
     return this.menuRef.current;
   }
 
+  getToggle() {
+    return this.toggleRef.current;
+  }
```
davidacevedo, line 115: I don't think we need this method as on line `119` we are not using this.

illiteratewriter: We are using that on line `154`, so added to `119` as well.

davidacevedo:
Did this PR get a bit broken? It seems to have a lot more changes now?

illiteratewriter:
I was wondering the same, I actually merged `master` in local and pushed it but it is still showing merge conflicts also.

davidacevedo (approved):
Looks good, took me a bit to figure out what was going on, so sorry for the delay!

PR #2542, opened 2022-06-07
fix: add a flag to DropdownMenu for Popper update

aaronmars:
Add a boolean prop to DropdownMenu to update the Popper position every time an item is selected (see `update` documentation here: https://popper.js.org/react-popper/v2/render-props/#children).

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/DropdownMenu.js
+++ b/src/DropdownMenu.js
@@ -18,6 +18,7 @@ const propTypes = {
   persist: PropTypes.bool,
   strategy: PropTypes.string,
   container: targetPropType,
+  updateOnSelect: PropTypes.bool,
   right: deprecated(PropTypes.bool, 'Please use "end" instead.'),
 };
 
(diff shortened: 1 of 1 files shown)


src/DropdownMenu.js
```diff
@@ -18,13 +18,15 @@ const propTypes = {
...
   container: targetPropType,
+  updateOnSelect: PropTypes.bool,
   right: deprecated(PropTypes.bool, 'Please use "end" instead.'),
 };
 
 const defaultProps = {
   tag: 'div',
   flip: true,
   modifiers: [],
+  updateOnSelect: false,
```
davidacevedo, line 29: Do we need false if `undefined` is also falsey?

aaronmars: Fair enough. I was very concerned about making it obvious this is opt-in. You are correct that it's unnecessary.

davidacevedo (approved):

PR #2544, opened 2022-06-08
docs(Spinner): update Spinner docs to follow Bootstrap documentaion

illiteratewriter:
Add new examples
Add description to props

- [ ] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/Spinner.js
+++ b/src/Spinner.js
@@ -4,12 +4,33 @@ import classNames from 'classnames';
 import { mapToCssModules, tagPropType } from './utils';
 
 const propTypes = {
+  /**
+   * Set a custom element for this component
+   */
   tag: tagPropType,
-  type: PropTypes.string,
+  /**
+   * Change animation of spinner
+   */
+  type: PropTypes.oneOf(['border', 'grow']),
+  /**
+   * Change size of spinner
+   */
   size: PropTypes.string,
-  color: PropTypes.string,
+  /**
+   * Change color of spinner
+   */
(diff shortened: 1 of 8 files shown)


davidacevedo:
I'll be adding some sort of commitlint for this, but can you reference the issue #2545 somewhere in the commit message. I've said to do feat(#num), but it might be better to add it to the message itself as that will provide a link in the changelog. 🤷 This is not too important of course.

src/Spinner.js
```diff
@@ -4,12 +4,33 @@ import classNames from 'classnames';
...
   tag: tagPropType,
-  type: PropTypes.string,
+  /**
+   * Change animation of spinner
+   */
+  type: PropTypes.oneOf(['border', 'grow']),
+  /**
+   * Change size of spinner
+   */
   size: PropTypes.string,
```
davidacevedo, line 18: I think we should note that spinner only supports `sm` as size.

src/Spinner.js
```diff
@@ -4,12 +4,33 @@ import classNames from 'classnames';
 import { mapToCssModules, tagPropType } from './utils';
 
 const propTypes = {
+  /**
+   * Set a custom element for this component
+   */
```
davidacevedo, line 9: Since these are not adding any other jsdoc params, you can just type it as:
```suggestion
  /** Set a custom element for this component */
```

davidacevedo (approved):

PR #2571, opened 2022-07-07
Add prettier

illiteratewriter:
- [ ] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 


aaronmars (commented):
This looks fine to me.
I'm not sure `eslint-plugin-prettier` is really necessary. Without it, we would just need to add another package.json script to run prettier (and `prettier -w`) explicitly. 
It's mentioned in the prettier docs here: https://prettier.io/docs/en/integrating-with-linters.html#notes

illiteratewriter:
The issue when I ran prettier on its own was it was conflicting with ESLint. This caused circular issues like ESLint would do one thing and prettier would do the exact opposite.

aaronmars:
I see. I suppose that just means there are conflicting rules. `eslint-config-prettier` is supposed to disable the eslint rules for the formatting things that prettier handles, but there might be something more esoteric.
Do you have an example?

illiteratewriter:
@davidacevedo can we add `prettier` also to the checks that are run?

illiteratewriter:
@davidacevedo Cool, will raise a PR for that separately.

davidacevedo (approved):

PR #2575, opened 2022-07-17
fix: remove overflow value after closing modal

### LINKED ISSUE #2574: Modal - incorrect overflow value
- components: `Modal`
- reactstrap version `#9.1.2`
- bootstrap version `#5.1.3`

### Real life example

1. `FormModal` contains a form.
2. After submitting the form we want to close the `FormModal` and open the `ConfirmationModal`.
3. After closing the `ConfirmationModal`, the page is blocked (cannot be scrolled).

### What should be happening?

Overflow `hidden` should be removed from the `<body>` element.

### Steps to reproduce issue

1. Open `FormModal`.
4. Open `ConfirmationModal` and close `FormModal` at the same time (see `handleOnClick` function).
5. Close `ConfirmationModal`.
6. Try to scroll (the entire page is blocked).

### Code

[CodeSandbox example](https://codesandbox.io/s/reactstrap-modal-issue-2chlsg?file=/src/App.js)

```
import React, { useState } from "react";
import { Button, Modal, ModalBody, ModalFooter } from "reactstrap";
import "bootstrap/dist/css/bootstrap.min.css";

export default function App() {
  const [formModal, setFormModal] = useState(false);
  const [confirmationModal, setConfirmationModal] = useState(false);

  const handleOnClick = () => {
    setFormModal(false);
    setConfirmationModal(true);
  };

  return (
    <div>
      <Button onClick={() => setFormModal(true)}>Open form</Button>
      <Modal isOpen={formModal} toggle={() => setFormModal((s) => !s)}>
        <ModalBody>Form</ModalBody>
        <ModalFooter>
          <Button onClick={handleOnClick}>Open confirmation modal</Button>
        </ModalFooter>
      </Modal>
      <Modal
        isOpen={confirmationModal}
        toggle={() => setConfirmationModal((s) => !s)}
      >
        <ModalBody>Confirmation</ModalBody>
      </Modal>

      <p style={{ margin: "150px 0" }}>Some Text</p>
      <p style={{ margin: "150px 0" }}>Some Text</p>
      <p style={{ margin: "150px 0" }}>Some Text</p>
      <p style={{ margin: "150px 0" }}>Some Text</p>
      <p style={{ margin: "150px 0" }}>Some Text</p>
    </div>
  );
}
```

lukbak:
Closes #2574 

This is what Bootstrap does as well.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -116,7 +116,6 @@ class Modal extends React.Component {
 
     this._element = null;
     this._originalBodyPadding = null;
-    this._originalBodyOverflow = null;
     this.getFocusableChildren = this.getFocusableChildren.bind(this);
     this.handleBackdropClick = this.handleBackdropClick.bind(this);
     this.handleBackdropMouseDown = this.handleBackdropMouseDown.bind(this);
@@ -371,9 +370,6 @@ class Modal extends React.Component {
     }
 
     this._originalBodyPadding = getOriginalBodyPadding();
-    this._originalBodyOverflow = window.getComputedStyle(
-      document.body,
-    ).overflow;
     conditionallyUpdateScrollbar();
 
     if (Modal.openCount === 0) {
@@ -419,7 +415,7 @@ class Modal extends React.Component {
       document.body.className = document.body.className
         .replace(modalOpenClassNameRegex, ' ')
         .trim();
-      document.body.style.overflow = this._originalBodyOverflow;
+      document.body.style.removeProperty('overflow');
     }
     this.manageFocusAfterClose();
     Modal.openCount = Math.max(0, Modal.openCount - 1);


illiteratewriter:
@lukbak doing this would cause pages which had already set `overflow: hidden` to the body to not have that property once someone closes the modal. We need to set the previous state back once the modal is hidden.

lukbak:
@illiteratewriter I understand the previous idea, but I do not fully agree with it.
1. If someone adds overflow to the body element, inline, makes a **big mistake**.
2. The second argument is that this is exactly how Bootstrap does it.

Even if we assume that we want to store the previous overflow value in `_originalBodyOverflow`. How do we want to get it? 
In the case from #2575, it is always `hidden` there.

illiteratewriter:
I think [what bootstrap does](https://github.com/twbs/bootstrap/blob/ed448925daa4fb20344f94e35c37dded9e0af8af/js/src/modal.js#L113) is, it adds a class `.modal-open` to the body which has `overflow: hidden` and removes that class once the modal is hidden. But bootstrap.css file doesn't have `modal-open` class with the above property.

lukbak:
@illiteratewriter You are right, but also check lines [111](https://github.com/twbs/bootstrap/blob/ed448925daa4fb20344f94e35c37dded9e0af8af/js/src/modal.js#L111) and [250](https://github.com/twbs/bootstrap/blob/ed448925daa4fb20344f94e35c37dded9e0af8af/js/src/modal.js#L250)

Issue #2552, opened 2022-06-10
Using Reactstrap 9 from CDNs


afialapis:
### Issue description

- reactstrap version `>=9.0.1`
- import method `umd`
- react version `#16.13.1`
- bootstrap version `#5.1.3`

### What is happening?

I am trying to do a simple test of Reactstrap using bundled scripts from unpkg.com. Things work fine for Reactstrap 8 (where we had the `full.js` bundle); but they don't work for Reactstrap 9, where I am including every dependency.

What am I doing wrong?

### Error message in console

```
Uncaught TypeError: PropTypes__default.default is undefined
    <anonymous> utils.js:130
    <anonymous> reactstrap.umd.js:4
    <anonymous> reactstrap.umd.js:5
[utils.js:130:30](https://www.unpkg.com/reactstrap@9.0.1/src/utils.js)

Uncaught ReferenceError: Reactstrap is not defined
    <anonymous> file:////.../index.html line 127202 > injectedScript:3
    run https://unpkg.com/@babel/standalone/babel.js:127202
    check https://unpkg.com/@babel/standalone/babel.js:127255
```

### Code

```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8"/>
  <title>reactstrap 9</title>
  <link rel="stylesheet" type="text/css" href="https://www.unpkg.com/bootstrap@5.1.3/dist/css/bootstrap.css"/>

  <style>
    .App {
      font-family: sans-serif;
      text-align: center;
      max-width: 200px;
      margin: 2rem auto;
    }

  </style>
</head>
<body>
  <div id="content"></div>

  <script type="text/javascript" src="https://www.unpkg.com/react@16.13.1/umd/react.development.js"></script>
  <script type="text/javascript" src="https://www.unpkg.com/react-dom@16.13.1/umd/react-dom.development.js"></script>
  <script src="https://unpkg.com/prop-types@15.8.1/prop-types.js"></script>

  <!--For Reactstrap 8 things work well with the full bundle -->
  <!--
  <script type="text/javascript" src="https://www.unpkg.com/reactstrap@8.5.1/dist/reactstrap.full.js"></script>
  -->
  
  <!-- Reactstrap 9 complains about PropTypes -->
  

  <script type="text/javascript" src="https://www.unpkg.com/classnames@2.3.1/index.js"></script>  
  <script type="text/javascript" src="https://www.unpkg.com/react-popper@2.3.0/dist/index.umd.js"></script>
[...]

PR #2579, opened 2022-07-20
fix(#2552): unable to use from CDN

illiteratewriter:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "cover": "npm test -- --coverage",
     "start": "start-storybook -s ./static -p 8080",
     "build:docs": "build-storybook -s ./static -o ./build",
-    "build": "microbundle --generateTypes false --jsx React.createElement --compress false",
+    "build": "microbundle --generateTypes false --jsx React.createElement --compress false --globals react=React,react-dom=ReactDOM,prop-types=PropTypes,react-transition-group=ReactTransitionGroup,react-popper=ReactPopper,classnames=classNames",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "cover": "npm test -- --coverage",
     "start": "start-storybook -s ./static -p 8080",
     "build:docs": "build-storybook -s ./static -o ./build",
-    "build": "microbundle --generateTypes false --jsx React.createElement --compress false",
+    "build": "microbundle --generateTypes false --jsx React.createElement --compress false --globals react=React,react-dom=ReactDOM,prop-types=PropTypes,react-transition-group=ReactTransitionGroup,react-popper=ReactPopper,classnames=classNames",
```
davidacevedo, line 23: Would this affect importing the library without those globals being defined?

illiteratewriter: I didn't understand the question, but the UMD build was not working without these globals. If you remove `react=React` from the above, then it'll throw `React__default.default is undefined`. In [this issue](https://github.com/reactstrap/reactstrap/issues/2552), it throws error for `propTypes`.

davidacevedo:
I'll try testing this a bit more later to see if I can validate my assuptions. It does look good otherwise

illiteratewriter:
Wouldn't be needed as we are moving away from microbundle

PR #2583, opened 2022-07-27
Migrate to react testing library

illiteratewriter:
- [ ] Bug fix 
- [ ] New feature 
- [x] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/__tests__/Accordion.spec.js
+++ b/src/__tests__/Accordion.spec.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, fireEvent, waitFor } from '@testing-library/react';
+import { render, fireEvent, waitFor, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { Accordion, AccordionItem, AccordionHeader, AccordionBody } from '..';
 import { testForCustomClass, testForCustomTag } from '../testUtils';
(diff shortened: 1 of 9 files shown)


src/__tests__/Accordion.spec.js
```diff
@@ -1,37 +1,148 @@
...
+        Accordion Item 3
+      </AccordionHeader>
+      <AccordionBody accordionId="3" data-testid="body-3">
+        This is the body of third item.{' '}
+      </AccordionBody>
+    </AccordionItem>,
+  ];
+
+  it('should render a div with "accordion" class', () => {
+    const { getByTestId } = render(
```
aaronmars, line 45: Let's standardize on one way of getting these query functions (and other things). It seems like using `screen` from the import is the recommended way:
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

So:
```javascript
import { render, screen } from '@testing-library/react';

.....

const node = screen.getByTestId('test');
```

src/__tests__/Accordion.spec.js
```diff
@@ -1,37 +1,148 @@
...
+  it('should call toggle with clicked item id', () => {
+    const mockFn = jest.fn(() => {});
+    const { getByText } = render(
+      <Accordion open="1" toggle={mockFn} data-testid="accordion">
+        {accordionItems}
+      </Accordion>,
     );
 
-    expect(wrapper.find('.accordion-flush').length).toBe(1);
+    fireEvent.click(getByText(/accordion item 2/i));
```
aaronmars, line 102: It would probably be best to use the `user-event` companion library instead of the raw events directly.

aaronmars (changes requested):
Added a couple of comments to standardize on a couple of patterns before things get copy/pasted.

aaronmars (approved):

PR #2628, opened 2022-10-05
test(Table): migrate to testing library

iamyuu:
- [ ] Bug fix 
- [ ] New feature 
- [x] Chore 
- [ ] Breaking change 


iamyuu:
Since using the testing library, we can now test the hover state, should I add it?

And also by using the testing library, render text as children `table` isn't correct, I suggest to render a proper HTML table by adding `tr` and `td` should I add them? or should I add mock the implementation `console.error` to disabled the warning?

Screen Shot 2022-10-05 at 23 20 58

PR #2639, opened 2022-10-27
fix: ensure initial overflow status is stored when opening multiple modals

fabio-aiello-appfolio:
### Correct bug that prevents scrolling after closing modal

When modals are opened, scrolling is disabled on the main page's body by setting its overflow property to 'hidden'. In certain situations when multiple modals are active, scrolling is not restored after they are closed. This fix corrects the behavior and restores the overflow status that exists on the main page's body at the time before the first modal's opening.

- [x] Bug fix 
- [x] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -116,7 +116,6 @@ class Modal extends React.Component {
 
     this._element = null;
     this._originalBodyPadding = null;
-    this._originalBodyOverflow = null;
     this.getFocusableChildren = this.getFocusableChildren.bind(this);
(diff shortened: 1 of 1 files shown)


davidacevedo:
This solves the issue where there were two modals open and the 2nd modal previous body overflow was hidden due to first modal setting the body overflow to hidden. This caused a bug where a page could not scroll anymore if opening two modals or opening 1 right after another has closed.

davidacevedo (approved):
Looks good 👍

PR #2647, opened 2022-11-11
feat!: add forwardRef to all components that contain innerRef

davidacevedo:
This will make us follow React best practices and depend on libraries that are expecting ref to exist

- [ ] Bug fix 
- [x] New feature 
- [ ] Chore 
- [x] Breaking change 

Supporting `forwardRef` to allow usage of reactstrap with other libraries like `react-hook-form`. I left `innerRef` so current usage can be backwards compatible, but this will still be a breaking change as that's [common practice with React](https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers). This will also break any usage where someone might use the implementation of a component.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -2,9 +2,9 @@ name: Test Coverage
 
 on:
   push:
-    branches: [ master, bootstrap5 ]
+    branches: [ master, bootstrap5, v10 ]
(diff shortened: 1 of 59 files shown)


steventhan (approved):

davidacevedo:
resolves #2650

PR #2648, opened 2022-11-12
Migrate to react testing library

illiteratewriter:
- [ ] Bug fix 
- [ ] New feature 
- [x] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/__tests__/Input.spec.js
+++ b/src/__tests__/Input.spec.js
@@ -1,272 +1,246 @@
 import React from 'react';
 import { shallow, mount } from 'enzyme';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
 import { Input } from '..';
+import { testForDefaultTag } from '../testUtils';
 
 describe('Input', () => {
   it('should render with "input" tag when no type is provided', () => {
-    const wrapper = shallow(<Input />);
-
-    expect(wrapper.type()).toBe('input');
+    testForDefaultTag(Input, 'input')
   });
 
   it('should render with "type" tag when type is "select"', () => {
-    const wrapper = shallow(<Input type="select">Yo!</Input>);
-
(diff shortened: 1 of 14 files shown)


src/__tests__/Input.spec.js
```diff
@@ -1,272 +1,250 @@
...
-    expect(wrapper.type()).toBe('input');
+    testForDefaultTag(Input, 'input');
   });
 
   it('should render with "type" tag when type is "select"', () => {
-    const wrapper = shallow(<Input type="select">Yo!</Input>);
-
-    expect(wrapper.type()).toBe('select');
+    render(<Input type="select">Yo!</Input>);
+    expect(screen.getByText(/yo/i).tagName.toLowerCase()).toMatch('select');
```
davidacevedo, line 14: You can also check if `document.querySelector('select')` exists. Or more specifically you can do `container.querySelector` where `container` is from `const { container } = render(...)`

src/__tests__/ListGroup.spec.js
```diff
@@ -1,73 +1,59 @@
...
   });
 
   it('should render with "flush"', () => {
-    const wrapper = shallow(<ListGroup flush>Yo!</ListGroup>);
+    render(<ListGroup flush>Yo!</ListGroup>);
 
-    expect(wrapper.text()).toBe('Yo!');
-    expect(wrapper.hasClass('list-group')).toBe(true);
-    expect(wrapper.hasClass('list-group-flush')).toBe(true);
+    expect(screen.getByText('Yo!')).toHaveClass('list-group-flush');
```
davidacevedo, line 18: Maybe still include the assertion from `list-group` as well for all the variations. Just protection if someone does end up adding some sort of `if` statement to change that later.... but doesn't seem likely.

davidacevedo (approved):
Looks good 👍

PR #2649, opened 2022-11-13
Migrate to RTL (Dropdown, ToastHeader)

illiteratewriter:
- [ ] Bug fix 
- [ ] New feature 
- [x] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/__tests__/Dropdown.spec.js
+++ b/src/__tests__/Dropdown.spec.js
@@ -1,52 +1,67 @@
 import React from 'react';
 import { mount, shallow } from 'enzyme';
 import { Popper, Reference } from 'react-popper';
+import { createEvent, fireEvent, render, screen } from '@testing-library/react';
+import user from '@testing-library/user-event';
+import '@testing-library/jest-dom';
 import { Dropdown, DropdownToggle, DropdownMenu, DropdownItem } from '..';
 import { keyCodes } from '../utils';
+import { testForChildrenInComponent } from '../testUtils';
 
 describe('Dropdown', () => {
-  let isOpen;
-  let toggle;
-  let element;
-
   beforeEach(() => {
-    isOpen = false;
-    toggle = () => {
-      isOpen = !isOpen;
-    };
-    element = document.createElement('div');
-    document.body.appendChild(element);
+    jest.useFakeTimers();
   });
 
   afterEach(() => {
-    if (jest.isMockFunction(Dropdown.prototype.componentDidUpdate))
-      Dropdown.prototype.componentDidUpdate.mockRestore();
-    if (jest.isMockFunction(Dropdown.prototype.handleProps))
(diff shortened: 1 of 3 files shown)


src/__tests__/ToastHeader.spec.js
```diff
@@ -1,44 +1,30 @@
...
-    expect(wrapper.find('button.btn-close').length).toBe(1);
+    expect(screen.getByLabelText(/close/i)).toBeInTheDocument();
   });
 
   it('should render custom tag', () => {
-    const wrapper = shallow(<ToastHeader tag="p">Yo!</ToastHeader>).childAt(0);
-
-    expect(wrapper.text()).toBe('Yo!');
-    expect(wrapper.type()).toBe('p');
+    render(<ToastHeader tag="p">Yo!</ToastHeader>);
```
davidacevedo, line 22: I believe you made a `testForCustomTag` utility function. So maybe use that?

src/__tests__/Dropdown.spec.js
```diff
@@ -1,388 +1,320 @@
...
             <DropdownItem>Test</DropdownItem>
           </DropdownMenu>
         </Dropdown>,
       );
 
-      expect(wrapper.find('.dropdown').prop('inNavbar')).toBe(undefined);
-      expect(wrapper.find('.dropdown').prop('toggle')).toBe(undefined);
-      expect(wrapper.find('.dropdown').prop('a11y')).toBe(undefined);
-      expect(wrapper.find('.dropdown').prop('isOpen')).toBe(undefined);
+      const dropdown = document.getElementsByClassName('dropdown')[0];
```
davidacevedo, line 103: Not 100%, but you might be able to get this element from `const { container } = render(...)` where container is dropdown

davidacevedo (approved):
This test was a pretty good read 🥳

src/__tests__/ToastHeader.spec.js
illiteratewriter: `TestHeader` works a bit differently. The `prop` name for changing the tag for external class in `ToastHeader` is `wrapTag` and not `tag`. So can't use that function here.

src/__tests__/Dropdown.spec.js
illiteratewriter: I think `container` returns a `div` above the `Dropdown` and not dropdown itself. And it is a fiber node and not a DOM node. I'm also not sure tbh, but was unable to make it work with `container`

Something like 
```
<body>
  <div> // container is this
    <Dropdown />
  </div>
<body>
```

davidacevedo: I see, thanks for checking that out!

src/__tests__/ToastHeader.spec.js
davidacevedo: Makes sense and sounds good!

Issue #1289, opened 2018-11-09
Warnings in reactstrap when using React's strict mode


edmorley:
- components: `Transition`, `Modal`
- reactstrap version `6.5.0`
- import method `es`
- react version `16.6.1`
- bootstrap version `4.1.3`

### What is happening?

React 16.3 added a new strict mode, which highlights potential issues in an app (such as those that will cause problems when using the upcoming async mode):
https://reactjs.org/docs/strict-mode.html

Enabling it for our app resulted in several errors coming from `reactstrap`:

```
Warning: Legacy context API has been detected within a strict-mode tree: 
    in StrictMode (at job-view/index.jsx:28)

Please update the following components: Transition
```

-> see: https://reactjs.org/docs/strict-mode.html#detecting-legacy-context-api

```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode. Instead, add a ref directly to the element you want to reference.

    in div (created by Transition)
    in Transition (created by Fade)
    in Fade (created by Modal)
    in div (created by Modal)
    in Portal (created by Modal)
    ...
```
-> see: https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage

```
Warning: Unsafe lifecycle methods were found within a strict-mode tree:
    in StrictMode (at job-view/index.jsx:28)

componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Modal, Transition

componentWillUpdate: Please update the following components to use componentDidUpdate instead: Modal
```
-> see: https://reactjs.org/docs/strict-mode.html#identifying-unsafe-lifecycles
(this last one is half-covered by #1159, but not fully)

### What should be happening?

No warnings/errors when using reactstrap with `React.StrictMode`.

### Steps to reproduce issue

1. Clone https://github.com/mozilla/treeherder (tested using `3ad320d`)
2. Edit `ui/job-view/index.jsx` to replace the render call with:
```js
render(
  <React.StrictMode>
    <App />
  </React.StrictMode>,
  document.getElementById('root'),
);
```
3. `yarn`
4. `yarn start`
5. In the new browser tab opened by webpack-dev-server (which should end up redirecting to `http://localhost:5000/#/jobs?repo=mozilla-inbound`), open devtools to the console tab
6. Click on one of the green symbols shown on the right hand side.
7. In the panel that opens at the bottom click the "three dots" on the toolbar, then select "Custom action..."

[...]

PR #2672, opened 2022-12-20
fix(#1289): prevent error in Strict Mode

illiteratewriter:
- remove findDomNode error which was thrown for components using Fade.js file in ReactStrict mode

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 


src/Fade.js
```diff
@@ -49,15 +49,15 @@ function Fade(props) {
     className,
     cssModule,
     children,
-    innerRef,
+    innerRef = useRef(null),
```
illiteratewriter, line 52: Does this break the rules of hooks (conditonal)?

should we do something like 
```
const ref = useRef(null);

const {
  innerRef = ref,
} = props;
```

aaronmars (approved):

Issue #2666, opened 2022-12-13
Added missing toggle to Accordion


Samettkaya:
- components: `name`
- reactstrap version `#9.1.5`
- import method `umd/csj/es` 
- react version `#17.0.2`
- bootstrap version `#5.2.3`

### Steps to reproduce issue

1. ...
2. ...

### Error message in console

```
Warning: Failed prop type: The prop `toggle` is marked as required in `Accordion`, but its value is `undefined`.
```

aaronmars:
Fixed: https://github.com/reactstrap/reactstrap/pull/2642

spostad7:
> Fixed: #2642

I am using v9.2.0. The problem still  exist

Issue #2681, opened 2023-01-12
AccordionBody className prop sets class of parent div, not accordion-body div


annahansen229:
### Issue description

- components: `AccordionBody`
- reactstrap version `#9.1.5`
- react version `#16.12.0`
- bootstrap version `#5.2.3`

### What is happening?
I am trying to set zero padding on the `div.accordion-body` as follows:
`<AccordionBody className='p-0'>

I expect to see the result:
```
<div class='accordion-collapse {...}'>
  <div class='accordion-body p-0'>
     {...}
  <div/>
<div/>
```

Instead, I see the result:
```
<div class='accordion-collapse {...} p-0'>
  <div class='accordion-body'>
     {...}
  <div/>
<div/>
```

### What should be happening?

AccordionBody className props should set the class of the div with the class `accordion-body` instead of the parent div

PR #2684, opened 2023-01-23
Error fix in accordian #2666 issue fix #1

qburst-harikrishnanp:
- [ ] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/UncontrolledAccordion.js
+++ b/src/UncontrolledAccordion.js
@@ -22,7 +22,7 @@ const defaultProps = {
 };
 
 function UncontrolledAccordion({ defaultOpen, stayOpen, ...props }) {
-  const [open, setOpen] = useState(defaultOpen || (stayOpen ? [] : undefined));
+  const [open, setOpen] = useState(defaultOpen || (stayOpen ? [] : ''));
   const toggle = (id) => {
     if (stayOpen) {
       if (open.includes(id)) {
@@ -31,7 +31,7 @@ function UncontrolledAccordion({ defaultOpen, stayOpen, ...props }) {
         setOpen([...open, id]);
       }
     } else if (open === id) {
-      setOpen(undefined);
+      setOpen('');
     } else {
       setOpen(id);
     }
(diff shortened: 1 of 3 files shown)


qburst-harikrishnanp:
@illiteratewriter @aaronmars  Please approve

qburst-harikrishnanp:
@illiteratewriter @aaronmars Please approve

illiteratewriter:
The `classes` object which contain `.accordion-collapse` is supposed to be in the parent of `.accordion-body` ([Accordion Bootstrap](https://getbootstrap.com/docs/5.0/components/accordion/#example)).

@qburst-harikrishnanp Can you let us know what exactly you're trying to achieve with this particular PR?

febababy:
@illiteratewriter 
The PR is regarding the issue from https://github.com/reactstrap/reactstrap/issues/2681 to set styles to accordion-body.
There is a warning msg while using the accordion, https://github.com/reactstrap/reactstrap/issues/2666. This is due to the undefined while using setOpen props.

This PR will resolve both the above mentioned issues.

illiteratewriter:
The current PR has the issue that it is changing the structure of how accordion works in bootstrap. In Bootstrap docs the accordion works as following:
```jsx
<div class="accordion-collapse collapse">
 <div class="accordion-body">
  This is the second item's accordion body.
  </div>
</div>
```

As you can see, `.accordion-body` is the child of `.accordion-collapse`. But according to this PR, in the output we get something like: 
```jsx
<div class="collapse">
 <div class="accordion-body accordion-collapse ">
  This is the second item's accordion body.
  </div>
</div>
```

This also breaks the library for those who have applied classes previously, because there is a change in behavior.

The solution to this would be to add an additional prop, something like `bodyClassName` which can be applied to the second div. In essence, we would keep the `className` prop working like how it was previously and add a new prop (`bodyClassName`) to add custom classes to accordion body.

[Modal.js ](https://github.com/reactstrap/reactstrap/blob/827b98b284973ff2852999e0a8548da7923ba73b/src/Modal.js#L553)implements something similar to this. You can look at it for inspiration. 

Let me know if you have any questions.

febababy:
Thanks for the input.

febababy:
PR updated with issue fix for https://github.com/reactstrap/reactstrap/issues/2666 , reverted the changes done for https://github.com/reactstrap/reactstrap/issues/2681.

image

PR #2752, opened 2023-05-15
fix(#2662): remove many instances of default props

### LINKED ISSUE #2662: Warning: Container: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
React seems to be going away from defaultProps, and we should use default parameters instead

```
const defaultProps = {
  tag: 'div',
};
//Remove above and add like this i assume?
function Container(props) {
  const { className, cssModule, fluid, tag: Tag = 'div', ...attributes } = props;
```

Low hanging fruit for new contributor to carry out?

amillward:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

Partially fixes #2662

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,6 +137,14 @@
 
 
 
+## [9.1.10](https://github.com/reactstrap/reactstrap/compare/v9.1.9...v9.1.10) (2023-05-13)
+
+
+### Bug Fixes
+
+* **#2427:** error building esm files ([#2748](https://github.com/reactstrap/reactstrap/issues/2748)) ([04d571d](https://github.com/reactstrap/reactstrap/commit/04d571dcaa630ccbc8b8363ad06dad436f11780e))
+* properly pass in `isOpen` in modal test ([#2745](https://github.com/reactstrap/reactstrap/issues/2745)) ([8cb0515](https://github.com/reactstrap/reactstrap/commit/8cb051522ddfad48c08651604744caea25b5e68e))
+
 ## [9.1.9](https://github.com/reactstrap/reactstrap/compare/v9.1.8...v9.1.9) (2023-04-11)
 
 
--- a/babel.esm.config.cjs
+++ b/babel.esm.config.cjs
@@ -1,5 +1,5 @@
 module.exports = {
-  extends: './babel.build.config.cjs',
+  extends: './babel.config.cjs',
   presets: [
     // Dont transform modules for the esm build
     ['@babel/preset-env', { modules: false }],
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "reactstrap",
-  "version": "9.1.9",
+  "version": "9.1.10",
   "description": "React Bootstrap components",
   "source": "src/index.js",
   "main": "lib/index.js",


amillward:
I'm aware this doesn't fully address `defaultProps` for all components (I am too time-poor right now to do the more complex cases) but hopefully this is a start as this warning is now extremely noisy

src/ButtonToggle.js
```diff
@@ -13,12 +13,8 @@ const propTypes = {
   cssModule: PropTypes.object,
 };
 
-const defaultProps = {
-  defaultValue: false,
-};
-
 function ButtonToggle(props) {
-  const [toggled, setToggled] = useState(props.defaultValue);
+  const [toggled, setToggled] = useState(props.defaultValue || false);
```
illiteratewriter, line 17: Would the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) be a better choice here?

src/CardTitle.js
```diff
@@ -12,12 +12,8 @@ const propTypes = {
   tag: tagPropType,
 };
 
-const defaultProps = {
-  tag: 'div',
-};
-
 function CardTitle(props) {
-  const { className, cssModule, tag: Tag, ...attributes } = props;
+  const { className, cssModule, tag: Tag = 'div', ...attributes } = props;
```
illiteratewriter, line 16: Since this is a title, I think it'd be better to use `h5` or something here.

src/DropdownItem.js
```diff
@@ -42,7 +37,7 @@ class DropdownItem extends React.Component {
       this.props.onClick(e);
     }
 
-    if (this.props.toggle) {
+    if (this.props.toggle || true) {
```
illiteratewriter, line 40: Here too, I think we can use [??](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)

src/Fade.js
illiteratewriter, line 1: Can we remove `Fade.defaultProps = defaultProps` for this file?

src/NavbarToggler.js
```diff
@@ -15,21 +15,16 @@ const propTypes = {
...
+  const { className, cssModule, children, tag: Tag = 'button', ...attributes } = props;
 
   const classes = mapToCssModules(
     classNames(className, 'navbar-toggler'),
     cssModule,
   );
 
   return (
-    <Tag aria-label="Toggle navigation" {...attributes} className={classes}>
+    <Tag aria-label="Toggle navigation" {...{type: 'button',...attributes}} className={classes}>
```
illiteratewriter, line 27: Do we need to do `{...{type: 'button',...attributes}}`? Shouldn't just `{type: 'button',...attributes}` work?

src/Toast.js
```diff
@@ -33,11 +30,11 @@ function Toast(props) {
   const {
     className,
     cssModule,
-    tag: Tag,
-    isOpen,
+    tag: Tag = 'div',
+    isOpen = true,
     children,
     transition,
```
illiteratewriter, line 36: We can add `transition` also as a default parameter right? Is there any issue in doing so and removing `Toast.defaultProps`?

illiteratewriter (changes requested):

illiteratewriter (changes requested):

illiteratewriter:
Thank you so much for the PR.

I added comments earlier instead of a review. Sorry about that.

Also, Reactstrap uses `eslint` for [linting](https://www.freecodecamp.org/news/what-is-linting-and-how-can-it-save-you-time/), and `prettier` for code styles. 
If you run `yarn lint` before making any code changes now, you'll be able to see a couple of errors. Eslint helps with such small errors.
Also, once you're done; run `yarn prettier`. It keeps code style consistent across the project.

src/CardTitle.js
amillward: My intention is to not change any behaviour with this pr. The original default value is `div` which I have preserved.

src/Fade.js
amillward: `Fade.defaultProps` is used in a number of other modules so would take a decent chunk more time to remove which I don't have right now - sorry!

src/NavbarToggler.js
amillward: `{type: 'button',...attributes}` defaults `type` if it's not present in `attributes`. The wrapping `{...obj}` spreads the resulting object to be rendered as html attributes

src/Toast.js
amillward: I was lazy and avoided this as it is not a simple case. I think it needs a deep merge because there are nested objects. Simply doing 
```
transition = {
    ...Fade.defaultProps,
    unmountOnExit: true,
  }
```

would not allow for, e.g. prop of `{transition: {appear: false}}`. `transition` wouldn't be empty, so it would lose all the other defaultProp values

amillward:
@illiteratewriter thanks for the feedback. I have responded to them all. Happy to discuss any further :)

src/Toast.js
TheSharpieOne: > I was lazy and avoided this as it is not a simple case. I think it needs a deep merge because there are nested objects. Simply doing
> 
> ```
> transition = {
>     ...Fade.defaultProps,
>     unmountOnExit: true,
>   }
> ```
> 
> would not allow for, e.g. prop of `{transition: {appear: false}}`. `transition` wouldn't be empty, so it would lose all the other defaultProp values

That is how default props work now; defaultProps doesn't merge objects, it just uses the default if the prop is `undefined`. Later in the code, a shallow merge is done to add any missing default values on the object root (which, looking at the Fade component, is all that really matters.) Maybe `unmountOnExit: true,` should be added there (but that would technically be a change to how it currently would work.)

amillward: @TheSharpieOne ah I did not realise that (I've never used `defaultProps` myself). I've updated in this instance. Thank you

singleseeker:
@illiteratewriter Hi, what's next? Do you plan to merge this PR?

illiteratewriter (approved):

Issue #2744, opened 2023-05-05
Toast animation is not working


abaksha-sc:
- components: `Toast`
- reactstrap version `9.1.9`
- import method `es` 
- react version `18.2.0`
- bootstrap version `5.2.3`
- 
### What is happening?

Toast has prop `fade` which is `true` by default and means (as I understand) the animation.  
But it's not working (even in documentation).

### What should be happening?

Toast should appear with animation

### Steps to reproduce issue

1. Open the fiddle
2. Click button "Toggle toast"

### Code

https://stackblitz.com/edit/reactstrap-aauune?file=Example.js

### Additional

Seems like prop `transition.timeout` just delays the appearing instead of to make animation longer.  
Here is an example with `timeout: 1000`: https://stackblitz.com/edit/reactstrap-t1eesr?file=Example.js

illiteratewriter:
I am not sure what is causing this at the moment. Seems like the class `toast` is causing some issues. [On removing the class `toast`](https://github.com/reactstrap/reactstrap/blob/8cb051522ddfad48c08651604744caea25b5e68e/src/Toast.js#L45) from Toast.js, everything's working as expected.

PR #2770, opened 2023-09-15
fix(#2744): Toast fade fix with display block

davidlyons:
- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

Boostrap has a [.toast style](https://github.com/twbs/bootstrap/blob/main/scss/_toasts.scss#L35) that breaks Reactstrap's Toast fade transition.  
This PR overrides the Toast's display: none with the `d-block` utility class to fix the fade transition.

Fix for #2744

### DIFF
--- a/src/Toast.js
+++ b/src/Toast.js
@@ -35,7 +35,10 @@ function Toast(props) {
     ...attributes
   } = props;
 
-  const classes = mapToCssModules(classNames(className, 'toast'), cssModule);
+  const classes = mapToCssModules(
+    classNames(className, 'toast d-block'),
+    cssModule,
+  );
 
   const toastTransition = {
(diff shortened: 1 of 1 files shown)


illiteratewriter:
@davidlyons Great catch on finding what was causing the issue.

The PR kind of fixes the issue but this is causing Reactstrap to behave differently from how Bootstrap is behaving right now. Ideally, we would like to have the same behaviour that Bootstrap is providing.

mizozobu:
How about resetting bootstrap style? The repo has no CSS, so we have to figure out where to place the code though.

```css
.toast:not(.show) {
  display: initial;
}
```

PR #2791, opened 2024-01-26
fix: carousel should stop immediately after interval is set to false

### LINKED ISSUE #2779: Carousel: switching interval to false does not deactivate autoplay directly
- components: `carousel`
- reactstrap version `#8.10.1`
- react version `#17.0.2`
- bootstrap version `#4.5.2`

### What is happening?

With the Carousel, we can deactivate the autoplay by setting the interval props to `false`. However, this is not taken into account directly but only after one last automatic transition to the next slide. I think this is due to the the following piece of code  in carousel.js : 

```
componentDidUpdate(prevProps, prevState) {
    if (prevState.activeIndex === this.state.activeIndex) return;
    this.setInterval();
  }
```

An additional check between prevState.interval and this.state.interval should be perform, no?

### What should be happening?

Autoplay should stop as soon as the interval is set to false

### Steps to reproduce issue

1. Load the component with an interval
2. Switch the interval to false

MariusJam:
- [x ] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 
- [ x] There is an open issue which this change addresses
- [x ] I have read the **[CONTRIBUTING](./CONTRIBUTING.md)** document.
- [x ] My commits follow the [Git Commit Guidelines](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits)
- [x ] My code follows the code style of this project.
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.

Closes #2779


illiteratewriter:
Thank you so much for the PR.

A better solution, in my opinion, would be to clear the interval if the props has changed.
Something like:
```
 componentDidUpdate(prevProps, prevState) {
    if (prevProps.interval !== this.props.interval) {
      this.setInterval();
    }
    if (prevState.activeIndex === this.state.activeIndex) return;
    this.setInterval();
  }
```

Currently, there is another bug which is once the carousel is stopped, it does not restart even if you toggle the interval back to a number. the above fixes that issue also. (We are calling setInterval again because it has a clearInterval inside it which takes care of clearing the interval and running the interval again only if needed).

Also the second thing is currently we are deprecating enzyme so it would be great if you can change the tests to run using react testing library instead.

src/__tests__/Carousel.spec.js
```diff
@@ -765,5 +766,23 @@ describe('Carousel', () => {
       jest.advanceTimersByTime(1000);
       expect(next).toHaveBeenCalled();
     });
```
illiteratewriter, line 768: It would be great if you can rewrite the test to use react testing library

src/Carousel.js
```diff
@@ -112,7 +112,12 @@ class Carousel extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    if (prevState.activeIndex === this.state.activeIndex) return;
+    if (
```
illiteratewriter, line 115: better solution as mentioned in the comments

illiteratewriter (changes requested):

src/Carousel.js
MariusJam: Hi,

I can implement the solution you propose indeed but there will be one side effect (should it be considered as a breaking change?): it will restart the interval to 0 at every change. So, let's say you set an interval at 5s and, after 4s, you change the interval to 10s, it will wait 10 more seconds before transitioning. So, in total, it will be 14 seconds before the first transition whereas the current behaviour is different: it does the first transition after 5 seconds and a second transition after 10 seconds.

I guess that if we want to avoid this side effect, we should check that we are in a case of stopping/restarting the carousel. For instance:

```javascript
 componentDidUpdate(prevProps, prevState) {
    if (prevProps.interval !== this.props.interval && (prevProps.interval === false || this.props.interval === false)) {
      this.setInterval();
    }
    if (prevState.activeIndex === this.state.activeIndex) return;
    this.setInterval();
  }
```

Would you be okay with this proposal?

illiteratewriter: `&& (prevProps.interval === false || this.props.interval === false)) {` this will be the case every time we stop and restart right? which means this is always going to be true I guess.

MariusJam: > this will be the case every time we stop and restart right?

Yes indeed and I think that's what we want. In that way we fix both issues: the one described in https://github.com/reactstrap/reactstrap/issues/2779 and the one you mentioned in your first comment.

> which means this is always going to be true I guess (as in on the edge case you mentioned).

I don't think so. Or I'm missing something. :thinking:. As far as I understand, `prevProps.interval` would be equal to `5000` and `this.props.interval` to `10000` in my example. So, in that case, `&& (prevProps.interval === false || this.props.interval === false)) {` is going to be false.

illiteratewriter: My bad. Sure, sounds good to me.

PR #2806, opened 2024-05-21
chore(fixes #2801): use function default parameters

### LINKED ISSUE #2801: Fade: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
chunk-FUOFC43Q.js?v=2a0f9713:521 Warning: Fade: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
    at Fade (

islam-kamel:
chore(fixes #2801): use function default parameters instead of **defaultProps**

- [x] Bug fix 
- [ ] New feature 
- [x] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -264,8 +264,8 @@ export function toNumber(value) {
   return isBinary || /^0o[0-7]+$/i.test(value)
     ? parseInt(value.slice(2), isBinary ? 2 : 8)
     : /^[-+]0x[0-9a-f]+$/i.test(value)
-    ? NAN
-    : +value;
+      ? NAN
+      : +value;
 }
 
 export function isFunction(value) {
@@ -381,3 +381,23 @@ export const focusableElements = [
   'video[controls]',
   '[contenteditable]:not([contenteditable="false"])',
 ];
+
+export function addDefaultProps(defaultProps, props) {
+  if (!defaultProps || !props) return props;
+
+  let result = { ...props };
+
+  Object.keys(defaultProps).forEach((key) => {
+    if (result[key] === undefined) {
+      result[key] = defaultProps[key];
+    }
+    if (
+      Object.keys(defaultProps[key] || {}).length > 0 &&
+      typeof defaultProps[key] === 'object'
+    ) {
+      addDefaultProps(defaultProps[key], result);
+    }
+  });
+
+  return result;
+}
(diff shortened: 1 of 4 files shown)


islam-kamel:
@illiteratewriter please check these changes.

kamenminkovcom:
Please, update all the components that use defaultProps.

illiteratewriter:
LGTM

sbaechler:
What is the reason for not using Javascript default props?

sbaechler:
> What is the reason for not using Javascript default props?

Ah, because even undefined props are set (by the prop types?) to undefined and would override the default props.

src/utils.js
```diff
@@ -381,3 +381,23 @@ export const focusableElements = [
...
+
+  let result = { ...props };
+
+  Object.keys(defaultProps).forEach((key) => {
+    if (result[key] === undefined) {
+      result[key] = defaultProps[key];
+    }
+    if (
+      Object.keys(defaultProps[key] || {}).length > 0 &&
+      typeof defaultProps[key] === 'object'
```
illiteratewriter, line 396: This is not considering array (`typeof` array will be object, so it would be better to have something like `!Array.isArray(defaultProps[key])`)

illiteratewriter, line 396: This is not handling the array and the `null` case correctly. We do not need to merge if it is an array so the conditions should be somwhere around the lines of `!Array.isArray(defaultProps[key])` and should also have a check for `null` case `defaultProps[key] !== null` because `typeof null === 'object'` is true

illiteratewriter (approved):

charlesfayal:
I'm still seeing this issue with v9.2.3 on the Popover component 

`console.js:288 Warning: Popover: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
    at Popover (http://localhost:3000/static/js/bundle.js:503660:95)`

SabihYallaMotor:
> I'm still seeing this issue with v9.2.3 on the Popover component
> 
> `console.js:288 Warning: Popover: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead. at Popover (http://localhost:3000/static/js/bundle.js:503660:95)`

same for tooltip and popover component...

PR #2812, opened 2024-06-28
refactor(#2801): Remove defaultProps api for function components 

### LINKED ISSUE #2801: Fade: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.

sbaechler:
- [ ] Bug fix 
- [ ] New feature 
- [x] Chore 
- [ ] Breaking change 

Fixes #2801 

This pull request is based on https://github.com/reactstrap/reactstrap/pull/2806 by @islam-kamel.

The defaultProps were removed from the 3 functional components Fade, Tooltip and Popover.

It does require a new method that actively checks the props for `undefined`. Most likely because propTypes adds props that are not passed-in. However, the new method only does a shallow comparison (since functional components only do a shallow compare).

### DIFF
--- a/src/Fade.js
+++ b/src/Fade.js
@@ -3,12 +3,13 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { Transition } from 'react-transition-group';
 import {
+  addDefaultProps,
   mapToCssModules,
   omit,
   pick,
+  tagPropType,
   TransitionPropTypeKeys,
   TransitionTimeouts,
-  tagPropType,
 } from './utils';
 
 const propTypes = {
@@ -50,12 +51,13 @@ function Fade(props) {
     children,
     innerRef = ref,
     ...otherProps
-  } = props;
+  } = addDefaultProps(defaultProps, props);
 
   const transitionProps = pick(
     { defaultProps, ...otherProps },
     TransitionPropTypeKeys,
   );
+
   const childProps = omit(otherProps, TransitionPropTypeKeys);
 
   return (
@@ -77,6 +79,5 @@ function Fade(props) {
 }
 
 Fade.propTypes = propTypes;
-Fade.defaultProps = defaultProps;
 
 export default Fade;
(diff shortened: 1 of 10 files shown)


src/Fade.js
```diff
@@ -50,12 +51,12 @@ function Fade(props) {
     children,
     innerRef = ref,
     ...otherProps
-  } = props;
+  } = addDefaultProps(fadeDefaultProps, props);
 
-  const transitionProps = pick(
-    { defaultProps, ...otherProps },
+  const transitionProps = pick(otherProps,
```
sbaechler, line 56: @amillward I changed this line to remove the `defaultProps` prop. 
I was assuming the previous version was a mistake because `defaultProps` would be passed-in as prop. I could not find a hint in the code or the documentation for `Transition` that this prop is supported. There is no change in the test results either. Can you confirm that this change is ok?

sbaechler: Actually, it would be immediately removed by the `pick` function.

vladdg-dev:
Please remove defaultProps for all affected components. It's not only Fade, Tooltip and Popover. It's also Nav, NavItem, Badge, Breadcrumb, BreadcrumbItem, Container, Row, Col, Card, CardHeader, CardBody. Thank you.

sbaechler:
> Please remove defaultProps for all affected components. It's not only Fade, Tooltip and Popover. It's also Nav, NavItem, Badge, Breadcrumb, BreadcrumbItem, Container, Row, Col, Card, CardHeader, CardBody. Thank you.

@vladdg-dev defautProps are only deprecated for function components. They are still supported for class components and all other components are class components.

mizozobu:
We need this. Some components are referencing [`Fade.defaultProps`](https://github.com/search?q=repo%3Areactstrap%2Freactstrap%20Fade.defaultProps&type=code), but `Fade.defaultProps` is `undefined`.

I get the warning when I use `Toast`.

```
Warning: Failed prop type: The prop `timeout` is marked as required in `Fade`, but its value is `undefined`.
```

PR #2814, opened 2024-08-02
fix: prevent setopen to set undefined value for open

kevinrawal:
## Issue Description

**Problem:**
A component that uses an accordion component is experiencing a warning while executing the test cases in jest when toggled to a closed state. The `open` prop, defined as required, is being set to `undefined` during this process. This leads to test failures in components utilizing the Accordion due to the required prop violation.

**fix**
There is no issue in the accordion component the only concern is how we are using it, currently, the use defined in the example is violating the tests, fixing the `setOpen` to an empty string instead of undefined fix the issue. 

- [x] Bug fix 
- [ ] New feature 
- [] Chore 
- [ ] Breaking change 

### DIFF
--- a/src/UncontrolledAccordion.js
+++ b/src/UncontrolledAccordion.js
@@ -27,7 +27,7 @@ function UncontrolledAccordion({ defaultOpen, stayOpen, ...props }) {
         setOpen([...open, id]);
       }
     } else if (open === id) {
-      setOpen(undefined);
+      setOpen('');
     } else {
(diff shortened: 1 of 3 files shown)


kevinrawal:
@illiteratewriter can you please review this PR? Fixed a minor but impactful issue in the example code to ensure proper functionality, as many developers directly use it.

illiteratewriter (approved):

PR #2815, opened 2024-08-15
fix: change default open state to empty string

kevinrawal:
UncontrolledAccordion inherits the properties of the Accordion component. We are setting the `defaultOpen` prop in UncontrolledAccordion, This defaultOpen will then be used to set the open prop, if we do not put any values to defaultOpen it will initialize the open prop with an undefined value which breaks the test case of the component which uses the `UncontrolledAccordion`. because the open prop is marked as required and we have set it to undefined. 

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 

This PR is related to https://github.com/reactstrap/reactstrap/pull/2814, The same issue was resolved in the previous PR but a minor change is required in uncontrolled accordion.

### DIFF
--- a/src/UncontrolledAccordion.js
+++ b/src/UncontrolledAccordion.js
@@ -18,7 +18,7 @@ const propTypes = {
 };
 
 function UncontrolledAccordion({ defaultOpen, stayOpen, ...props }) {
-  const [open, setOpen] = useState(defaultOpen || (stayOpen ? [] : undefined));
+  const [open, setOpen] = useState(defaultOpen || (stayOpen ? [] : ''));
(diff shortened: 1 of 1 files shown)


kevinrawal:
@illiteratewriter can you please review this PR, Fixed the minor issue related to uncontrolled accordion which is similar to my previous commit https://github.com/reactstrap/reactstrap/pull/2814. This changes insure proper functionality of `UncontrolledAccordion`.

VanemalP:
Setting `isOpen` to a string is breaking the prop type validation in Collapse component.

Issue #4297, opened 2025-04-22
Badge colour is not working with `bootstrap@5.3.x`


namti:
- components: `Badge`
- reactstrap version `9.2.3`
- import method `es`
- react version `18.2.0`
- bootstrap version `5.3.3`

### What is happening?

The new `bootstrap@5.3.x` has changed the class name from `bg-{color}` to `text-bg-{color}`

When use 
```jsx
<Badge color="warning" className="rounded-pill">DEV</Badge>
```
The text color is not correct.

![Image](https://github.com/user-attachments/assets/11b50161-0fdd-476b-83f8-ac585b6b2a29)

The colour is expected to be:

![Image](https://github.com/user-attachments/assets/3c7cbc71-46c6-41c1-bb37-480cbb7c5b37)

### Code

HTML generated by `reactstrap@9.2.3`

```html
<span class="badge bg-warning rounded-pill">DEV</span>
```

Should be:

```html
<span class="badge text-bg-warning rounded-pill">DEV</span>
```

PR #4337, opened 2025-07-26
bugfix Badge colour is not working with bootstrap@5.3.x #4297

elaboratecon:
…lor standards on the badge component

- [x] Bug fix 
- [ ] New feature 
- [ ] Chore 
- [ ] Breaking change 


elaboratecon:
This addresses issue #4297 

I reproed the issue using Codesandbox, first modifying package.json and associated code to reflect the library versions listed in the issue:
[https://codesandbox.io/p/devbox/blazing-currying-lzk4ky?workspaceId=ws_FCSZKN4NMddcmrKh6txxF4](url)

This updates the color class from the deprecated bg-color (ie "bg-warning") to match Bootstrap's 5.3.x standards (ie "text-bg-warning"):
[https://getbootstrap.com/docs/5.3/components/badge/](url)

Apologies for the crazy branch name.

nachogarciait (approved):

src/Badge.js
```diff
@@ -38,7 +38,7 @@ function Badge(props) {
     classNames(
       className,
       'badge',
-      'bg-' + color,
+      'text-bg-' + color,
```
deAtog: The `text-bg-*` classes were added 5.2.0, for this to be compatible with all versions of 5.x it would need to use the bg-* classes and add the `text-dark` class if the color is `info`, `warning`, or `light` like shown here: https://getbootstrap.com/docs/5.1/components/badge/#background-colors Otherwise, for those using Bootstrap 5.1.x this would be a breaking change.

Admittedly, if ReactStrap had a peerDependency on which version of Bootstrap it was compliant with we would know whether or not we could directly target 5.2.x and newer.. ¯\_(ツ)_/¯

